PR #81, opened 2017-09-27
Remove `bin` from package.json

swashcap:
* **Problem:** `npm link` doesn't work because _package.json_’s `bin` entry points to a non-existent file
* **Solution:** Fix `bin` entry
* **Testing:** Try `npm link` in the repo's root (`npm unlink` to undo)

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "flow"
   ],
   "bin": {
-    "ern": "ern-global-cli/bin/ern.js"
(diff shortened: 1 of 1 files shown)


krunalsshah:
@swashcap Actually please remove the `bin` section altogether. Thanks .

swashcap:
@krunalsshah :+1: Dropped in bf4e0f6d45f7ca2d35c60e742875208f6aec5d6f.

PR #84, opened 2017-09-27
Add metadata to package.json

swashcap:
* **Problem:** The package.json lacks some metadata. npm complains:
    ```
    npm WARN ern-platform@0.5.1 No repository field.
    ```
* **Solution:** Add the metadata!
* **Testing:** Make sure the JSON isn't malformed:
    ```
    node -pe 'require("./package.json")'
    ```

    (Should output JSON and not throw)

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.5.1",
   "description": "",
   "keywords": [],
-  "author": "",
+  "author": "Walmart Labs <http://walmartlabs.com/>",
   "license": "Apache-2.0",
   "scripts": {
     "setup-dev": "node setup-dev.js",
@@ -77,8 +77,13 @@
   },
   "engines": {
     "node": ">=4.5"
+  },
+  "bugs": {
+    "url": "https://github.com/electrode-io/electrode-native/issues"
+  },
   "repository": {
     "type": "git",
     "url": "https://github.com/electrode-io/electrode-native.git"
-  }
+  },
+  "homepage": "http://www.electrode.io"
 }
--- a/package.json
+++ b/package.json
@@ -77,5 +77,8 @@
   },
   "engines": {
     "node": ">=4.5"
+  "repository": {
+    "type": "git",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -3,7 +3,7 @@
   "version": "0.5.1",
   "description": "",
   "keywords": [],
-  "author": "",
+  "author": "Walmart Labs <http://walmartlabs.com/>",
```
swashcap: Wasn't sure if there's an email or preferred electrode author here. We should add [a `contributors` field](https://docs.npmjs.com/files/package.json#people-fields-author-contributors) to list all the folks!

belemaire: This top level `package.json` is not actually used for publication, we don't publish this module on NPM, it's just the top level `package.json` mostly for lerna (multi package setup). All `ern-*` packages are getting published and the contributors field is there for each of them in their `package.json`. 
For this one we should just remove the author, unless there is a good reason to use this field (for all `ern-*` modules we just removed this field). I feel like it's good to use this field if the author is a real person, not an org.
FYI all packages get published using `electrode-native` user on GitHub https://www.npmjs.com/~electrode-native, it also has the email `electrodenative@gmail.com` ... So in case we use an author we should use this email and username. But I feel it's not really needed here, lets just remove it.

belemaire:
Merging this one, thanks a lot @swashcap 

Will update author as another PR. Once again this package is not published to NPM so we don't really need all these fields, but it makes it more proper to the eye ;)

package.json
swashcap: Totally makes sense! We should maybe add `"private": true` to the root `package.json` to make sure nobody accidentally publishes.Totally makes sense! We should maybe add `"private": true` to the root `package.json` to make sure nobody accidentally publishes.

swashcap: Totally makes sense! We should maybe add `"private": true` to the root `package.json` to make sure nobody accidentally publishes.

PR #90, opened 2017-09-28
Fix Typos in getting started and remove text

krunalsshah:
Amended section of the documentation.

### DIFF
--- a/docs/getting-started/getting-started.md
+++ b/docs/getting-started/getting-started.md
@@ -215,8 +215,8 @@ Now let's add an API to the MiniApp so that we can retrieve movies from the nati
 
 ## Adding the MoviesApi to the MovieList MiniApp
 
-We have already generated and published the `MoviesApi` in this tutorial.
-You may view the generated API code at [react-native-ernmovie-api](https://github.com/electrode-io/react-native-ernmovie-api) and [ern add react-native-ernnavigation-api](https://github.com/electrode-io/react-native-ernnavigation-api) repo.
+We have already created and published the `MoviesApi` in this tutorial.
+You may view the created API code at [react-native-ernmovie-api](https://github.com/electrode-io/react-native-ernmovie-api) and [ern add react-native-ernnavigation-api](https://github.com/electrode-io/react-native-ernnavigation-api) repo.
 
 1) Using the `ern add` command add `MoviesApi` , other dependent APIs and [react-native-electrode-bridge](https://github.com/electrode-io/react-native-electrode-bridge) as a dependency of MovieListMiniApp.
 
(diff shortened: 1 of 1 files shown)


docs/getting-started/getting-started.md
```diff
@@ -433,15 +433,15 @@ You can see that the UI now shows the movies that are stored in the native appli
 
 ## Adding the Navigation API
 
-We created this API for you for this tutorial--you can view it in the GitHub repository located here: [this GitHub repository](https://github.com/electrode-io/react-native-ernnavigation-api). This very simple API will be used for navigating from the `MovieListMiniApp` to the `MovieDetailsMiniApp`.
+We generated this API for you for this tutorial and the code is hosted at [react-native-ernnavigation-api](https://github.com/electrode-io/react-native-ernnavigation-api) repo. This API will be used for navigating from the `MovieListMiniApp` to the `MovieDetailsMiniApp`.
```
belemaire: Can you revert this edit.
Let's keep `created` instead of `generated`. `create-api` is the command. The fact that it's generated is a technical detail of how it gets created. So yes let's stick to the higher level simpler term in the getting started.
Also let's keep the `GitHub repository located here`. Not too big fan of `view` but anyway the first phrase is better, also without abbreviation `repo` vs `repository`. we shouldn't use too much abbr. in a getting started.
this was edited by Catherine I think, there were good reasons behind this wording for the most part ;)

docs/getting-started/getting-started.md
```diff
@@ -433,15 +433,15 @@ You can see that the UI now shows the movies that are stored in the native appli
...
 
-{% sample lang="android" %}  
-
-1) Mofify the `index.android.js` file as follows so that when selecting a movie in the list, the `MovieListMiniApp` will call the navigation API to navigate to the `MovieDetailsMiniApp` for this movie.
+{% sample lang="android" %}
+1) Modify the `index.android.js` file so that clicking the movie item navigates from  `MovieListMiniApp` to `MovieDetailsMiniApp`.
```
belemaire: mmmm.
Prefer the first version here as well. First we don't really `click` in a mobile app (yes well not my fault if some mobile dev say that or if Android have an onClick method ;P) , we `tap` or we `select` or `press`. `click` is really for web / desktop with my old good mouse :). And not just for that, I think first version works better .

docs/getting-started/getting-started.md
```diff
@@ -433,15 +433,15 @@ You can see that the UI now shows the movies that are stored in the native appli
...
 
 {% sample lang="ios" %}  
-1) Modify the `index.ios.js` file as follows so that when selecting a movie in the list, the `MovieListMiniApp` will call the navigation API to navigate to the `MovieDetailsMiniApp` for this movie.
-{% common %}  
+1) Modify the `index.ios.js` file so that clicking the movie item navigates from  `MovieListMiniApp` to `MovieDetailsMiniApp`.
```
belemaire: Same remark as before, in part, so I would keep previous version

docs/getting-started/getting-started.md
```diff
@@ -478,7 +478,7 @@ import { NavigationApi } from 'react-native-ernnavigation-api'
   }
 ```  
 
-4) Add a method below the `render` method to send the `navigate` request when a movie is selected in the list of movies:
+4) Add a method below the `render` to send the `navigate` request when a movie is selected in the list of movies:
```
belemaire: Yes let's keep this one. `method` was missing from it.

docs/getting-started/getting-started.md
```diff
@@ -489,11 +489,10 @@ import { NavigationApi } from 'react-native-ernnavigation-api'
   }
 ```
 
-5) Implement the `NavigationAPI`.
-The `navigate` method of the `NavigationApi` is not yet implemented. We'll implement it in the native application, as we did for the `MovieApi`.  
+5) Implement the `NavigationApi` in the native application, as we did for the `MovieApi`.
 
 {% sample lang="android" %}  
-* Replace all of the `MainActivity.java` content with the following code:
+* Replace the content of `MainActivity.java` with the following code:
```
belemaire: I like your better as well, let's keep it.

docs/getting-started/getting-started.md
belemaire: Still .. in the original version :`in the GitHub repository located here: [this GitHub repository](https://github.com/electrode-io/react-native-ernnavigation-api)`

That's not right, just noticed it ... should rather be `in the [NavigationApi GitHub repository](https://github.com/electrode-io/react-native-ernnavigation-api)`

docs/getting-started/getting-started.md
```diff
@@ -585,7 +584,7 @@ public class MainActivity extends AppCompatActivity {
 
 To complete the tutorial, add the `MovieDetailsMiniApp` to the application.
 
-We've developed and published this MiniApp so you can reuse it for this tutorial. If you want to view the code of the MiniApp, you can view it in [this GitHub repository](https://github.com/electrode-io/MovieDetailsMiniApp)
+We've developed and published this MiniApp to reuse it in this tutorial. You may view the source code of [MovieDetailsMiniApp](https://github.com/electrode-io/MovieDetailsMiniApp) repo.
```
belemaire: I prefer your wording. But for the second phrase would use instead something like `You may view the code of this MiniApp in the [MovieDetailsMiniApp repository](https://github.com/electrode-io/MovieDetailsMiniApp)`

PR #91, opened 2017-09-28
fix minor places in getting started

clairewli:
- Modify Nav Bar names for both miniapp
- set nav bar translucent to NO


belemaire:
Thanks @clairewli :+1:

PR #107, opened 2017-09-28
Correct loading message

swashcap:
Noticed this while re-booting:

wat

This corrects the loading message typo!


swashcap:
Mandatory:

WAT lady meme

PR #157, opened 2017-10-06
Cleanup api impl gen

deepueg:


ern-api-impl-gen/src/index.js
```diff
@@ -55,22 +56,37 @@ export async function generateApiImpl ({
...
-function createOutputFolder (outputFolderPath: string, forceGenerate) {
+async function createOutputFolder (outputFolderPath: string, forceGenerate) {
   if (!forceGenerate && fs.existsSync(outputFolderPath)) {
-    Utils.logErrorAndExitProcess(`An implementation directory already exists in ${outputFolderPath}. \nIf you want to force the creation of this API implementation project use -f option in your command.`)
-    // RETURN
-  } else {
-    if (forceGenerate && fs.existsSync(outputFolderPath)) {
-      log.info(`Deleting the existing directory and recreating a new one in ${outputFolderPath}`)
-      shell.rm('-R', outputFolderPath)
+    let result = await inquirer.prompt(
```
belemaire: What you can do here to simplify and make it more aligned with other `inquirer` calls throughout the code, is to rely on ES6 object destructuring. i.e: 

instead of `let result = await inquirer.prompt` and later on `result.confirmation`, you can just do 
`let { confirmation } = await inquirer.prompt` and then use `confirmation` directly.

Would also suggest using `const` instead of `let` (always favor `const` first, only use `let` in case you need to modify the variable later on) and maybe renaming `confirmation` in something less vague, maybe `shouldRegenerate` or something like that.

this way the `if` statement later on is more readable (`if (!shouldRegenerate)`) instead of `if (!result.confirmation)`. For the latter, a developer has to really look what is `result.confirmation` about, where its set, whereas for the former, the statement is self explanatory.

Also ... I'm not sure about this. Why do we ask the user here for anything ? If he provided the `-f` flag then it means he already wants to do that. If it's a confirmation then the message is not proper, it should be instead something like `Are you sure you want to delete the existing implementation and regenerate it ?` and the variable name `userHasConfirmedRegeneration` instead of `confirmation` or `shouldRegenerate`

ern-local-cli/src/commands/create-api-impl.js
```diff
@@ -80,21 +82,25 @@ exports.handler = async function ({
...
+      nativeOnly,
+      forceGenerate: force,
+      reactNativeVersion,
+      paths: {
+        apiImplHull: path.join(platformPath, `ern-api-impl-gen/hull`),
+        pluginsDownloadFolder: PLUGIN_FOLDER,
+        workingFolder: WORKING_FOLDER,
+        outFolder: ''
+      }
+    }).then(() => {
```
belemaire: Do not use `.then` is ... no really proper to mix/match ES7 async/await constructs with old ES6 promises.
you can just do instead : 

```javascript
try {
  await generateApiImpl({....})
  log.info('Success!')
} catch (e) {
  log.error(...)
}
```

PR #165, opened 2017-10-09
Add GLOSSARY, define complete & partial native app descriptor

krunalsshah:
Add GLOSSARY , define complete & partial native app descriptor

### DIFF
--- a/docs/GLOSSARY.md
+++ b/docs/GLOSSARY.md
@@ -1,0 +1,11 @@
+## complete native application descriptor
+
+A complete native application descriptor is a string with format `[nativeAppName]:[platform]:[version]` 
+- `nativeAppName` : alphanumeric native application name , cannot contain character ':'
+- `platform` : one of `android` or `ios`
+- `version` : alphanumeric , cannot contain character ':'
+
+## partial native application descriptor
+
+A partial native application descriptor is a string with format `[nativeAppName]`, `plaform` and `version` can remain optional. 
+- `nativeAppName` : alphanumeric native application name , cannot contain character ':'
--- a/docs/cli/cauldron/add/nativeapp.md
+++ b/docs/cli/cauldron/add/nativeapp.md
@@ -1,6 +1,6 @@
 ## `ern cauldron add nativeapp <descriptor>`
 #### Description
-* Add a new native application version to the currently activated Cauldron  
+* Add a new native application version to the currently activated Cauldron.
 
 #### Syntax
 `ern cauldron add nativeapp <descriptor>`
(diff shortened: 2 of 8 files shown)


docs/GLOSSARY.md
```diff
@@ -0,0 +1,12 @@
+## complete native application descriptor
+
+The complete native application descriptor is a string with format `[nativeAppName]:[platform]:[version]`
```
belemaire: `A complete ...` rather than `The complete ...` (there is not a single one)

docs/GLOSSARY.md
```diff
@@ -0,0 +1,12 @@
+## complete native application descriptor
+
+The complete native application descriptor is a string with format `[nativeAppName]:[platform]:[version]` 
+- `nativeAppName` : alphanumeric native application name , cannot contain character ':'
+- `platform` : one of `android` or `ios`
+- `version` : alphanumeric , cannot contain character ':'
+
+## partial native application descriptor
+
+The complete native application descriptor is a string with format `[nativeAppName]:[platform]`
```
belemaire: Same remark as previous, but also fix `complete` (its `partial` section here), bad copy paste ;)

Instead of `The complete`, should be `A partial`

belemaire: Also partial can be only the native app name, i.e `[nativeAppName]`

docs/cli/cauldron/add/nativeapp.md
```diff
@@ -1,6 +1,6 @@
 ## `ern cauldron add nativeapp <descriptor>`
 #### Description
-* Add a new native application version to the currently activated Cauldron  
+* Add a new native application version to the currently activated Cauldron. The *complete native application descriptor* is a string with format `[nativeAppName]:[platform]:[version]`
```
belemaire: Please make this addition in the `remarks` section of the command, not in the description of the command.
i.e add a bullet in remarks where it says `The descriptor provided to this command should be a complete native application descriptor`. No need to add the format as it is linked to the glossary

docs/platform-parts/cauldron.md
```diff
@@ -135,7 +135,7 @@ $ ern cauldron add <native-app-descriptor>
 // for example
 // ern cauldron add nativeapp MyApp:ios:0.0.1
 ```
-
+The *complete native application descriptor* is a string with format `[nativeAppName]:[platform]:[version]`
```
belemaire: Same remark as before, no need to precise format here, its linked to the glossary, better to keep it centralized there.
Also no need to put it in italic, there is already format processing because of the auto linking with glossary that underlines it and put it in blue by default. we can put all glossary terms links in italic directly in the css styling for these links if we want to put them in italic (which I don't think is needed anyway, just a color difference is nice enough)

PR #183, opened 2017-10-12
Add support for generating composite MiniApp with extra JS dependencies

belemaire:
This PR add support to add extra JavaScript dependencies to the composite MiniApp. 
`start` command is now using this support through a new option `--extraJsDependencies`

This proves to be useful in case some dependencies of the MiniApps that are `dev` only, needs to be part of the Composite. For example, if a MiniApp is making use of `remote-redux-devtools` as a `devDependency`, the dependency won't be part of the Composite by default, due to the fact that only `production` dependencies are installed when generating a Composite. 

With this support, it is now possible to add the dependency to the Composite, if needed.

Example:

```bash
$ ern start \
-n testbinary:android:17.18.0 \
-p com.walmart.android \
-a app.main.MainActivity \
-e remote-redux-devtools@^0.5.7
```

### DIFF
--- a/ern-local-cli/src/lib/start.js
+++ b/ern-local-cli/src/lib/start.js
@@ -34,14 +34,16 @@ export default async function start ({
   watchNodeModules = [],
   packageName,
   activityName,
-  bundleId
+  bundleId,
+  extraJsDependencies
 } : {
   miniapps?: Array<string>,
   descriptor?: string,
   watchNodeModules: Array<string>,
   packageName?: string,
   activityName?: string,
-  bundleId?: string
+  bundleId?: string,
+  extraJsDependencies?: Array<DependencyPath>
 } = {}) {
   let miniAppsPaths: Array<DependencyPath> = _.map(miniapps, DependencyPath.fromString)
   let napDescriptor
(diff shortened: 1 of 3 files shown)


ern-local-cli/src/lib/start.js
```diff
@@ -34,14 +34,16 @@ export default async function start ({
...
+  extraJsDependencies
 } : {
   miniapps?: Array<string>,
   descriptor?: string,
   watchNodeModules: Array<string>,
   packageName?: string,
   activityName?: string,
-  bundleId?: string
+  bundleId?: string,
+  extraJsDependencies?: Array<DependencyPath>
```
deepueg: Pass Dependency instead of DependencyPath?

belemaire: well .. I guess we will need to review and potentially refactor our `Dependency` vs `DependencyPath` ... Because `DependencyPath` allows for NPM path / file / git path, whereas `Dependency` just basically boils down to `scope/name/version` (so no file / git support).

In the case of `extraJsDependencies`, file and git schemes can be provided for the dependencies, therefore choosing `DependencyPath` over `Dependency`.

PR #206, opened 2017-10-17
Promisify exec command

wbowlin:
The exec command was being wrapped in a promise with the same copy and paste error handling. Created a utility function to promisify the exec command. Used the promisified version of the code to make the exec command processing more concise.

### DIFF
--- a/ern-util/src/android.js
+++ b/ern-util/src/android.js
@@ -20,6 +20,18 @@ async function delay (ms: number) {
   })
 }
 
+function execp (cmd: string) {
+  return new Promise((resolve, reject) => {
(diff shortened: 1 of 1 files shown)


belemaire:
👍 Great, thank you !

We have some other code in other `ern` modules that are manually promisfying commands, this function might make its way to our `ern-util` module at some point I think.

Unfortunately, as much as I'd like to, I can't merge it as of now. As for many other open source project (react-native included), you'll have to sign a CLA first (for legal reason) before we can accept your code contributions ... We are currently working out to setup this process properly, to make it easier for people to easily contribute. Will let you know more by tomorrow.

Thanks !

Issue #103, opened 2017-09-28
Allow device configuration?


swashcap:
The manual intervention in `ern run-ios` can be kinda annoying, especially if a developer will be running this multiple times a day:

![config-me-dawg](https://user-images.githubusercontent.com/1858316/30994068-d8aa82f6-a466-11e7-8c11-5b973c0c681e.gif)

The boot process is otherwise automated. Is there support for putting a desired default device ID/string in configuration such that `ern run-ios` and `ern run-android` always boot the default simulator?

belemaire:
`ern run-android` and `ern run-ios` commands are pretty heavy in term of processing, they take a while to run, so ideally they shouldn't be run that often. Running these commands is only required whenever a new native dependency is added/removed/updated from the MiniApp. Otherwise, if only JS code is changing, the commands does not need to be run and a reload in the emulator/simulator is all what's needed. So we expect them not to be run many times during development.

For `ern run-android` if an emulator (or connected device) is running, the command won't ask user to select from a list of devices and will use the one launched. For `ern run-ios` this is unfortunately not the case -yet-.

If no emulator/device is running, we could indeed save the emulator choice from previous run and offer it as the default  one in the list, to make it easier. However we don't really want to get rid of the list, so it's difficult to know from there when we should show it v.s not show it. Maybe we could add an option flag (bool) to the command line to use the previously selected emulator by default (and not show the list) and if not passed, show the list.

swashcap:
+1 for the flag idea. That seems like a nice compromise.

PR #207, opened 2017-10-17
Allow device configuration

krunalsshah:
Fixes https://github.com/electrode-io/electrode-native/issues/103


belemaire:
:+1: Thanks Krunal

PR #212, opened 2017-10-19
Change container, runner to relative paths

swashcap:
* **Problem:** Running a linked ern fails due to hard-coded paths
* **Solution:** Change to package-relative paths
* **Testing:** Not sure as I don't think there's unit/integration tests for the altered modules. Try `ern run-ios` on a project?

### DIFF
--- a/ern-container-gen/src/generateContainer.js
+++ b/ern-container-gen/src/generateContainer.js
@@ -43,7 +43,7 @@ export default async function generateContainer ({
   const COMPOSITE_MINIAPP_DIRECTORY = path.join(workingDirectory, 'compositeMiniApp')
 
   const paths : ContainerGeneratorPaths = {
-    containerHull: path.join(Platform.currentPlatformVersionPath, 'ern-container-gen', 'hull', generator.platform),
+    containerHull: path.join(__dirname, '..', 'hull', generator.platform),
     containerTemplates: path.join(Platform.currentPlatformVersionPath, 'ern-container-gen', 'templates', generator.platform),
     compositeMiniApp: COMPOSITE_MINIAPP_DIRECTORY,
     pluginsDownloadDirectory: PLUGINS_DOWNLOAD_DIRECTORY,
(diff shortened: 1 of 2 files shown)


ern-runner-gen/src/index.js
```diff
@@ -169,7 +170,7 @@ export async function regenerateIosRunnerConfig (
     pathToElectrodeContainerXcodeProj: path.join(containerGenWorkingDir, 'out', 'ios')
   }
   const pathToRunnerConfig = path.join(pathToRunnerProject, 'ErnRunner/RunnerConfig.m')
-  shell.cp(`${platformPath}/ern-runner-gen/runner-hull/ios/ErnRunner/RunnerConfig.m`, pathToRunnerConfig)
+  shell.cp(path.join(runnerHullPath}, 'ios', 'ErnRunner', 'RunnerConfig.m', pathToRunnerConfig))
```
swashcap: :eyes:

ern-container-gen/src/generateContainer.js
```diff
@@ -43,7 +43,7 @@ export default async function generateContainer ({
   const COMPOSITE_MINIAPP_DIRECTORY = path.join(workingDirectory, 'compositeMiniApp')
 
   const paths : ContainerGeneratorPaths = {
-    containerHull: path.join(Platform.currentPlatformVersionPath, 'ern-container-gen', 'hull', generator.platform),
+    containerHull: path.join(__dirname, '..', 'hull', generator.platform),
     containerTemplates: path.join(Platform.currentPlatformVersionPath, 'ern-container-gen', 'templates', generator.platform),
```
belemaire: Could you please also update the path here the same way you did for `hull` directory, as the `templates` directory sits at the same level. To keep coherency. You can then also remove the `Platform` import as this line is the only usage left of it.

belemaire: Nevermind, I'll merge it and do this part of another PR, noticed some other cleanup that can be done following your refactoring and was working on a some cleanup work right now, so that will include that. Thanks again.

swashcap: :+1: Sounds good. Seeing some similar path-related problems with `ern run-android`. I can open another PR if needed.

PR #217, opened 2017-10-20
check if {apiName}-impl exists while create-api-impl

krunalsshah:
Problem : create-api-impl command checks if `{apiName}` exists in npm and prompts user 
Solution : check if {apiName}-impl exists while running create-api-impl and prompt user

### DIFF
--- a/ern-local-cli/src/commands/create-api-impl.js
+++ b/ern-local-cli/src/commands/create-api-impl.js
@@ -64,10 +64,12 @@ exports.handler = async function ({
   outputDirectory: string,
   hasConfig: boolean
 }) {
-  const isPackageNameInNpm = await utils.doesPackageExistInNpm(api)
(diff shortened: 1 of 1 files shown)


ern-local-cli/src/commands/create-api-impl.js
```diff
@@ -64,10 +64,12 @@ exports.handler = async function ({
   outputDirectory: string,
   hasConfig: boolean
 }) {
-  const isPackageNameInNpm = await utils.doesPackageExistInNpm(api)
+  // check if the packageName for specified {apiName}-impl exists
+  const apiImpl = api.concat('-impl')
```
deepueg: use ES6 template instead? `${api}-impl` ?

krunalsshah: Makes sense to use string interpolation

PR #220, opened 2017-10-20
Light refactor of NativeApplicationDescriptor class

belemaire:
- Remove needless use of `noop` function
- Remove `noop` function and exports from `ern-util`
- Add some guard clauses to NativeApplicationDesccriptor constructor
- Add unit tests for NativeApplicationDescriptor class


deepueg:
Thanks for the tests.!

PR #221, opened 2017-10-20
Container gen for api-impl

deepueg:
This PR adds support for generating api-impl initialization code inside a container.

### DIFF
--- a/ern-api-impl-gen/src/generators/ApiImplGen.js
+++ b/ern-api-impl-gen/src/generators/ApiImplGen.js
@@ -14,6 +14,7 @@ import {
 import _ from 'lodash'
 import chalk from 'chalk'
 import path from 'path'
+import fs from 'fs'
 import ApiImplMavenGenerator from './android/ApiImplMavenGenerator'
 import ApiImplGithubGenerator from './ios/ApiImplGithubGenerator'
 import { ApiImplGeneratable } from '../ApiImplGeneratable'
@@ -35,8 +36,9 @@ export default class ApiImplGen {
 
     await this.downloadApiAndDependencies(apiDependency, paths.pluginsDownloadDirectory, reactNativeVersion)
 
-    const schemaJson = path.join(paths.pluginsDownloadDirectory, `node_modules`, apiDependency.scopedName, `schema.json`)
+    const schemaJson = path.join(paths.pluginsDownloadDirectory, 'node_modules', apiDependency.scopedName, 'schema.json')
     const apis:Array<Object> = await ApiGenUtils.extractApiEventsAndRequests(schemaJson)
+    this.updatePackageJsonWithApiNames(paths.outDirectory, apis)
 
     const generators: Array<ApiImplGeneratable> = this.getGenerators(platforms)
     for (let generator of generators) {
@@ -108,6 +110,13 @@ export default class ApiImplGen {
           return new NullApiImplGenerator()
       }
     })
+  }
+
+  updatePackageJsonWithApiNames (outputDirectoryPath: string, apis: Array<Object>) {
+    const packageJsonPath = path.join(outputDirectoryPath, 'package.json')
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'))
+    packageJson.ern.containerGen.apiNames = _.map(apis, api => api.apiName)
+    fs.writeFileSync(packageJsonPath, JSON.stringify(packageJson, null, 2))
   }
 }
 
--- a/ern-api-impl-gen/src/generators/ApiImplGen.js
+++ b/ern-api-impl-gen/src/generators/ApiImplGen.js
@@ -78,25 +78,35 @@ export default class ApiImplGen {
     await spin(`Downloading ${dependency.toString()}`, yarn.add(dependency.path))
   }
 
-  async getDependencies (apiDependency: Dependency) : Promise<Array<Dependency>> {
+  async getDependencies (apiDependency: Dependency): Promise<Array<Dependency>> {
     try {
       log.info(`Looking for peerDependencies`)
       const apiPackageInfo = await yarn.info(apiDependency.path, {json: true})
 
-      let dependencies = []
+      let pluginsNames = []
+
       if (apiPackageInfo.data.peerDependencies) {
-        let pluginsNames = []
-        for (let dependency in apiPackageInfo.data.peerDependencies) {
-          pluginsNames.push(`${dependency}@${apiPackageInfo.data.peerDependencies[dependency]}`)
-        }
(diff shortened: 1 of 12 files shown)


ern-api-impl-gen/src/generators/ApiImplGen.js
```diff
@@ -102,6 +123,13 @@ export default class ApiImplGen {
       }
     })
   }
+
+  updatePackageJsonWithApiNames (outputDirectoryPath: string, apis: Array<Object>) {
+    const packageJsonPath = path.join(outputDirectoryPath, 'package.json')
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'))
+    packageJson.ern.containerGen.apiNames = _.map(apis, (api) => { return api.apiName })
```
belemaire: `_map` function call can be shortened to 
```javascript
 _.map(apis, api => api.apiName)
```
- If only one param the parenthesis are not needed
- If only one statement, curly braces should not be used and the arrow function will return the evaluated statement

ern-api-impl-gen/src/generators/ApiImplGen.js
```diff
@@ -69,27 +78,39 @@ export default class ApiImplGen {
...
-      return dependencies
+      return _.map(pluginsNames, Dependency.fromString)
     } catch (e) {
       throw new Error(`getDependencies: ${e}`)
     }
   }
 
+  pushDependencyNames (dependencies: Object, pluginsNames: Array<string>) : Array<string> {
+    for (const dependency in dependencies) {
+      if (dependencies.hasOwnProperty(dependency)) {
```
belemaire: This is a bit hard to grasp ;) This can be simplified using `Object.keys` function instead :

> The Object.keys() method returns an array of a given object's own enumerable properties, in the same order as that provided by a for...in loop (the difference being that a for-in loop enumerates properties in the prototype chain as well).

So your `for` loop can just become 
```javascript
for (const dependency of Object.keys(dependencies))
```

And you can get rid of the conditional.

ern-api-impl-gen/src/generators/ApiImplGen.js
```diff
@@ -31,11 +36,15 @@ export default class ApiImplGen {
 
     await this.downloadApiAndDependencies(apiDependency, paths.pluginsDownloadDirectory, reactNativeVersion)
 
+    const schemaJson = path.join(paths.pluginsDownloadDirectory, `node_modules`, apiDependency.scopedName, `schema.json`)
```
belemaire: Minor but let's use single quote instead of backtick when the string does not contain any placeholder (i.e just a string, not a template litteral). I already started doing that in some places.

ern-api-impl-gen/src/generators/android/ApiImplMavenGenerator.js
```diff
@@ -65,12 +65,12 @@ export default class ApiImplMavenGenerator implements ApiImplGeneratable {
       shell.cp(`-R`, path.join(paths.apiImplHull, 'android', '*'), outputDirectory)
 
       for (let plugin: Dependency of plugins) {
-        log.debug(`Copying ${plugin.name} to ${outputDirectory}`)
+        log.debug(`\nCopying ${plugin.name} to ${outputDirectory}`)
```
belemaire: Reason for `\n` here ? I guess there is one because you added it.
Looks odd. All log lines have a line return injected at the end by default, so I don't really see a reason.

ern-container-gen/src/generators/android/AndroidGenerator.js
```diff
@@ -231,6 +237,23 @@ export default class AndroidGenerator implements ContainerGenerator {
     }
   }
 
+  populateApiImplMustacheView (pluginSourcePath: string, mustacheView: Object) {
+    const packageJson = JSON.parse(fs.readFileSync(`${pluginSourcePath}/package.json`, 'utf-8'))
```
belemaire: Should use `path.join`to build the path here.

```javascript
path.join(pluginSourcePath, 'package.json')
```

ern-container-gen/src/generators/android/AndroidGenerator.js
```diff
@@ -231,6 +237,23 @@ export default class AndroidGenerator implements ContainerGenerator {
...
+    const containerGenConfig = packageJson.ern.containerGen
+    if (containerGenConfig && containerGenConfig.apiNames) {
+      mustacheView.apiImplementaions = mustacheView.apiImplementaions ? mustacheView.apiImplementaions : []
```
belemaire: typo `apiImplementaions` => `apiImplementations`

ern-container-gen/src/generators/android/AndroidGenerator.js
```diff
@@ -231,6 +237,23 @@ export default class AndroidGenerator implements ContainerGenerator {
...
+      for (const apiName of containerGenConfig.apiNames) {
+        let api = {}
+        api.apiName = apiName
+        api.hasConfig = containerGenConfig.hasConfig
+        api.apiVariableName = utils.camelize(apiName, true)
```
belemaire: Instead of creating an empty object and adding properties to it, just create the object directly. i.e :

```javascript
let api = {
  apiName,
  hasConfig: containerGenConfig.hasConfig,
  apiVariableName: utils.camelize(apiName, true)
}
```

You can even `push` the object directly instead of going through an intermediate `api` variable, up to you.

ern-api-impl-gen/src/generators/ApiImplGen.js
```diff
@@ -8,8 +8,13 @@ import {
 import {
   yarn
 } from 'ern-core'
+import {
```
krunalsshah: can go in 1 line

ern-api-impl-gen/src/generators/ApiImplGen.js
krunalsshah: `package.json` and `schema.json` can go as a Constant as done in `ern-api-gen` module, if it is used at 1 or more places, we can import as a contant.

ern-api-impl-gen/src/generators/android/ApiImplMavenGenerator.js
deepueg: This was placed for debugging, forgot to remove.!

ern-api-impl-gen/src/generators/ApiImplGen.js
deepueg: This is done dynamically, this dependency will only be downloaded at this point.

deepueg:
@belemaire @krunalsshah Thanks for the review. Addressed all comments.!

PR #225, opened 2017-10-24
Api impl regen

deepueg:
This PR adds a new command for regenerating an API implementation.

`ern regen-api-impl`


ern-api-impl-gen/src/ApiImpl.js
```diff
@@ -0,0 +1,144 @@
...
+  nativeOnly,
+  forceGenerate,
+  reactNativeVersion,
+  hasConfig = false,
+  paths
+} : {
+  apiDependency: Dependency,
+  outputDirectory: string,
+  nativeOnly: boolean,
+  forceGenerate: boolean,
```
belemaire: Looking in details at the implementation I can tell that `forceGenerate` is not required (by default it won't force generate).

However if I just look at this function object parameter I have no way to easily tell apart required vs optional properties (not from the object destructuring block itself nor from the flow typedef).  `flow` will show an error if `forceGenerate` is not provided by the caller.

You should typedef it as `forceGenerate?: boolean` instead, which makes it an optional object property. Same I think for `hasConfig`, you shouldn't use a `false` default, just type it as `hasConfig?: boolean` as well.  This way just by looking at the signature of the function (or looking at flow info when invoking the function) it's easy to distinguish required vs optional object properties.

ern-api-impl-gen/src/ApiImpl.js
```diff
@@ -0,0 +1,144 @@
...
+  let apiName = API_NAME_RE.exec(apiDependency.name)[1]
+  return outputDirectoryPath
+    ? path.join(outputDirectoryPath, `${apiName}-impl`)
+    : path.join(process.cwd(), `${apiName}-impl`)
+}
+
+function getPlatforms (nativeOnly: boolean): Array<string> {
+  return nativeOnly
+    ? [`android`, `ios`]
+    : [`js`]
```
belemaire: minor. 
Backticks instead of single quotes.

ern-api-impl-gen/src/ApiImpl.js
```diff
@@ -0,0 +1,144 @@
...
+}
+
+function ernifyPackageJson (outputDirectoryPath, nativeOnly, hasConfig) {
+  const packageJsonPath = path.join(outputDirectoryPath, 'package.json')
+  const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'))
+  const moduleType = nativeOnly ? `${ModuleTypes.NATIVE_API_IMPL}` : `${ModuleTypes.JS_API_IMPL}`
```
belemaire: no need to use template litteral here. 
`ModuleTypes.NATIVE_API_IMPL : ModuleTypes.JS_API_IMPL`
instead of
``` `${ModuleTypes.NATIVE_API_IMPL}` : `${ModuleTypes.JS_API_IMPL}` ```

File: ern-api-impl-gen/src/ApiImplRegen.js [hunk 4]
```diff
@@ -0,0 +1,24 @@
...
+    api: Dependency,
+    paths: Object,
+    reactNativeVersion: string,
+    platforms: Array<string>
+  }) {
+  try {
+    log.debug(`Entering regenerateApiImpl ${JSON.stringify(arguments[0])}`)
+    await new ApiImplGen().generateApiImplementation(api, paths, reactNativeVersion, platforms, true)
+  } catch (e) {
+    throw new Error(e)
```
belemaire: please just `throw(e)` to `rethrow` the error.
actually you dont need a `try/catch` block at all here as you don't do anything specific in it but you just retrhow.

ern-local-cli/src/commands/regen-api-impl.js
```diff
@@ -0,0 +1,121 @@
...
+    describe: 'a specific version of the api for which an implementation needs to be generated. \n The version should be higher than the version for which an implementation is already generated'
+  }).option('hasConfig', {
+    type: 'bool',
+    describe: 'Indicates if this api implementation requires some config during initialization. \nThis command will be stored and reused during container generation to enforce config initialization'
+  }).epilog(cliUtils.epilog(exports))
+}
+
+const ERROR_MSG_NOT_IN_IMPL = 'api-regen-impl can only be run inside an implementation project directory. \n Looks like it is ran elsewhere'
+const WORKING_DIRECTORY = path.join(Platform.rootDirectory, `api-impl-gen`)
+const PLUGIN_DIRECTORY = path.join(WORKING_DIRECTORY, `plugins`)
```
belemaire: minor, backticks again :S

ern-local-cli/src/commands/regen-api-impl.js
```diff
@@ -0,0 +1,121 @@
...
+
+exports.handler = async function ({
+                                    apiVersion,
```
belemaire: minor, indentation of these three lines (two spaces only)

ern-local-cli/src/commands/regen-api-impl.js
```diff
@@ -0,0 +1,121 @@
...
+        workingDirectory: WORKING_DIRECTORY,
+        outDirectory: process.cwd()
+      },
+      reactNativeVersion,
+      hasConfig,
+      platforms: getPlatforms()
+    })
+    log.info('\nSuccessfully regenerated api implementation!')
+  } catch (e) {
+    log.error(`${e} \n\nApiImpl regeneration failed.`)
```
belemaire: `\n\n`

ern-local-cli/src/commands/regen-api-impl.js
```diff
@@ -0,0 +1,121 @@
...
+        apiImplHull: path.join(Platform.currentPlatformVersionPath, 'ern-api-impl-gen', 'hull'),
+        pluginsDownloadDirectory: PLUGIN_DIRECTORY,
+        workingDirectory: WORKING_DIRECTORY,
+        outDirectory: process.cwd()
+      },
+      reactNativeVersion,
+      hasConfig,
+      platforms: getPlatforms()
+    })
+    log.info('\nSuccessfully regenerated api implementation!')
```
belemaire: `\n`

Issue #149, opened 2017-10-05
`ern regen-api` must clear the existing folders 


krunalsshah:
`ern regen-api` must clear the existing folders except `schema.json` before re generating the api's.

deepueg:
This was already done. There was an issue with IOS but that was fixed a while ago.

belemaire:
@krunalsshah can you confirm and if that's the case please close this issue. thanks !

krunalsshah:
@belemaire @deepueg 
Currently we delete the folders below.
```
shell.rm('-rf', path.join(outFolder, 'javascript'))
shell.rm('-rf', path.join(outFolder, 'swift'))
shell.rm('-rf', path.join(outFolder, 'android'))
shell.rm('-rf', path.join(outFolder, 'IOS'))
``` 

Which means that `package.json`, `yarn.lock` & `.flowconfig` are not deleted. For instance change to remove the `"prepublish": "ern regen-api -u same"`had to be made manually.

PR #232, opened 2017-10-25
ern regen-api must clear the existing files

krunalsshah:
Fixes https://github.com/electrode-io/electrode-native/issues/149

### DIFF
--- a/ern-api-gen/src/apigen.js
+++ b/ern-api-gen/src/apigen.js
@@ -1,6 +1,6 @@
 // @flow
 
-import generateProject, { generateSwagger, generateFlowConfig } from './generateProject'
+import generateProject, { generateSwagger, generateFlowConfig, generatePackageJson } from './generateProject'
 import normalizeConfig from './normalizeConfig'
 import fs from 'fs'
 import path from 'path'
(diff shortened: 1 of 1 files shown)


ern-api-gen/src/apigen.js
```diff
@@ -103,24 +85,40 @@ export async function regenerateCode (options: Object = {}) {
...
+    apiDescription : pkg.description,
+    apiAuthor : pkg.author,
+    bridgeVersion : options.bridgeVersion,
+    ...extra
+  }
+  // Regenerate package.json
+  fileUtils.writeFile(`${process.cwd()}/${PKG_FILE}`, generatePackageJson(pkgConfig))
+  // Regenerate .flowconfig file
+  fileUtils.writeFile(`${process.cwd()}/${FLOW_CONFIG_FILE}`, generateFlowConfig())
+
```
belemaire: For these two lines please make use of `path.join` instead of building the path manually, for better cross platform support. 

instead of 

```javascript
`${process.cwd()}/${PKG_FILE}` 
``` 

use 

```javascript
path.join(process.cwd(), PKG_FILE)
```

ern-api-gen/src/apigen.js
```diff
@@ -132,9 +130,11 @@ async function checkValid (message: string) {
   }
   return pkg
 }
+
 async function readPackage () {
   return fileUtils.readJSON(`${process.cwd()}/${PKG_FILE}`)
```
belemaire: Same remark as previous, please use `path.join` here (not a change you did part of this PR, but can be updated part of this PR)

ern-api-gen/src/apigen.js
krunalsshah: true. should have noticed that, that's how we are adapting code lately.

krunalsshah:
@belemaire done and thanks for the review

PR #233, opened 2017-10-25
Api implementation enhancements

deepueg:
1. Make copied library files read-only
2. Perform version check and display proper messages.
3. Minor cleanups.

### DIFF
--- a/ern-api-impl-gen/src/generators/android/ApiImplMavenGenerator.js
+++ b/ern-api-impl-gen/src/generators/android/ApiImplMavenGenerator.js
@@ -3,7 +3,8 @@
 import {
   Dependency,
   mustacheUtils,
-  shell
+  shell,
+  fileUtils
 } from 'ern-util'
 import {
   manifest,
@@ -23,7 +24,8 @@ type PluginConfig = {
 }
 
 export const ROOT_DIR = shell.pwd()
-
+const READ_EXECUTE = '555'
+const READ_WRITE_EXECUTE = '777'
 export default class ApiImplMavenGenerator implements ApiImplGeneratable {
   regenerateApiImpl:boolean
   get name (): string {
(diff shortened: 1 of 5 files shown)


ern-util/src/fileUtil.js
```diff
@@ -35,3 +38,15 @@ export async function writeFile (
...
   })
 }
+
+/**
+ * Recursively apply file mode for a given path
+ * @param fileMode
+ * @param path
+ * @returns {Promise.<void>}
+ */
+export async function chmodr (fileMode: string, path: string) {
```
belemaire: Could you please use `shelljs` library to `chmod` instead of running the command.
Otherwise running the platform directly is not cross platform and will fail on Windows.
https://github.com/shelljs/shelljs#chmodoptions-symbolic_mode-file
Make sure to import `shell` from `ern-util`. Do not import `shell` library directly

ern-api-impl-gen/src/generators/android/ApiImplMavenGenerator.js
```diff
@@ -61,36 +63,48 @@ export default class ApiImplMavenGenerator implements ApiImplGeneratable {
       shell.cd(ROOT_DIR)
 
       const outputDirectory = path.join(paths.outDirectory, `android`)
```
belemaire: Not a change from this PR but could you please replace backticks with single quotes here as part of this PR.

ern-util/src/fileUtil.js
belemaire: This function is not `async`, it does not return a `Promise`. 
`exec` returns a `ChildProcess` instance, not a `Promise`.
Would suggest just keeping it sync (after refactoring with `shelljs`) instead. `chmoding` is pretty quick anyway.

ern-api-impl-gen/src/generators/android/ApiImplMavenGenerator.js
```diff
@@ -124,6 +142,7 @@ export default class ApiImplMavenGenerator implements ApiImplGeneratable {
         }
         log.debug(`Api implementation files successfully generated for ${api.apiName}Api`)
       }
+      return {editableFiles}
```
belemaire: Is there a point in returning the result in an object considering it is a single return value ? 

Doing

```javascript
return editableFiles
```

and at the calling site :

```javascript
const editableFiles = await this.generateRequestHandlerClasses(apiDependency, paths, apis)
```

makes it a bit less clumpsy

deepueg:
@belemaire Review comments addressed.!

PR #235, opened 2017-10-25
Clean all dist directories as part of rebuild script

belemaire:
Electrode Native platform rebuild script was not properly cleaning up ern modules dist directories, resulting in some deleted src files in ern modules to remain in the dist directories.

The rebuild script will now remove all dist folders from ern modules before proceeding with lerna clean/bootstrap (that in part regenerates the dist folders).


krunalsshah:
Thanks, this would remove the ghost files that magically appear in the dist folder. 👍

PR #238, opened 2017-10-26
resolve create-api-impl incorrectly prompting for scoped api

krunalsshah:
* Refactor `import`
* Change places where string interpolation is not required
* resolve create-api-impl incorrectly prompting for scoped api
* Refactor `path.join`

### DIFF
--- a/ern-api-gen/src/apigen.js
+++ b/ern-api-gen/src/apigen.js
@@ -7,8 +7,7 @@ import path from 'path'
 import semver from 'semver'
 import {
   PKG_FILE,
-  FLOW_CONFIG_FILE,
-  FLOW_BIN_VERSION
+  FLOW_CONFIG_FILE
 } from './Constants'
 import {
   Dependency,
@@ -30,7 +29,7 @@ import {
 export async function generateApi (options: Object) {
   let config = normalizeConfig(options)
 
-  const outFolder = `${process.cwd()}/${config.moduleName}`
+  const outFolder = path.join(process.cwd(), config.moduleName)
   if (fs.existsSync(outFolder)) {
     log.error(`${outFolder} directory already exists`)
     process.exit(1)
--- a/ern-local-cli/src/commands/create-api-impl.js
+++ b/ern-local-cli/src/commands/create-api-impl.js
@@ -12,19 +12,14 @@ import {
   DependencyPath
 } from 'ern-util'
 import cliUtils from '../lib/utils'
-
 import inquirer from 'inquirer'
-
-const path = require('path')
+import path from 'path'
 
 exports.command = 'create-api-impl <api>'
 exports.desc = 'Commands to generate API implementation skeleton.'
 
 exports.builder = function (yargs: any) {
-  return yargs.option('api', {
-    type: 'string',
-    describe: `npm package name OR the git location of the api f OR the file path for which an implementation needs to be generated.`
-  }).option('nativeOnly', {
+  return yargs.option('nativeOnly', {
     alias: 'n',
     type: 'bool',
     describe: 'Generate native projects with proper dependencies (Implementation of the API has to be written in native'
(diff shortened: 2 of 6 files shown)


ern-local-cli/src/commands/create-api-impl.js
```diff
@@ -64,8 +62,12 @@ exports.handler = async function ({
   outputDirectory: string,
   hasConfig: boolean
 }) {
+  // react-native-name-api@0.0.3 should be created as react-native-name-api-impl@0.0.3
+  // @org/react-native-name-api@0.0.3 should be created as @org/react-native-name-api-impl@0.0.3
+  // react-native-name-api should be created as react-native-name-api-impl
+  const apiImpl = api.lastIndexOf('@') > 1 ? `${api}`.slice(0, api.lastIndexOf('@')) + '-impl@' + `${api}`.slice(api.lastIndexOf('@') + 1) : `${api}-impl`
```
krunalsshah: tried https://eslint.org/docs/rules/multiline-ternary
`standardjs` doesn't allow

deepueg: the version number of API or apiImpl may not be same, we should not append the api version to api-impl.

deepueg: Also, it would be best to compose a Dependency object here create impl name out of that ?
Api dependency can be moved up for reuse as below:
```
const apiDep:Dependency = Dependency.fromPath(DependencyPath.fromString(api))
const apiImpl = `${apiDep.name}-impl`

 await generateApiImpl({
      apiDependency: apiDep,
```

ern-api-gen/src/apigen.js
```diff
@@ -30,7 +29,7 @@ import {
 export async function generateApi (options: Object) {
   let config = normalizeConfig(options)
 
-  const outFolder = `${process.cwd()}/${config.moduleName}`
+  const outFolder = path.join(process.cwd(), `${config.moduleName}`)
```
deepueg: can change `${config.moduleName}` to `config.moduleName`

ern-local-cli/src/commands/create-api-impl.js
deepueg: Another thing we need to keep in mind is that a file path or git path can be passed as an API as well, using a dependency will be ideal in that case.

krunalsshah: Can you please provide the different arguments that can go with this command? As per slack the conversation `react-native-name-api@0.0.3` -> react-native-name-api-impl@0.0.3

krunalsshah: argument -> react-native-name-api@0.0.3 , prompt `react-native-name-api-impl` ??

krunalsshah: w.r.t `file path or git path `, wouldn't the `create-api` , `create-miniapp` get impacted too? Please elaborate. Thanks

ern-api-gen/src/apigen.js
krunalsshah: 👍

ern-local-cli/src/commands/create-api-impl.js
deepueg: ` react-native-name-api@0.0.3 , prompt react-native-name-api-impl ` Yes

It will not impact `create-api , create-miniapp` since they are not dependent on anything else, whereas an api-impl is dependent on an api. 

you could do the following,

`ern create-api-impl git@github.com:electrode-io/electrode-native.git`
`ern create-api-impl file://path`

This may not be supported currently but will be added in the future. 
https://github.com/electrode-io/electrode-native/blob/master/ern-util/src/Dependency.js#L52

krunalsshah:
@belemaire @deepueg Please take a look. I will issue few PR's

PR #240, opened 2017-10-26
Light refactoring around child_process exec/spawn

belemaire:


ern-core/src/MavenPublisher.js
```diff
@@ -40,23 +43,6 @@ export default class MavenPublisher implements Publisher {
...
-    return new Promise((resolve, reject) => {
-      exec(cmd,
-        (err, stdout, stderr) => {
-          if (err) {
-            log.error(err)
-            reject(err)
-          }
-          if (stderr) {
-            // Supress Lint warning as auto generated code isn't meant to be modified by the developer
-            stderr && stderr.includes('unchecked') ? log.warn('') : log.warn('\n' + stderr)
```
krunalsshah: so we lose some of the logging we did for specific conditions, and report `stderr`

PR #243, opened 2017-10-27
Light refactor around git

belemaire:
- Create a lightweight wrapper around `simple-git` lib in ern-util (`gitCli`)
- Use it in Cauldron to remove some duplication due to using `simple-git` directly
- Remove `GitUtils` class
- Replace `GitUtils` use with`gitCli` instead
- Remove `bluebird` flow type definition (bugged)
- Cleanup package.json of ern modules to remove dependencies on `simple-git` and `bluebird` now that they are only used in `ern-util`


belemaire:
flow checks are failing on the CI server but not locally.
figuring out why.
do not merge !

belemaire:
Oh damn !
It's because I was connected to internal walmart npm when doing a run rebuild, so most of the packages are pointing to `npme.walmart` repo.
So to keep in mind ... when modifying a yarn lock (add/update/remove an ern module package.json dependency) and npm run rebuild, always ensure that its done when connected to public npm repo.

belemaire:
Tests green. OK to proceed with review/merge.

flow-typed/npm/bluebird_v3.x.x.js
```diff
@@ -1,194 +0,0 @@
-// flow-typed signature: 4c1347a1bf7c69a614112e551403195f
```
krunalsshah: `"bluebird": "^3.5.0"` is part of `ern-util` package. Verifying if this file is intentionally deleted?

belemaire: @krunalsshah 
I just removed the flow type definitions of bluebird, not the dependency itself.
There is an invalid type def in the file leading to flow errors. Given that we make ultra limited use of bluebird it's ok if we don't have type definitions for it.

PR #245, opened 2017-10-27
Remove the options if specified as command arguments

krunalsshah:
Remove the options if specified as command arguments

### DIFF
--- a/docs/cli/cauldron/add/nativeapp.md
+++ b/docs/cli/cauldron/add/nativeapp.md
@@ -15,9 +15,14 @@
 * A complete native application descriptor representing the native application version to be added to the Cauldron.
 
 **Example**  
+
 `ern cauldron add nativeapp MyNativeApp:ios:17.14.0`  
 
-**Options**  
+**Options** 
+
+`--platformVersion/-v`
+
+* Use specified platform version
 
 `--copyFromVersion/-c <version>`
 
--- a/docs/cli/platform/config.md
+++ b/docs/cli/platform/config.md
@@ -6,7 +6,7 @@
 
 #### Syntax
 
-`ern platform config <key>`
+`ern platform config <key> [value]`
 
 **Arguments**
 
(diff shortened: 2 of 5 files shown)


docs/cli/platform/config.md
```diff
@@ -14,12 +14,6 @@
 
 * The key of the configuration element to get or set
 
-**Options**
-
-`--value/-v`
```
belemaire: This one is special because is a positional optional argument of the command.

We should not remove the doc for the value option all together.

Rather could you please update command syntax properly in this page (i.e `ern platform config <key> [value]`) and move `value` in arguments. But instead of `--value/-v` it should be `[value]`).

This way we don't lose the description in the doc of what is the value  and the syntax is the proper one.

ern-local-cli/src/commands/cauldron/add/nativeapp.js
```diff
@@ -17,10 +17,6 @@ exports.desc = 'Add a native application to the cauldron'
 
 exports.builder = function (yargs: any) {
   return yargs
-  .option('platformVersion', {
```
belemaire: This one should be kept as an option and not a positional argument.
Just remove `[platformVersion]` in the command syntax.

docs/cli/cauldron/add/nativeapp.md
```diff
@@ -14,6 +14,10 @@
 
 * A complete native application descriptor representing the native application version to be added to the Cauldron.
 
+`platformVersion`
```
belemaire: Move to options section instead (`--platformVersion/-p`) per my other comment on the command source.

ern-local-cli/src/commands/cauldron/add/nativeapp.js
krunalsshah: https://github.com/yargs/yargs/blob/master/docs/advanced.md#positional-arguments
Yup make sense

krunalsshah:
@belemaire thanks for the review, the comments are addressed.

PR #247, opened 2017-10-27
Redirect standard streams to process in auto-rebuild script

belemaire:


belemaire:
@krunalsshah 
Yes let me rewrite that more properly

belemaire:
@krunalsshah 
ok, done. this will redirect stdin/stdout and stderr to parent process.

PR #253, opened 2017-10-30
Improve system-tests

belemaire:
- Rename updatePlugin option in regen-api to skipVersion and update logic accordingly.
- Set process environment variable SYSTEM_TESTS during system tests execution. 
- AndroidGenerator is making use of this to disable Container maven publication during tests (as we are using a NodeJs environment on Travis which does not come with Android SDK). This is temporary.
- Fix ApiImplGitHubGenerator to avoid system test failure due to thrown error.
Improve system tests
- Update .travis.yml file to run system tests as a cron job (daily)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,11 @@ script:
 - npm run flow version
 - npm run flow
 - npm run test
+- 'if [ ${TRAVIS_EVENT_TYPE} != "cron" ]; then
+    npm install -g electrode-native && ern;
+    node setup-dev;
+    node system-tests;
+fi'
 notifications:
   slack:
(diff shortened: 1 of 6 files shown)


ernplatformtest:
- Rename `updatePlugin` option in `regen-api` to `skipVersion` and update logic accordingly.
- Set process environment variable `SYSTEM_TESTS` during system tests execution. AndroidGenerator is making use of this to disable Container maven publication during tests (as we are using a NodeJs environment on Travis which does not come with Android SDK). This is temporary.
- Fix `ApiImplGitHubGenerator` to avoid system test failure due to thrown error.
- Improve system tests
- Update `.travis.yml` file to run system tests as a cron job (daily)

Issue #265, opened 2017-10-31
create-api-impl proceeds with command execution before checking if api exists


krunalsshah:
`ern create-api-impl react-native-ernmovie-api-asjfdfjksfjdskfds -s`
`ern create-api-impl react-native-ernmovie-api-asjfdfjksfjdskfds`

proceeds with command execution, creates the folder and fails with error below.
```An error hapenned : Error: Command failed: /Users/k0s00ly/workspace/walmart/ern-platform/ern-core/node_modules/.bin/yarn add react-native-ernmovie-api-asjfdfjksfjdskfds --ignore-engines --exact
error Couldn't find package "react-native-ernmovie-api-asjfdfjksfjdskfds" on the "npm" registry.
```
Solution : Check if api exists before proceeding with the command

PR #266, opened 2017-10-31
check if api exisits before proceeding with create-impl

krunalsshah:
Fixes https://github.com/electrode-io/electrode-native/issues/265

### DIFF
--- a/ern-local-cli/src/commands/create-api-impl.js
+++ b/ern-local-cli/src/commands/create-api-impl.js
@@ -66,23 +66,27 @@ exports.handler = async function ({
   hasConfig: boolean,
   skipNpmCheck?: boolean
 }) {
-  const apiDep = Dependency.fromPath(DependencyPath.fromString(api))
-  const apiImplName = `${apiDep.name}-impl`
-
-  // Skip npm check code execution
-  if (!skipNpmCheck) {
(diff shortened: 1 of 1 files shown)


ern-local-cli/src/commands/create-api-impl.js
```diff
@@ -66,6 +66,12 @@ exports.handler = async function ({
   hasConfig: boolean,
   skipNpmCheck?: boolean
 }) {
+  // Fixes https://github.com/electrode-io/electrode-native/issues/265
+  if (!await cliUtils.doesPackageExistInNpm(api)) {
```
belemaire: Can you move this block in the `try` block down below, and throw an error instead of logging an error and returning ?

```javascript
throw new Error(`Couldn't find package ${api} to generate the api implementation`) 
```

This is the direction we'll favor and that I started taking in a recent PR. All commands when failing should log the error and exit the process with a non zero status code (to make it easier for running automated system tests is one of the reason, but also to keep things coherent and conventional). The catch block in the command will call the function that just does that.

belemaire:
Thanks, always good to add validity checks and fail early 👍

krunalsshah:
@belemaire thanks for the review. pushed the changes

PR #270, opened 2017-11-01
Add new validity check isValidNpmPackageName

belemaire:
- Add new check function `isValidNpmPackageName` to `Ensure` class and `logErrorAndExitIfNotSatisfied` function.
- Add tests

Will accordingly update CLI commands that can make use of this check in a subsequent PR.


gmbharath12:
wow 👍

PR #273, opened 2017-11-01
Add new validity check isValidElectrodeNativeModuleName

belemaire:


krunalsshah:
@belemaire we'd have to add this to the documentation glossary, if have not. thanks.

Issue #292, opened 2017-11-03
Daily system test build failure


belemaire:
Build failed for a legit reason. Still need to fix `system-tests` accordingly to reflect recent changes to `create-api`/`create api-impl` command 

```
Running ern create-api TestApi
[v1000.0.0] [Cauldron: cauldron-automation]
pwd
pwd
✖ TestApi is not a valid NPM package name
```

https://travis-ci.org/electrode-io/electrode-native/builds/296616836#L3543

belemaire:
We cannot trigger CI system tests build job manually for now. Whoever intends to fix will have to run the system tests locally through 

```console
node system-tests
```

to ensure that system tests are passing, post fix.

PR #294, opened 2017-11-03
Daily system test build failure

krunalsshah:
Fixes https://github.com/electrode-io/electrode-native/issues/292

### DIFF
--- a/system-tests.js
+++ b/system-tests.js
@@ -11,7 +11,7 @@ const gitHubCauldronRepositoryName = `cauldron-system-tests-${getRandomInt(0, 10
 const cauldronName = 'cauldron-automation'
 const miniAppName = 'MiniAppSystemTest'
 const miniAppPackageName = 'miniapp-system-test'
-const apiName = 'TestApi'
+const apiName = 'testapi'
 const nativeApplicationName = 'system-test-app'
 const nativeApplicationVersion = '1.0.0'
 const androidNativeApplicationDescriptor = `${nativeApplicationName}:android:${nativeApplicationVersion}`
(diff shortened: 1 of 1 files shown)


belemaire:
We have 4 different Electrode Native module types as of now (MINIAPP, API, JS-API-IMPL, NATIVE-API-IMPL). If we drop the distinction between JS/Native Api Implementation, we can say that we have three high level module types : `MiniApp`, `API` and `API implementation`.

We should stay coherent in term of these modules naming, not to confuse the user, to align commands used to create them, but also to simplify code base not doing things differently in term of naming for similar high level objects (electrode native modules). If things works the same way it makes it easier to share code.

An Electrode Native module should have two names :

- The name of the module itself (only composed of upper or lower case letters, validated by `isValidElectrodeNativeModuleName`)
- The package name of the module (should be a valid npm package name, validated by `isValidNpmPackageName`)

That's what is done for `create-miniapp`. 
This way the name of the module itself is used in the code of the module (for example for a MiniApp name this is what we use to generate the Activity class name, but it is also the name of the MiniApp itself used in the React JS component). And the package name is used to set the `package.json` name. We store the name of the MiniApp in the `ern` section of the package.json so that we can use it when we generate container or other needs. i.e :

```
"ern": {
    "version": "1000.0.0",
    "moduleType": "ern-miniapp",
    "miniAppName": "HelloWorld"
  }
```

Should do the same for all module types. 
We chose a bad naming for `miniAppName` property in `ern` object. Should have been `moduleName` instead. Ideally that's what all current module types (and future potential ones) should inject in the `ern` object kept in the package.json :

```
...
    "moduleType": "ern-miniapp",
    "moduleName": "HelloWorld"
  }
```

Can discuss it more with you.

krunalsshah:
@belemaire your explanation is precise :) If this is the way forward to go `should inject in the ern object kept in the package.json` , it surely makes sense to align the create commands.

PR #296, opened 2017-11-03
update file mode for files instead of directory

deepueg:


ern-api-impl-gen/src/generators/android/ApiImplMavenGenerator.js
```diff
@@ -89,19 +92,24 @@ export default class ApiImplMavenGenerator implements ApiImplGeneratable {
...
-  async updateFilePermissions (pluginOutputDirectory: string, editableFiles: Array<string>) {
+  async updateFilePermissions (srcOutputDirectory: string, editableFiles: Array<string>) {
     log.debug('Updating file permissions')
-    fileUtils.chmodr(READ_EXECUTE, pluginOutputDirectory)
-    for (const editableFile of editableFiles) {
-      fileUtils.chmodr(READ_WRITE_EXECUTE, editableFile)
+    const files = shell.find(srcOutputDirectory).filter((file) => file.endsWith('.java'))
+    for (const file of files) {
+      if (editableFiles.includes(file)) {
+        log.debug(`Filemode set to read_write_execute for: ${file}`)
```
belemaire: Let's remove this debug log. Will also remove some others to try not to pollute debug logging too much. This is more of a trace log level given the atomicity of the operation and because under the hood it calls a command on shell lib (chmod), the whole command will be logged at trace level already. 

This way you can also use ternary operator instead of `if`/`else` to gain some space

```js
editableFiles.includes(file)
  ? fileUtils.chmodr(READ_WRITE_EXECUTE, file)
  : fileUtils.chmodr(READ_EXECUTE, file)
```

PR #297, opened 2017-11-03
Fix isdep chck

deepueg:


krunalsshah:
commit 2518f27 👍 [approve]

Issue #260, opened 2017-10-31
Prevent API names to have special characters


belemaire:
api names should not have special characters since java does not support it.! Example error below for reference.

```
/Users/d0g00g4/.ern/containergen/out/android/lib/src/main/java/com/blah-delete/ern/api/WalmartItemApi.java:1: error: ';' expected
package com.blah-delete.ern.api;
                ^
/Users/d0g00g4/.ern/containergen/out/android/lib/src/main/java/com/blah-delete/ern/api/WalmartItemApi.java:10: error: ';' expected
import com.blah-delete.ern.model.Item;
               ^
/Users/d0g00g4/.ern/containergen/out/android/lib/src/main/java/com/blah-delete/ern/api/WalmartItemEvents.java:1: error: ';' expected
package com.blah-delete.ern.api;
                ^
/Users/d0g00g4/.ern/containergen/out/android/lib/src/main/java/com/blah-delete/ern/api/WalmartItemEvents.java:12: error: ';' expected
    import com.blah-delete.ern.model.Item;
                   ^
/Users/d0g00g4/.ern/containergen/out/android/lib/src/main/java/com/blah-delete/ern/api/WalmartItemRequests.java:1: error: ';' expected
package com.blah-delete.ern.api;
                ^
/Users/d0g00g4/.ern/containergen/out/android/lib/src/main/java/com/blah-delete/ern/api/WalmartItemRequests.java:12: error: ';' expected
import com.blah-delete.ern.model.Item;
               ^
/Users/d0g00g4/.ern/containergen/out/android/lib/src/main/java/com/blah-delete/ern/model/Item.java:1: error: ';' expected
package com.blah-delete.ern.model;
```

belemaire:
Can we distinguish between API name and package name if that's not already the case (doing kind of something similar we did for MiniApps where upper case letters are not allowed for package name but the way to go for the name of the api itself).
i.e if I name my API : MyCoolApi, the package name should be transformed to my-cool-api but the api name used for code generate will be MyCoolApi.

belemaire:
Assigning to @krunalsshah as this one will be closed by https://github.com/electrode-io/electrode-native/pull/300

PR #300, opened 2017-11-06
 create-api with package option, apiName is ElectrodeNativeModule

krunalsshah:
* create-api with package option, apiName is ElectrodeNativeModule
*  change the logic 
*  throw error in command to standardize logErrorAndExitProcess
*  adapt regen-api to work with package name 
* introduce `moduleName` for `apiName` in `ern` object 
*  update the documentation 
* update the system tests
Please note changes are backward compatible with existing api's


system-tests.js
```diff
@@ -125,7 +125,7 @@ process.chdir(workingDirectoryPath)
 
 // api
 run(`ern create-api ${apiName} --skipNpmCheck`)
-const apiPath = path.join(process.cwd(), `react-native-${apiName}-api`)
+const apiPath = path.join(process.cwd(), `${apiName}`)
```
belemaire: minor : can remove the template literal string here and just use the variable directly

ern-local-cli/src/commands/regen-api.js
```diff
@@ -39,7 +39,7 @@ exports.handler = async function ({
     if (!bridgeVersion) {
       const bridgeDep = await manifest.getNativeDependency(Dependency.fromString('react-native-electrode-bridge'))
       if (!bridgeDep) {
-        return log.error(`react-native-electrode-bridge not found in manifest. please provide explicit version`)
+        throw new Error(`react-native-electrode-bridge not found in manifest. please provide explicit version`)
```
belemaire: minor : single quotes

ern-local-cli/src/commands/create-api.js
```diff
@@ -38,54 +43,63 @@ exports.builder = function (yargs: any) {
...
+  try {
+    await utils.logErrorAndExitIfNotSatisfied({
+      isValidElectrodeNativeModuleName: {
+        name: apiName
+      }
+    })
+
+    // Construct the package name
+    if (!packageName) {
+      packageName = core.getDefaultPackageNameForModule(apiName, ModuleTypes.API)
```
belemaire: Could you reuse the same logic as in `create-miniapp`
i.e, if the `packageName` is not explicitly provided as a command argument, the default generated name is just the default suggestion, we still prompt the user if he/she wants to use this name or a different one

https://github.com/electrode-io/electrode-native/blob/master/ern-local-cli/src/commands/create-miniapp.js#L62

You can also copy/paste the `promptForPackageName` function (just replace `MiniApp` by `API` in the message string). 

The same way will be done for create-api-impl so that the create commands are properly aligned and consistent.

As they will share most of the same logic it makes it easier to refactor (regroup logic) later on as well.

ern-api-gen/src/normalizeConfig.js
```diff
@@ -124,11 +126,15 @@ export default function normalizeConfig ({
   if (apiSchemaPath) {
     config.apiSchemaPath = apiSchemaPath
   }
-  if (!artifactId) {
-    config.artifactId = config.moduleName
+  if (!config.artifactId) {
+    config.artifactId = `react-native-${simpleName}-api`
```
belemaire: Not sure what `artifactId` is, so just double checking if that's ok for it to have the old naming convention.

belemaire:
Closes https://github.com/electrode-io/electrode-native/issues/260

ern-api-gen/src/normalizeConfig.js
krunalsshah: @belemaire I couldn't find any reference to artifactId usage , except in internal swagger `js` file.

ern-local-cli/src/commands/regen-api.js
krunalsshah: oops

ern-local-cli/src/commands/create-api.js
krunalsshah: @belemaire by design for create-api and create-api-impl I thought it's better to avoid prompt  for default package name. [documentation states the default name convention]
this way we avoid adding additional option to avoid CI prompting etc. It's generated code IMHO and we also have introduce a flag if user wishes to override default name for package. Thoughts please?

krunalsshah: nvm . ignore the comment. I think we want to align the logic across so makes sense.

krunalsshah:
@belemaire the review comments have been addressed.

PR #311, opened 2017-11-07
create-api-impl to accepts apiImplName as argument

krunalsshah:
* create-api-impl to accepts apiImplName as argument, apiImplName is ElectrodeNativeModule
* change the logic of the command
* modmoduleName to ern.containerGen
* update the documentation
* update the system tests


docs/cli/create-api-impl.md
```diff
@@ -7,14 +7,18 @@
...
+`ern create-api-impl <apiName> [apiImplName]`  
 
 **Arguments**
 
 `<apiName>`
 
 * The package descriptor of the API for which to create an API implementation project
 
+`[apiImplName]`
+* The name of the API IMPL project.  API IMPL name must contain only alphabets.
```
belemaire: Here and in the whole page please replace `API IMPL` by `API implementation`.

belemaire: `API Impl name must contain only alphabets` => `The API implementation name must only contain upper and/or lower case letters`.

As discussed, might create at some point an entry (not part of this PR) in the glossary for `Electrode Native module name` (tbd) and clarify it there.

krunalsshah:
👍 for the review. Review comments have been addressed. Thanks

PR #313, opened 2017-11-07
add options to the regen-api command documentation

krunalsshah:
add options to the regen-api command documentation

### DIFF
--- a/docs/cli/regen-api.md
+++ b/docs/cli/regen-api.md
@@ -8,7 +8,16 @@
 
 `ern regen-api`  
 
+**Options** 
+`--skipVersion/-s`
+* Skip the step to update api version and publish to NPM.
+
+`--bridgeVersion/-b`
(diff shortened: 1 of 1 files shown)


docs/cli/regen-api.md
```diff
@@ -8,6 +8,13 @@
 
 `ern regen-api`  
 
+**Options** 
+`--skipVersion/-s`
+* Skip the step to update api version and publish to NPM.
+
+`--bridgeVersion/-b`
+* Version of [ElectrodeNative](https://github.com/electrode-io/react-native-electrode-bridge) bridge to use.
```
belemaire: Shouldn't it rather be :

Version of the [Electrode Native Bridge](https://github.com/electrode-io/react-native-electrode-bridge) to use
Also let's try not to use inline urls but rather place them at the end of the page (can then be reused throughout the page but also improve readability of raw markdown.

PR #314, opened 2017-11-07
Remove unused configFilePath from normalizeConfig

krunalsshah:
Remove unused configFilePath from normalizeConfig


krunalsshah:
@belemaire I don't recollect using `apigen.conf.json` anymore.

belemaire:
yes. looked at it. might be of use but not using it right now (which is not a solid reason to remove if that's the only one). But I don't feel it's the right place for it anyway, would need a bit of refactoring. Removing it for now is ok to make `normalizeConfig` code leaner.

Issue #317, opened 2017-11-07
Unsupported plugin. No configuration found in manifest for react-native@0.50.1


jmatsushita:
Hi there,

I'm giving a try at the [Getting Started instructions for electrode-native](https://electrode.gitbooks.io/electrode-native/content/getting-started/getting-started.html) but bumping into an `Unupported plugin` error. 

Please find below the execution log of `ern run-android` which yields this error with the `DEBUG=*`
env variable.

```
~/electrode-native/MovieListMiniApp > DEBUG=* ern run-android
 ___ _        _               _       _  _      _   _
| __| |___ __| |_ _ _ ___  __| |___  | \| |__ _| |_(_)_ _____
| _|| / -_) _|  _| '_/ _ \/ _` / -_) | .` / _` |  _| \ V / -_)
|___|_\___\__|\__|_| \___/\__,_\___| |_|\_\__,_|\__|_|\_/\___|
[v0.10.0] [Cauldron: -NONE-]

A React Native Packager is already running in a different process
⠋ Creating local Container and publishing AAR to maven local  simple-git [ 'remote',
  simple-git   'set-url',
  simple-git   'origin',
  simple-git   'https://github.com/electrode-io/electrode-native-manifest.git' ] +0ms
  simple-git [ 'fetch', 'origin', 'master' ] +22ms
⠦ Generating Container locally  simple-git [ 'reset', '--hard', 'origin/master' ] +536ms
  simple-git [ 'pull', 'origin', 'master' ] +18ms
⠧ Generating Container locally[fillContainerHull] Something went wrong: Error: Unsupported plugin. No configuration found in manifest for react-native@0.50.1
[generateContainer] Something went wrong. Aborting Container Generation
✖ Unsupported plugin. No configuration found in manifest for react-native@0.50.1
runLocalContainerGen failed: Error: Unsupported plugin. No configuration found in manifest for react-native@0.50.1
An error hapenned : Unsupported plugin. No configuration found in manifest for react-native@0.50.1
```

belemaire:
@jmatsushita Thanks for reporting this one. It looks like we have a regression in new version 0.10.0 that we just released. We have received similar internal reports.
We are looking into it now.

belemaire:
@jmatsushita This has been fixed in #319
Will release patch version `0.10.1` of platform once PR is merged.

belemaire:
@jmatsushita 
Thanks again for reporting :+1:
[...]

PR #319, opened 2017-11-07
Fix manifest platform version filtering

belemaire:
Fixes https://github.com/electrode-io/electrode-native/issues/317

Fix a bug that has already been there but only surfaced with platform version `0.10.0`.
Code was working fine for previous versions `v0.8.0`, `v0.9.0` or even `v1000.0.0` as the version comparison statement was properly evaluated to `true` for the following :

```js
'0.5.0' <= '0.9.0'  // true
'0.5.0' <= '1000.0.0' // true
```

However for `v0.10.0` the string comparison was returning false, leading to the bug.

```js
'0.5.0' <= '0.10.0' // false
```

**Fix :** Do not use string comparison when comparing versions but rather rely on `semver` lib that has proper logic to compare versions.


belemaire:
This is a critical issue. Will release `0.10.1` to include this fix.

gmbharath12:
👍

PR #320, opened 2017-11-07
Internalize imports

swashcap:
* **Problem:** Debugging `ern` CLI commands locally via symlink-ed packages is difficult as certain packages `require` themselves. 
* **Solution:** Change `import`s to reference the modules directly
* **Testing:**
    * Set up for debug development

        ```shell
        ern platform use 0.10.0
        cd <development electrode-native dir>
        npm link
        cd ~/.ern/versions/
        rm -rf 0.10.0/node_modules
        ln -s <development electrode-native dir> node_modules
        ```

        Running `ern --help` should display help output and not explode 💥
    * CI should pass?

### DIFF
--- a/ern-core/src/iosUtil.js
+++ b/ern-core/src/iosUtil.js
@@ -3,11 +3,9 @@ import {
   mustacheUtils,
   shell
 } from 'ern-util'
-import {
-  manifest,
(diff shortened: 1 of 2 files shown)


belemaire:
Thanks ! Missed those.

Issue #315, opened 2017-11-07
Prompt User to suffix relevant moduleName 


krunalsshah:
Inspiration : 
Just re-did the Getting Started guide and when I ran ern create-miniapp MovieListMiniApp it propose a package name as : movie-list-mini-app-miniapp

Recommendation 
If the module name that the user provide to command, is not suffixed by Api, suggest the user to suffix the Api through a prompt.
i.e, something like

$ ern create-api Weather
> To follow naming conventions we suggest suffixing the name of your API with Api
> Do you want to name the Api as WeatherApi rather than Weather ? Y/n

Commands Impacted : `create-miniapp`, `create-api` and `create-api-impl`

PR #323, opened 2017-11-08
Prompt User to use suffix for moduleName

krunalsshah:
Fixes https://github.com/electrode-io/electrode-native/issues/315


ern-local-cli/src/lib/utils.js
```diff
@@ -635,6 +636,62 @@ async function performPkgNameConflictCheck (name: string) : Promise<boolean> {
   return true // If package name doesn't exist continue with command execution
 }
 
+function checkIfModuleSuffixExists (moduleName: string, moduleType: string): boolean {
```
belemaire: This function does not check if the module suffix exists. It checks if the module name contains the suffix, but can be in any location of the module name. 
Should fix the function so that it does what it says it does ;)
You can achieve it with some string manipulation, but probably easier through a regex.

ern-local-cli/src/lib/utils.js
```diff
@@ -635,6 +636,62 @@ async function performPkgNameConflictCheck (name: string) : Promise<boolean> {
...
+      case ModuleTypes.JS_API_IMPL:
+        suffixedModuleName = `${moduleName}Impl`
+        message = `We recommend suffixing the name of ${moduleName} with Impl, Do you want to use ${suffixedModuleName}?`
+        break
+      case ModuleTypes.NATIVE_API_IMPL:
+        suffixedModuleName = `${moduleName}Impl`
+        message = `We recommend suffixing the name of ${moduleName} with Impl, Do you want to use ${suffixedModuleName}?`
+        break
+      default:
+        // do nothing
```
belemaire: throw an error instead of doing nothing. if a non recognized moduleType is provided to this function, it should fails instead of silently doing nothing.

krunalsshah:
@belemaire thanks for review . Will make the changes.

krunalsshah:
@belemaire
This is coming in a new PR i have already worked on.  
- Also let's rework the `getDefaultPackageNameForModule` function as part of this PR, because it does not check if the suffix already exists or not, so it will duplicate it for the package name if it exists.

```
 case ModuleTypes.MINIAPP:
      return basePackageName.concat('-miniapp')
    case ModuleTypes.API:
      return basePackageName.concat('-api')
    case ModuleTypes.JS_API_IMPL:
      return basePackageName.concat('-js-api-impl')
    case ModuleTypes.NATIVE_API_IMPL:
      return basePackageName.concat('-native-api-impl')
```

ern-local-cli/src/lib/utils.js
krunalsshah: @belemaire The logic already handles logic, i'll rename the function instead. Thanks

krunalsshah:
@belemaire review comments are addressed. thanks

PR #325, opened 2017-11-08
fix(ern-util): remove false positive in iOS device parsing

swashcap:
#### Problem

The device chooser doesn't work when running `ern run-ios`:

host-device-is-there

It turns out that [the device check](https://github.com/electrode-io/electrode-native/blob/2f4cc50a5b3da81c324efd2fc31340e51382a706/ern-util/src/ios.js#L27-L30) now returns the host computer’s `ComputerName`! Very odd, this must be new behavior. Passing the host’s udid causes Xcode to fail: the computer itself isn't a valid iOS build target.

#### Solution

Add a check for the `ComputerName` in `parseIOSDevicesList`. Add some unit tests.

#### Testing

???

(See added unit tests.)

### DIFF
--- a/ern-util/src/ios.js
+++ b/ern-util/src/ios.js
@@ -26,7 +26,8 @@ export async function getiPhoneSimulators () {
 
 export function getiPhoneRealDevices () {
   const devices = execSync('xcrun instruments -s', {encoding: 'utf8'})
-  return parseIOSDevicesList(devices)
+  const computerName = execSync('scutil --get HostName', {encoding: 'utf8'})
+  return parseIOSDevicesList(devices, computerName)
 }
 
 export async function askUserToSelectAniPhoneDevice (devices: Array<IosDevice>) {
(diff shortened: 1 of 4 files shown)


belemaire:
Nice, thanks for the tests !

@clairewli LGTM but could you please have a look and approve/rebase, I know you did change this code recently (which might have introduced this issue).

ern-util/src/ios.js
```diff
@@ -26,7 +26,8 @@ export async function getiPhoneSimulators () {
 
 export function getiPhoneRealDevices () {
   const devices = execSync('xcrun instruments -s', {encoding: 'utf8'})
-  return parseIOSDevicesList(devices)
+  const computerName = execSync('scutil --get HostName', {encoding: 'utf8'})
```
swashcap: This `child_process` call will fail on non-macOS machines—I’m thinking CI boxes. Any ideas on how we can address?

ernplatformtest: `ios.js` is as of now only used to support Runner for launching MiniApp(s) on a iOS device or simulator through `run-ios` command. The `run-ios` command is disabled for non macOs machine (there is a platform check). 

Doesn't make a lot of sense for CI to use `run-ios` command. In the context of system testing it might have to ideally, but we know this is a limitation for now and we can't system test `run-ios` on CI on a linux box. For unit tests we're good though, `child_process` calls can be mocked with a little effort based on the needs of the test.

Issue #303, opened 2017-11-06
code-push command should support semver range


belemaire:
`code-push` command should accept a semver to target one or more versions (as Microsoft `code-push` command already offers). 

We should apply the same exact rules as the Microsoft `code-push` (considering `ern` ultimately forwards the command to Microsoft `code-push` client for execution).

If the semver used is a range that targets multiple version, `ern` should properly perform multiple `code-push` updates for each matching version and update the Cauldron accordingly.

devpascoe:
Following up on this one i noticed if i specify ern code-push release --targetBinaryVersion ^1.0 i get the target version under app center displaying the correct ^1.0 value.
But if i specify ern code-push release --targetBinaryVersion * i am unable to build in order to release. So it looks like the * wildcard does not work.

Issue #302, opened 2017-11-06
Store more CodePush related information in Cauldron


belemaire:
We currently store a minimal amount of information pertaining to CodePush updates in the Cauldron (just the list of MiniApps that were CodePushed). 

We need to keep more information in addition to the MiniApps names/versions :

- The flags / options that were used to push the update
- The deployment key that was used (one of the parameters to CodePush command). ex: Production, QA, Staging ...

For the CodePush execution logic we also need to distinguish between the different deployment keys and use the proper yarn.lock from the Cauldron (right now we have a single yarn.lock which is not right when different envs are used).

Should figure out the proper object schema and conduct the implementation.

krunalsshah:
This is a very nice extension to the code push command support

PR #331, opened 2017-11-13
CodePush refactoring

belemaire:
This pull request marks the first step in improving CodePush integration in Electrode Native.

This step is actually a small refactor that only changes the way CodePush updates are pushed to the server.

Instead of using the CodePush CLI (command line client), this refactor is now relying on the CodePush management SDK instead (CodePush CLI is anyway delegating command execution to the CodePush management SDK).

This removes an extra step (CLI -> Management SDK), and make the integration more proper as we don't have to deal with the CodePush command line CLI and parse output or what else. We just directly call in the management SDK programatically.

Also this opens the door for next improvements, respectively https://github.com/electrode-io/electrode-native/issues/302 and https://github.com/electrode-io/electrode-native/issues/303 

Once CodePush PR https://github.com/Microsoft/code-push/pull/533 is merged by the CodePush team, it will make it easy to retrieve the label and other information associated to the release, to store in the Cauldron. It will also make it easier to integrate the CodePush promotion flow in Electrode Native.

This Pull Request also introduce tests for a new introduced function `getCodePushAccessKey`.
These tests are making use of the [mock-fs](https://github.com/tschaub/mock-fs) mocking library, introduced as a devDependency part of this PR.
This library allows for easy mocking of the file system, creating a fake in memory file system, so that all files/directories can be mocked easily.


krunalsshah:
👍 thanks

PR #332, opened 2017-11-14
unit test android ern-util

krunalsshah:
unit test android `ern-util`


krunalsshah:
will adapt to `mock-fs` in future PR

Issue #362, opened 2017-11-20
previousEmulatorConfig conflict between ios and android.


deepueg:
The current implementation stores the emulator config for ios and android under the same key causing an error to happen when only one platform value is stored under the key. 

For example: when `run-ios` command is executed the `emulatorConfig` value is updated with ios emulator. A subsequent execution of `run-android` fetches `emulatorConfig` value and it gets the stored value of ios causing the command to fail. 

Solution: Store the config for ios and android under different keys to avoid any confusion and/or easy future maintenance.

PR #363, opened 2017-11-20
Fix emulator config storage

deepueg:
fixes https://github.com/electrode-io/electrode-native/issues/362


krunalsshah:
Fixes a specific issue? #363? It tags the PR itself?

deepueg:
Sorry, autofill issue. updated .

krunalsshah:
What is the benefit of hosting the emulatorConfig per platform? By having it contained in the `emulatorConfig`, and then having `android` and `ios` as inner object we are already implying it's specific to the platform.

krunalsshah:
I believe the tests will fail, as emulatorConfig is consumed at multiple places in the platform code.

PR #387, opened 2017-11-27
force ssh git url for cauldron

deepueg:


belemaire:
https://github.com/electrode-io/electrode-native/issues/389

belemaire:
let’s work on the fix instead …
otherwise `system-tests` are going to break.
https://github.com/electrode-io/electrode-native/blob/master/system-tests.js#L76
we’re adding the repo url as an https url (we provide username/password directly in the url)
cannot use ssh for system tests, as we cannot store an ssh key on the travis box if we’re not using a paid account :disappointed:

PR #407, opened 2017-12-05
fix android crash if Dev Settings is clicked on  Dev menu

krunalsshah:
fix android crash if Dev Settings is clicked on  Dev menu

### DIFF
--- a/ern-container-gen/hull/android/lib/src/debug/AndroidManifest.xml
+++ b/ern-container-gen/hull/android/lib/src/debug/AndroidManifest.xml
@@ -1,0 +1,6 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+          package="com.walmartlabs.ern.container">
+    <application>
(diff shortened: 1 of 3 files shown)


alexrwegener:
Can this be moved to the debug manifest so that it's not included in a release build? 

Per https://facebook.github.io/react-native/docs/integration-with-existing-apps.html
```
If you need to access to the DevSettingsActivity add to your AndroidManifest.xml:

<activity android:name="com.facebook.react.devsupport.DevSettingsActivity" />

This is only really used in dev mode when reloading JavaScript from the development server, so you can strip this in release builds if you need to.
```

krunalsshah:
Thanks @alexrwegener for the review. Absolutely makes sense to move `DevSettingsActivity` debug manifest during container generation. Necessary changes are included. 
@belemaire

alexrwegener:
Looks awesome @krunalsshah! Thank you! 👏

PR #409, opened 2017-12-05
Upgrade code-push version

belemaire:
Bump code-push version to [2.1.4](https://github.com/Microsoft/code-push/releases/tag/v2.1.4) to get access to the proper `promote` function implementation.

### DIFF
--- a/ern-util/package.json
+++ b/ern-util/package.json
@@ -50,7 +50,7 @@
   "dependencies": {
     "bluebird": "^3.5.0",
     "chalk": "^2.3.0",
-    "code-push": "^2.0.3-beta",
+    "code-push": "^2.0.4",
     "console-log-level": "^1.4.0",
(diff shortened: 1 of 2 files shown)


ern-util/package.json
```diff
@@ -50,7 +50,7 @@
   "dependencies": {
     "bluebird": "^3.5.0",
     "chalk": "^2.3.0",
-    "code-push": "^2.0.3-beta",
+    "code-push": "^2.0.4",
```
krunalsshah: wondering if we can move to v2.1.4?

belemaire: There is no `v2.1.4` I think they did a mistake and meant `2.0.4`

belemaire: After verification. 2.1.4 is the version of the CLI, not the SDK ... That's confusing ;P

belemaire: So we're all good.

Issue #398, opened 2017-11-29
Add method to list native apps in cauldron


ghostatron:
Currently, we have "ern cauldron get nativeapp <descriptor>".  It would be very helpful to be able to leave the descriptor empty and have the output be a list of available nativeapps/versions.  It could be a simple listing that is informational, or it could be the menu selector feature that is used in a few other commands.

PR #413, opened 2017-12-05
Improve cauldron get nativeapp command

belemaire:
Closes https://github.com/electrode-io/electrode-native/issues/398


krunalsshah:
Do we need to adapt system tests?

belemaire:
@krunal done / rebased

PR #422, opened 2017-12-12
update android build tools 

krunalsshah:
Modules impacted : 
`ern-api-gen` 
`ern-api-impl-gen` 
`ern-container-gen`   
`ern-runner-gen`

Bump Android Gradle Plugin version from 2.3.3 to 3.0.1
Bump Gradle version from 3.5 to 4.1


belemaire:
Any reason to bump the minSdk version to 19 ? Is there any code we added in the API generatedc ode that is not compatible with 16 ? If not let's stick to 16, otherwise applications that target 16 (JellyBean and above ... JellyBean is still in use) would not be able to add a Container and we need to state that Electrode Native is only supported on KitKat and above. Which is a greater restriction than react-native which still supports SDK 16+ (https://github.com/facebook/react-native/blob/master/ReactAndroid/build.gradle#L238).

krunalsshah:
@belemaire thanks for review. Updates have been committed & squashed to the last commit.

Issue #430, opened 2017-12-13
findNativeDependencies lookup improvement


belemaire:
Starting point : https://github.com/electrode-io/electrode-native/issues/428

What should be done :

- Remove `react-native` prefix filtering from`findNativeDependencies` lookup algorithm.
- If a detected third party native module does not have a Manifest configuration it can either be :
  - A false positive : This is not really a native module, and should not be injected in the Container, so missing Manifest configuration is legit.
  - A true positive : This is a native module, it should be injected in the Container and therefore should have an associated injection configuration in the Manifest.

In any case, instead of erroring during Container generation as of now, just log a warning to the user.

Later on we will probably improve this by maintaining a list of false positives in the Manifest.

Issue #428, opened 2017-12-13
NPM dependencies containing native code must be prefixed with 'react-native-' in order for Manifest config to be processed


abdyer:
When specifying the Manifest config/plugins for [AppCenter](https://github.com/Microsoft/AppCenter-SDK-React-Native), it took me awhile to realize that it was being ignored because the various AppCenter dependencies (`appcenter`, `appcenter-analytics`, `appcenter-tracking`) do not start with `react-native-` ([source code here](https://github.com/electrode-io/electrode-native/blob/0928a6dccc93aac099085ae5ac7976662baf0c40/ern-util/src/findNativeDependencies.js#L24)). The ugly workaround is to simply publish these to an internal NPM registry with the expected prefix, but a cleaner way to do this would be greatly appreciated.

Even still, this is a minor concession for using such a great platform for integrating React Native. Thanks again for the great work!

belemaire:
Thanks @abdyer for the feedback !

Actually, we just ran into a similar issue yesterday when someone internal created an Electrode Native API which package name was not prefixed with `react-native`. It was leading to the API not being injected in the Container during generation, because of the exact same line of code you are pointing to.

We're going a different road for this one though as we will impose the convention for API naming to follow `react-native-[API_NAME]-api`. This will be enforced by the `create-api` commands.

Prefixing react native modules that contain native code with `react-native` is a convention that is widely adopted, and for our APIs we want to enforce it as well (as well as the -api suffix) so that we can easily identify api modules just by looking at their name.

Unfortunately, we cannot control third party native modules naming and cannot expect that they will follow the react native common naming convention. As you can see in the code however, we just assume that the convention is followed. It was not creating any issue on our side and we forgot a bit about this line of code, as all the third party native modules that we are using internally up until now are all properly following the naming convention.

Historically, the problem we faced at one point (and the reason why we added this prefix check) was that a team was using a JS dependency in their MiniApp that was containing some native code inside (maybe appium/selenium testing tools). Because it was containing some native code, the `findNativeDependencies` lookup algorithm was flagging it as a native module and was then failing Container generation because no configuration was found for it in the Manifest on how to inject it in the Container. But it was a false positive because it was not a react native native module and we didn't want to inject this native code in the Container ... so we took the easy path at this time of assuming that native modules are starting with `react-native` prefix, fixing this issue.

The simplicity of this approach is biting as us now, as you just experienced ;)

[...]

Issue #426, opened 2017-12-13
Native dependencies lookup fail for apis that do not follow naming convention


belemaire:
In a recent release we relaxed naming convention of APIs to allow for naming APIs in any way. Unfortunately because of that, it makes it way harder to identify APIs modules in the node_modules during container generation. It actually fails to find the API if it's not following the naming convention (regression).

It prove not to be a wise choice to remove strict naming convention for APIs (and API implementation projects) naming. We should reinstate this strict naming convention.

Changes are required both for `create-api` and `create-api-impl` commands to enforce the naming convention upon creation of the API / API impl.

belemaire:
In the end, we'll keep non enforcement of naming convention. The native dependencies lookup algorithm will properly find apis that do not follow the naming convention.

PR #439, opened 2017-12-18
Refactor native dependencies lookup

belemaire:
- Relocate native dependencies lookup from `ern-util` to `ern-core`.
- Rename file from `findNativeDependencies` to `nativeDependenciesLookup` and break down in smaller functions.
- Update native dependencies lookup algorithm to better identify apis and native apis implementations by parsing the package.json.
- Instead of returning an array of all native dependencies, the function now returns an object that classifies native dependencies as follow :
  - `apis`
  Native dependencies that are APIs
  - `nativeApisImpl`
  Native dependencies that are native APIs implementations
  - `thirdPartyInManifest`
  Native dependency that is a third party react native plugin, and is declared in the Manifest
  - `thirdPartyNotInManifest`
  Native dependency that is a third party react native plugin, or just a package unrelated to react native but that contains native code, and is not declared in the Manifest
  - `all`
  All native dependencies (union of all previous categories)
- Update clients of former `findNativeDependencies` function to use the new way

Closes https://github.com/electrode-io/electrode-native/issues/430
Closes https://github.com/electrode-io/electrode-native/issues/428
Closes https://github.com/electrode-io/electrode-native/issues/426


ern-core/src/MiniApp.js
```diff
@@ -267,14 +271,14 @@ Are you sure this is a MiniApp ?`)
         await spin(`${versionLessDependency.toString()} is not declared in the manifest. Performing additional checks.`,
                     yarn.add(DependencyPath.fromString(dependency.toString())))
 
-        const nativeDependencies = findNativeDependencies(path.join(tmpPath, 'node_modules'))
-        if (nativeDependencies.length === 0) {
+        const nativeDependencies = await nativeDependenciesLookup.findNativeDependencies(path.join(tmpPath, 'node_modules'))
+        if (nativeDependencies.all.length === 0) {
```
krunalsshah: nativeDependencies && nativeDependencies.all.length

krunalsshah: or _isEmpty()

ern-core/src/MiniApp.js
```diff
@@ -267,14 +271,14 @@ Are you sure this is a MiniApp ?`)
...
           log.debug('Pure JS dependency')
           // This is a pure JS dependency. Not much to do here -yet-
           finalDependency = versionLessDependency
-        } else if (nativeDependencies.length >= 1) {
-          log.debug(`One or more native dependencies identified: ${JSON.stringify(nativeDependencies)}`)
-          for (let dep: Dependency of nativeDependencies) {
+        } else if (nativeDependencies.all.length >= 1) {
```
krunalsshah: if nativeDependencies is never going to be null. it is okay to leave it unguarded . Otherwise a check would guard it. same as above

ern-core/src/nativeDependenciesLookup.js
```diff
@@ -0,0 +1,77 @@
...
+import {
+  Dependency
+} from 'ern-util'
+import readDir from 'fs-readdir-recursive'
+import _ from 'lodash'
+import path from 'path'
+import manifest from './Manifest'
+import * as ModuleTypes from './ModuleTypes'
+
+export function findDirectoriesContainingNativeCode (rootDir: string) : Array<string> {
```
krunalsshah: I'll take this as an item to test this class .

ern-core/src/MiniApp.js
belemaire: yes it's never going to be null/undefined, empty array by default, but will use `_.isEmpty` instead, as it is more proper

ern-core/src/MiniApp.js
belemaire: yes it's never going to be null/undefined, empty array by default, but will use `_.isEmpty` instead, as it is more proper

belemaire: yes it's never going to be null/undefined

belemaire:
Review comments addressed, just used `_.isEmpty` where appropriate instead of using `length===0`

PR #448, opened 2017-12-20
handle incorrect copy from version value while adding native app

krunalsshah:
Execute 
ern cauldron add nativeapp walmart:android:17.24.0 -c 1000.1000.1
ern cauldron add nativeapp walmart:android:17.24.0 -c appname:android:12.2.2
`TypeError: Cannot read property 'nativeDeps' of undefined`

After Fix 
```
Error: Could not resolve native application version to copy Cauldron data from.
Examine current value : 1000.1000.1```


belemaire:
:+1: Thanks

PR #463, opened 2018-01-02
Remove ern-util

belemaire:
Move everything from `ern-util` to `ern-core`.

All tests green (unit & system).

Farewell `ern-util` :wave: :cry:


krunalsshah:
@belemaire thanks for this refactor. This was discussed so many times 👍

PR #465, opened 2018-01-02
add version number to the apis and implementations added to the getti…

deepueg:
…ng started app.


belemaire:
Bypassing tests (doc changes)

PR #468, opened 2018-01-03
Add more logs for platform config command.

deepueg:
Ref: https://github.com/Microsoft/appcenter-cli/issues/315

### DIFF
--- a/ern-local-cli/src/commands/platform/config.js
+++ b/ern-local-cli/src/commands/platform/config.js
@@ -28,6 +28,7 @@ exports.handler = function ({
         : value
 
       ernConfig.setValue(key, valueToset)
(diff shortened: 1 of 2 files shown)


ern-local-cli/src/commands/platform/config.js
```diff
@@ -28,6 +28,7 @@ exports.handler = function ({
         : value
 
       ernConfig.setValue(key, valueToset)
+      log.info(`Done! ${key}: ${ernConfig.getValue(key)}`)
```
krunalsshah: ${key} set to ${ernConfig.getValue(key) , that way we convey the message of the method purpose. minor suggestion.

deepueg: Done.

Issue #445, opened 2017-12-19
Figure out yarn.lock for git based MiniApps


belemaire:
Once we support specifying branches for MiniApps instead of npm published versions, there will probably be some kind of issue with the yarn.locks. Figure out the problem and do the necessary improvements to properly support git based (branch/tag/commit sha) MiniApps.

Issue #374, opened 2017-11-22
Allow non NPM published MiniApps to be released to Container/CodePush


belemaire:
As of now, only NPM published MiniApps versions can be added/updated in a target Container or released through a CodePush update.

This is needed because of the immutability of a published MiniApp version, which guarantees that the code and native dependencies included inside the published version will remain exactly the same (critical in order to know for sure that the native dependencies that are listed in a Container -stored in Cauldron- are matching the ones from the MiniApps). Without this immutability guarantee, for example if we were to store just a link to a MiniApp GitHub repo and branch instead of a published version, we would have no way of being sure that what is listed in the Cauldron Container data in term of native dependencies, reflect the current state of the GitHub branch (as branches are mutable and a developer could just add a new native dependency, commit to the branch and the Cauldron will have no idea about it).

However, the fact that we enforce only NPM published versions of MiniApps to be released (added/updated) to a Container or through a CodePush update, impose some constraints on MiniApp development teams, as they always need to publish a new version of their MiniApp in order to regenerate a Container. For some development workflows it might be unwanted.

In addition to supporting adding/updating NPM published versions of MiniApps, Cauldron should also allow for adding/updating MiniApps git repo / tags/commits.

PR #481, opened 2018-01-12
Add support for non registry MiniApp paths to Cauldron

belemaire:
This PR introduce support in Cauldron/Container Generator for non registry based MiniApp paths.
It makes it possible to add non NPM published MiniApps to a Cauldron and properly generate a container out of it. Due to [current yarn bug with the lock file in case of git dependency paths](https://github.com/yarnpkg/yarn/issues/4734), the lock will not be generated, nor used if one of the MiniApp path is git based.

This feature makes it more convenient for development and CI. For example, one can create a native app `MyAppDev:android:1.0.0` in Cauldron and add MiniApps as git paths to this specific native application name/version. All it takes then to regenerate a new Container including all the latest code changes from all MiniApps is just to run `ern cauldron regen-container -d MyAppDev:android:1.0.0` at any time. No need to publish MiniApps to NPM, nor update the Cauldron in any way, to propagate any code changes to a new Container.

closes https://github.com/electrode-io/electrode-native/issues/374
closes https://github.com/electrode-io/electrode-native/issues/445

### DIFF
--- a/docs/GLOSSARY.md
+++ b/docs/GLOSSARY.md
@@ -11,6 +11,21 @@ A partial native application descriptor is a string with format `[nativeAppName]
 - `nativeAppName` : alphanumeric native application name , cannot contain character ':'
 
 ## Electrode Native module name
+
 The Electrode Native module name applies to modules created with Electrode Native cli.
 - Electrode Native recommends module name must only contain upper and/or lower case letters. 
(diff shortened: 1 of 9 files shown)


deepueg:
That's a much-awaited upgrade many devs have been asking for. Thanks, @belemaire.

PR #486, opened 2018-01-15
Container generators system tests

belemaire:
Add two new system tests that are ensuring that the content of generated iOS and Android containers matches the ones of pre-defined fixtures.

The fixtures are two generated containers (iOS and Android) that are containing the movie list MiniApp, the movie details MiniApp and react-native-code-push (to reflect the containers that are generated during system tests).

The test check for exact file system structure and file content match for all files but a few.

For Android, we allow file content differences in `index.android.bundle` and `index.android.bundle.meta` as these files might legitimately differ between subsequent generations.

For iOS, we allow file content differences in `MiniApp.jsbundle` and `MiniApp.jsbundle.meta` as these files might legitimately differ between subsequent generations. We also allow file content difference in `project.pbxproj`. This file contains a lot of generated data that we should check, but unfortunately it also contains a lot of random ids that are creating noise for comparaison. Further improvements to system tests might be around partial comparison of `project.pbxproj`.


krunalsshah:
Will merge it.

PR #500, opened 2018-01-19
improvements ElectrodeContainer.java

deepueg:
1. Remove static reference to ReactInstanceManager since ReactInstanceManager stores an activity instance and keeping it static might cause memory leaks. 
2. Throw illegal state exception with a proper message if the container is accessed before it is initialized. 
3. Formatting
4. Proper annotations.
5. Proper logging. 

Tested and validated by generating a new container.

### DIFF
--- a/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
+++ b/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
@@ -17,11 +17,10 @@ import android.app.Activity;
 import android.app.Application;
 import android.content.Intent;
 import android.os.Build;
-import android.os.Bundle;
 import android.provider.Settings;
 import android.support.annotation.NonNull;
+import android.support.annotation.Nullable;
 import android.util.Log;
-import android.widget.Toast;
 
 import java.lang.reflect.Method;
 import java.lang.reflect.InvocationTargetException;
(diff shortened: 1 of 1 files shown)


ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
```diff
@@ -245,12 +265,19 @@ private static void notifyReactNativeReadyListeners() {
...
     public static void resetReactNativeReadyListeners() {
         reactNativeReadyListeners.clear();
     }
 
     public interface ReactNativeReadyListener {
-            void onReactNativeReady();
+        void onReactNativeReady();
+    }
+
+    private static void throwNotInitializedStateException() {
```
belemaire: Could you rename this method to  something`throwIfNotInitialized()` ?
The current name of the method is misleading. The name of the method states that calling it will throw a `NotInitializedStateException`.

ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
```diff
@@ -100,88 +97,102 @@ private ElectrodeReactContainer(Application application,
...
+            for (ReactPackage instance : reactPackages) {
                 try {
-                  Method onReactNativeInitialized =
+                    Method onReactNativeInitialized =
                     instance.getClass().getMethod("onReactNativeInitialized");
-                  onReactNativeInitialized.invoke(instance);
+                    onReactNativeInitialized.invoke(instance);
+                }
+                catch (NoSuchMethodException e) {
+                    Log.e(TAG, "Container Initialization failed: " + e.getMessage());
```
belemaire: We don't really want to log an error for `NoSuchMethodException` as it is going to happen for most of the packages (`react-native-electrode-bridge` is the only native module Package that expose this method. All other native modules (code-push for ex) don't have this method, and therefore calling `getMethod` will throw).

Keeping the logging for `IllegalAccessException` and `InvocationTargetException` is OK.

deepueg: Good catch!. Will remove.

ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
deepueg: sure, will update.

deepueg:
@belemaire Updated. Please review.

PR #502, opened 2018-01-19
do not store the ReactInstanceManager inside delegate, instead query …

deepueg:
…the container. This makes sure that the delegate instance generate before a container is initialized will not result into NPE.


ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactActivityDelegate.java
```diff
@@ -65,25 +60,26 @@ public void invokeDefaultOnBackPressed() {
     };
 
     public ElectrodeReactActivityDelegate() {
-        mReactInstanceManager = ElectrodeReactContainer.getReactInstanceManager();
+
```
belemaire: Can we get rid of this constructor as it is now useless (better than keep it empty for no reason) ?

deepueg:
@belemaire Done. Also added another commit to support multiple reactViews inside an activity.

PR #505, opened 2018-01-22
add missing hasIntance() method to container.

deepueg:
This commit was missed in the previous PR.

### DIFF
--- a/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
+++ b/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
@@ -205,8 +205,6 @@ public class ElectrodeReactContainer {
         return sInstance;
     }
 
-
-    
     @SuppressWarnings("WeakerAccess")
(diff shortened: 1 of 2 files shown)


ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
```diff
@@ -221,6 +219,10 @@ public static boolean isReactNativeReady() {
         return sIsReactNativeReady;
     }
 
+    public static boolean hasInstance() {
```
krunalsshah: We already have a getInstance() method. When we `initialize` the container we create a new instance if not already created , else return the existing instance. (getInstance() != null)  would be sufficient?  We have a `getReactInstanceManager()` method as well (getReactInstanceManager() != null) may be sufficient in that case. It's not clear if `hasInstance()` method is conveying instance of `ReactInstanceManager` or `ElectrodeReactContainer` , except by looking at the implementation.Lastly it is possible that `sInstance` may be null & may cause NPE on a object.

deepueg:
@krunalsshah Thanks.! fixed.

Issue #509, opened 2018-01-23
create-api ignores --schemaPath parameter


abdyer:
When running `ern create-api` with the `--schemaPath` parameter (relative path, absolute path, even with the schema file in the same directory) on ERN platform version 0.11.2, the resulting API project uses the default starter schema in each case. To workaround this, we copied our intended schema into the API project and ran `ern regen-api`.

Is there a non-obvious syntax for specifying the schema path for `ern create-api`?

krunalsshah:
Thanks for reporting the issue. We will take a look at this.

PR #512, opened 2018-01-23
 create-api uses --schemaPath to accept initial config

krunalsshah:
Fixes https://github.com/electrode-io/electrode-native/issues/509

@belemaire Planning to cleanup `shouldGenerateBlankApi` property as it is never used. We do not expose this option any way. Thoughts?

### DIFF
--- a/ern-api-gen/src/generateProject.js
+++ b/ern-api-gen/src/generateProject.js
@@ -1,4 +1,5 @@
 import path from 'path'
+import fs from 'fs'
 import {
   fileUtils,
   ModuleTypes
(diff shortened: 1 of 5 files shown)


ern-api-gen/src/generateProject.js
```diff
@@ -92,14 +93,20 @@ export function generatePackageJson ({
   }, null, 2)
 }
 
-export function generateInitialSchema ({
+export async function generateInitialSchema ({
```
belemaire: Why making it `async` ? It does not await on anything nor return a Promise.

ern-api-gen/src/generateProject.js
```diff
@@ -92,14 +93,20 @@ export function generatePackageJson ({
...
   namespace,
-  shouldGenerateBlankApi
+  shouldGenerateBlankApi,
```
belemaire: I thought you were getting rid of this ?

krunalsshah: @belemaire yes there is another PR i am testing and updating the test

ern-api-gen/src/generateProject.js
krunalsshah: @belemaire fileUtils.readFile is a async function

belemaire: My bad ;)

Issue #396, opened 2017-11-28
Cauldron improvements


belemaire:
- [x] Simple linking of Cauldron schema with Electrode Native version https://github.com/electrode-io/electrode-native/issues/392
Done in :

- [x] Keep min Electrode Native version required in cauldron-api https://github.com/electrode-io/electrode-native/issues/393
Done in :

- [x] Add `cauldron upgrade` command https://github.com/electrode-io/electrode-native/issues/394
Done in :

- [x] Improve Cauldron schema https://github.com/electrode-io/electrode-native/issues/395
Done in :

- [x] Distinguish between MiniApp / JS API impl https://github.com/electrode-io/electrode-native/issues/306
Done in :

Issue #395, opened 2017-11-28
Improve Cauldron schema 


belemaire:
Refer to https://github.com/electrode-io/electrode-native/issues/306 for suggested changes (not including `jsApiImpl` section which should stay in https://github.com/electrode-io/electrode-native/issues/306).

Issue #394, opened 2017-11-28
Add cauldron upgrade command


belemaire:
Depends on https://github.com/electrode-io/electrode-native/issues/392

Add `cauldron upgrade` command that will check the stored `minErnVersion` in currently activated Cauldron and will properly upgrade the Cauldron schema if needed.

- If `minErnVersion` stored in Cauldron is the same as the `minErnVersion` stored in current Electrode Native version, then this command should not perform anything and just log something like :

```
Nothing to upgrade !
[CAULDRON_ALIAS] Cauldron is already using the proper Cauldron schema.
```

- If `minErnVersion` stored in Cauldron is < `minErnVersion` stored in current Electrode Native version, then this command should properly upgrade the Cauldron with the new schema :
  - Run all upgrade scripts (including the one of each intermediate versions if needed). For example if Cauldron is currently indicating minErnVersion of 0.13.0 and user is installing 0.20.0 and there was two intermediate Cauldron schema versions updates, let's say 0.15.0 and 0.18.0, the upgrade command should run the 0.15.0 upgrade script, followed by the 0.18.0 upgrade script (all previous upgrade scripts should be kept in any Electrode Native version).
  - Update the cauldron new `minErnVersion` to the proper value.
  - Push all the changes to the Cauldron. 

- If `minErnVersion` stored in Cauldron is > `minErnVersion` stored in current Electrode Native version, then prompt the user that he is using a version of Electrode Native that is not compatible with the current Cauldron and should update his platform version, something similar to the error message in Depends on https://github.com/electrode-io/electrode-native/issues/392

```
Nothing to upgrade !
[CAULDRON_ALIAS] Cauldron only supports Electrode Native v0.11.0 and above.
You are currently using Electrode Native v0.10.2.
If you want to use this Cauldron you need to upgrade your Electrode Native version.
```

Issue #393, opened 2017-11-28
Keep min Electrode Native version required in cauldron-api


belemaire:
Depends on https://github.com/electrode-io/electrode-native/issues/392

Once minimum Electrode Native version information is stored in the Cauldron, we also need a way to know that a new Electrode Native version contains breaking changes to the Cauldron structure, so that we can update this `minErnVersion` in the Cauldron.

For example, given that `minErnVersion` stored in the Cauldron is `0.11.0` and I have just installed `0.14.0` of platform, which contains breaking changes to Cauldron schema, any command that is accessing the Cauldron should fail and indicate to the user that Cauldron schema needs to be upgraded (using the new `ern cauldron upgrade` command). Running this command will upgrade the Cauldron with the new schema and properly update the `minErnVersion` to be `0.14.0`.
If user skips `0.14.0` and goes directly to `0.15.0` (which let's say does not contain Cauldron schema breaking changes) it should still fail and ask user to run same command, which will then properly set `minErnVersion` to `0.14.0` in Cauldron. Therefore, the `minErnVersion` should be stored somewhere in the platform code (using the previous example, `0.14.0`, `0.14.1`, 0.15.0` would store the `minErnVersion` as `0.14.0`).

PR #513, opened 2018-01-23
Cauldron improvements

belemaire:
This PR contains the following :

### Add support for Cauldron JSON document structural upgrades

Cauldron schemas are now versioned. The first schema version as of the changes introduced in this PR is `1.0.0`, and will be released with `0.12.0` of the Electrode Native.
The schema version used by a Cauldron is now stored in a new field `schemaVersion` in the top level Cauldron JSON document. If the `schemaVersion` is missing (as of ElectrodeNative < `0.12.0`), the schema version is assumed to be `0.0.0`.
The version of the Cauldron schema used by an Electrode Native version, is stored in the `ern-cauldron-api` module.
Every time a Cauldron is accessed for an operation, Electrode Native now checks if the version of the Cauldron schema match the one of the current `ern-cauldron-api` module. Three outcomes can then occur :
  1. **The schema version of the Cauldron matches the one of current version of `ern-cauldron-api`**
  In that case, then no problem, the operation will proceed
  2. **The schema version of the Cauldron is newer than the one of current version of `ern-cauldron-api`**
  In that case the operation will fail and an error message will be logged to let the user know that he/she should upgrade to a more recent version of Electrode Native that support this Cauldron document version.
  3. **The schema version of the Cauldron is older than the one of current version of `ern-cauldron-api`**
  In that case the operation will fail and an error message will be logged to let the user know that he/she should run the `ern cauldron upgrade` command to automatically upgrade the cauldron structure to the one matching the current schema.

The `ern cauldron upgrade` command is based on a script that will properly update the Cauldron document structure. If multiple schema versions have been missed, the command will make sure that all scripts are run one after the other (for example if current Cauldron schema version is `1.0.0` and target one is `3.0.0` the upgrade will run the script to upgrade from `1.0.0` to `2.0.0` followed by the script to upgrade from `2.0.0` to `3.0.0`.

This feature allows for breaking structural Cauldron upgrades to be done, without having to tell users to manually upgrade their Cauldron document by hand (would be very painful), while discarding backward compatibility to ease platform maintenance. These structural changes won't hopefully happen very frequently, but it is a good thing to have a mechanism to allow them with as minimal friction as possible from the user and Electrode Native developer perspective.

### Improve Cauldron structure

This PR lightly improve the Cauldron structure, which is now considered as Cauldron schema version `1.0.0`.

- Move `miniApps` and `nativeDeps` inside `container` object 

**BEFORE**

```json
"nativeDeps": [
  "react-native@0.42.0",
  "react-native-code-push@1.17.1-beta",
  "react-native-stack-tracer@0.1.1"
],
"miniApps": {
  "container": [
    "@walmart/react-native-cart@4.0.0",
    "@walmart/react-native-thankyou@2.0.0"
  ]
},
"codePush": {}
```

**AFTER**

```json
"container": {
  "nativeDeps": [
    "react-native@0.42.0",
    "react-native-code-push@1.17.1-beta",
    "react-native-stack-tracer@0.1.1"
  ],
  "miniApps": [
    "@walmart/react-native-cart@4.0.0",
    "@walmart/react-native-thankyou@2.0.0"
  ]
},
"codePush": {}
```

This changes make the Cauldron structure more proper. We can see what is exactly inside the `container` (native dependencies and MiniApps). Before this change, the fact that container was inside miniapps while native dependencies where completely outside container, was not making any sense. Now things are clear. One can see what's inside the Container of a native application version, as well as what is inside codepush deployments (the two ways to deploy : container/codepush).

### Add `jsApiImpls` entry in Cauldron

Before this PR, JS API implementation were treated as if they were native dependencies. This was not a proper way and was leading to more complicated handling in platform code, and was not supported by code push at all.

JS API implementations are pure JS modules, so they don't contain any native code. They are not native dependencies. They are closer to being MiniApps (in the sense that they can be CodePushed), but shouldn't be fully considered as MiniApps.

To make things more coherent, this PR add a `jsApiImpls` entry to the Cauldron (in the `container` object, in addition to `nativeDeps` and `miniApps` arrays, `jsApiImpls` is just another array containing all JS API implementations part of the container.

The same applies to `codePush` entries in the Cauldron, in addition to containing an array of `miniapps` the `codePush` entries now also contain a `jsApiImpls` array. 

JS API implementations, as of this PR are also properly supported through `code-push release` command (a new option `--jsApiImpls` has been added to the command to complement `--miniapps`). It is now possible to CodePush MiniApps and JS API implementations updates in a single CodePush release, or just single MiniApp(s) or JS API implementation(s).

### Commands for add/update/del JS API Implementation in Cauldron

Introduce the following commands :

- `ern cauldron add jsapiimpls`
- `ern cauldron del jsapiimpls`
- `ern cauldron update jsapiimpls`

That works the same way as their `miniapps` counter parts, to add/del/update JS API Implementations to/from a Container stored in Cauldron.

We might consider soon a big refactor of commands as we start having too many for adding/del/updating things in Cauldron. We might not need this level of granularity as platform can know if its dealing with a JS API implementation, a native API implementation, a MiniApp or a native dependency. Maybe `ern cauldron add`, `ern cauldron del` and `ern cauldron update` can be enough and provided anything. The need to specifiy `jsapiimpls`, `miniapps` or `dependencies` might be too much and does not really add any benefits from the user point of view.

Given that these commands add/del/update things from the Container, it might be even more proper to rename `cauldron` to `container` in that case. `ern container add` ...

So yes, should be discussed, I think commands could be improved in their naming to make more sense as part as a command refactoring sprint.

This PR also contain tests (unit/system) as well as documentation updates.

Close https://github.com/electrode-io/electrode-native/issues/306
Close https://github.com/electrode-io/electrode-native/issues/392
Close https://github.com/electrode-io/electrode-native/issues/393
Close https://github.com/electrode-io/electrode-native/issues/394
Close https://github.com/electrode-io/electrode-native/issues/395
Close https://github.com/electrode-io/electrode-native/issues/396


ern-cauldron-api/src/upgrade-scripts/0.0.0-1.0.0.js
```diff
@@ -0,0 +1,38 @@
+import CauldronApi from '../CauldronApi'
+
+export default async function upgrade (cauldronApi: CauldronApi) {
+  try {
+    log.info('Upgrading Cauldron schema from v0.0.0 to v1.0.0')
```
krunalsshah: these versions will change in future, better to parameterize them

ern-cauldron-api/test/CauldronApi-test.js
```diff
@@ -98,6 +101,23 @@ describe('CauldronApi.js', () => {
     })
   })
 
+  // ========================================================== 
+  // getCauldronSchemaVersion
+  // ========================================================== 
+  describe('getCauldronSchemaVersion', () => {
+    it('should return the Cauldron schema version', async () => {
+      const schemaVersion = await cauldronApi().getCauldronSchemaVersion()
+      expect(schemaVersion).eql('1.0.0')
```
krunalsshah: anyway we can have a constant that states schemaVersionCurrent & schemaVersionNext ? This can be used at other parts of the app?

ern-local-cli/src/commands/cauldron/del/dependencies.js
```diff
@@ -80,27 +80,20 @@ exports.handler = async function ({
       }
     })
 
-    console.log(`heheheehehrrre`)
```
krunalsshah: eeks

ern-local-cli/src/commands/cauldron/upgrade.js
```diff
@@ -0,0 +1,24 @@
...
+exports.handler = async function () {
+  await utils.logErrorAndExitIfNotSatisfied({
+    cauldronIsActive: {
+      extraErrorMessage: 'A Cauldron must be active in order to use this command'
+    }
+  })
+  try {
+    const cauldron = await coreUtils.getCauldronInstance({ignoreSchemaVersionMismatch: true})
+    await cauldron.upgradeCauldronSchema()
+    log.debug(`Cauldron was succesfully upgraded !`)
```
krunalsshah: minor replace back ticks with '

ern-local-cli/src/lib/publication.js
```diff
@@ -392,7 +411,8 @@ miniApps: Array<PackagePath>, {
         releasedBy: codePushResponse.releasedBy,
         rollout: codePushResponse.rollout
       },
-      miniAppsToBeCodePushed)
+      miniAppsToBeCodePushed,
+      jsApiImplsToBeCodePushed)
 
     const pathToNewYarnLock = path.join(tmpWorkingDir, 'yarn.lock')
     await spin(`Adding yarn.lock to Cauldron`, cauldron.addOrUpdateYarnLock(napDescriptor, deploymentName, pathToNewYarnLock))
```
krunalsshah: minor remove back tick

ern-local-cli/src/commands/cauldron/del/dependencies.js
belemaire: yes seems like these logs slipped through a review ;P

ern-cauldron-api/src/upgrade-scripts/0.0.0-1.0.0.js
belemaire: There is one upgrade script per new schema version, as the operations to upgrade to a new schema version will be different for each version. So this is the script to upgrade from schema version `0.0.0` to `1.0.0` of the schema `thus the script is named 0.0.0-1.0.0.js`. If we come up with a new schema `2.0.0` we will create a script `1.0.0-2.0.0.js`. Some of the things might be common for all the scripts, but we will refactor if needed, once we have more than one script ;)

PR #514, opened 2018-01-24
generate toString() inside Java models.

deepueg:


krunalsshah:
@deepueg how about add equals to?

deepueg:
has to be in a separate PR since that needs more detailed implementation.

belemaire:
@deepueg : Approving it and will merge, but I have no real idea of what the generated output looks like 😕 Ideally it should come with at least one test fixture were we could see how the rendered method looks like.
But I understand this was not really done yet even for other methods. @gmbharath12 is working on system tests that will compare the whole content of generated APIs which is a good global safety net but not really granular.
I think we should consider breaking down the file into separate parts (methods being one of the part) and instead of inlining all templates in the model class, break it down into smaller templates (For example here for the toString method, in the class template we would insert `{{toStringMethod}}` and the template for `toStringMethod` would be kept separate, this way unit tests can generate fixtures just for the toString() method and granularly test it. Making it easier to tests and also cleaner when looking at the class model template, seeing all the sections.

PR #528, opened 2018-01-30
Update system tests for new platform version

belemaire:


belemaire:
Do not merge yet. Need to update fixtures.

Issue #536, opened 2018-02-05
ignoreRnpmAssets should be expose to create-container


belemaire:
The `ignoreRnpmAssets` flag is used by Container generator. If set, any RNPM asset of MiniApps won't be packaged inside the Container.

This flag can only be set inside the Container generator configuration stored in a Cauldron.

This flag should also be exposed to the `create-container` command so that it is possible to specify it when generating a Container locally without relying on a Cauldron.

PR #541, opened 2018-02-05
Ignore rnpm assets

belemaire:
Closes https://github.com/electrode-io/electrode-native/issues/536

Also removes `version` and `containerName` options from the `create-container` command as they are not used any-longer for creating a container (publication specific options. left-overs from past refactoring).


krunalsshah:
@belemaire I saw that `ignoreRnpmAssets` is propagated to `publication.js`, is `ignoreRnpmAssets` flag propagated to platform specific `ContainerGenerator` ?

belemaire:
@krunalsshah 
The flag is part of the Container generator config (actually the only argument to the Container generator `generate` method). It is not platform specific, part of the platform agnostic generator config provide to both iOS and Android generators generate method.

PR #551, opened 2018-02-07
Update ElectrodeContainer hull to swift 4.0

clairewli:
Migrate ElectrodeContainer hull to swift 4.0


belemaire:
Does this change needs to be synchronized with the bridge and apis migration to Swift 4 or is it completely uncoupled ? Or said differently, if we merge this today and release it in `0.13.0` of Electrode Native, will it break Container build if current bridge version / current apis versions are being injected in the container ?

belemaire:
Closing for now as we need to figure out a proper approach not to impose such a big breaking change on users of Electrode Native that are not yet using Swift 4.

PR #565, opened 2018-02-13
create-api command passes wrong property name downstream for --schema

### LINKED ISSUE #509: create-api ignores --schemaPath parameter

krunalsshah:
Fixes #509 

map the property name to the one used to normalize config.


gmbharath12:
Thanks @krunalsshah 👍

PR #584, opened 2018-02-16
print line by line difference if file comparison fails in system tests

krunalsshah:
print line by line difference if file comparison fails in system tests


belemaire:
Nice ! Good initiative. Thanks @krunalsshah

PR #587, opened 2018-02-16
Update GLOSSARY.md

jomnius:
I was trying to figure out correct package path format to use, when adding miniapp into cauldron. I think I got it now and would like to clarify this for next time by explaining which paths are related to which storage methods.


krunalsshah:
Appreciate you contributing :)

PR #598, opened 2018-02-19
Update requirements.md

jomnius:
FIX: I cannot find any references to ANDROID_SDK env variable in Android docs, only ANDROID_HOME which is expired. Therefore I propose using ANDROID_SDK_ROOT (works for me) based on Android Studio docs https://developer.android.com/studio/command-line/variables.html

FIX: link to NPM website, since all other tools have a similar link

FIX: Xcode is written Xcode (only one capital letter).


jomnius:
I pull this back, since my system won't work with ANDROID_SDK_ROOT, but wants ANDROID_HOME. Looks like it's in gradle settings, to be checked some other time. PR closed.

PR #607, opened 2018-02-19
Update manifest.md

jomnius:
I would like to make these things more clear:

- Cauldron should be handled via command line tool EXCEPT in one special case: when adding config option to define which manifest to use and how. I searched for a long time to figure out how to do it from command line. This webpage is so far the only place, which actually says you're allowed to modify cauldron manually. Therefore using bold format to highlight it.

- You can override manifest partially or fully. While main use case is partial override, the text seemed to talk about full override. That was confusing.

- How to create plugin configuration chapter was after a) "Why" chapter and b) "Note" chapter, therefore it was not obvious where the "How" chapter started. It looked like it was addition to "note" and thus I miss it each time I get on this page. Still it's crucial in order to get custom modules integrated.

### DIFF
--- a/docs/platform-parts/manifest.md
+++ b/docs/platform-parts/manifest.md
@@ -45,11 +45,11 @@ For each new Electrode Native version, the master manifest will contain updated 
 
 ### Overriding the master Manifest
 
-You can override the master manifest with your own manifest file:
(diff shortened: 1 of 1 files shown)


belemaire:
@jomnius Thanks a lot for all of your PRs to improve the documentation

PR #609, opened 2018-02-20
Fix yarn.lock

belemaire:
Top level platform `yarn.lock` was containing links to packages from internal Walmart registry.
This is because these packages were added while logged in to the internal Walmart registry, not the public one.
system tests were failing due to this, as they are run from Travis, which has no access to the private Walmart registry.


belemaire:
Created issue https://github.com/electrode-io/electrode-native/issues/610 to make sure this does not happen again in the future.

PR #614, opened 2018-02-20
Add a way to bypass copy from option of cauldron add nativeapp

belemaire:
Add new value `none` for `copyFromVersion` of `cauldron add nativeapp`

This can effectively be used to disable interactive mode triggered by the `cauldron add nativeapp` command in case no previous native app version copy is desired (otherwise, the only way to discard copy from was through the interactive prompt, which makes things much more difficult for a CI job)


gmbharath12:
👍

Issue #610, opened 2018-02-20
Add pre-commit hook to ensure yarn.lock validity


belemaire:
Reference : https://github.com/electrode-io/electrode-native/pull/609

To avoid this, we should code a script, run as part of the pre-commit checks, to ensure that no `npme.walmart.com` url can be found in the `yarn.lock` of the top level platform as well as the `yarn.lock` of each of the Electrode Native modules.

PR #620, opened 2018-02-21
Add pre-commit hook to ensure yarn.lock validity

krunalsshah:
Fixes https://github.com/electrode-io/electrode-native/issues/610

### DIFF
--- a/yarn-lock.js
+++ b/yarn-lock.js
@@ -1,6 +1,7 @@
 const fs = require('fs')
 const lockfile = require('@yarnpkg/lockfile')
 const chalk = require('chalk')
+const path = require('path')
 
 const yarnLock = 'yarn.lock'
 const yarnRegistryRe = /https:\/\/registry.yarnpkg.com/
@@ -8,12 +9,12 @@ const yarnRegistryRe = /https:\/\/registry.yarnpkg.com/
 let ernDirs = fs.readdirSync(process.cwd()).filter(x => x.startsWith('ern'))
 ernDirs.push(yarnLock) // Root yarn.lock
 
-ernDirs.forEach((current) => {
-  try {
-    let yarnLockPath = current === yarnLock
-      ? yarnLock
-      : `${current}/${yarnLock}`
+ernDirs.forEach(current => {
+  let yarnLockPath = current === yarnLock
+    ? yarnLock
+    : path.join(current, yarnLock)
 
+  try {
     let yarnLockContent = fs.readFileSync(yarnLockPath, 'utf8')
     let yarnLockParsed = lockfile.parse(yarnLockContent)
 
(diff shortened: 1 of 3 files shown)


belemaire:
Nice :)

Could you extend it to match the requirements of #610 (`as well as the yarn.lock of each of the Electrode Native modules`).

The code only checks for wrong registry urls in the top level `yarn.lock` file but each module also have its own `yarn.lock` (for ex when we add/update/remove a dependency in `ern-cauldron-api` it's the `yarn.lock` of `ern-cauldron-api` that is modified (`ern-cauldron-api/yarn.lock`). 

One way is to just to look to the `yarn.lock` stored at the root of directories matching `ern-*` naming pattern.

You can use `readdirSync` and apply filtering on the result to get the array of directories easily, and just iterate through them.

```
> fs.readdirSync('.').filter(x => x.startsWith('ern'))
[ 'ern-api-gen',
  'ern-api-impl-gen',
  'ern-cauldron-api',
  'ern-container-gen',
  'ern-core',
  'ern-local-cli',
  'ern-runner-gen',
  'ern-util-dev' ]
```

yarn-lock.js
```diff
@@ -0,0 +1,62 @@
...
+const lockfile = require('@yarnpkg/lockfile')
+const chalk = require('chalk')
+
+const yarnLock = 'yarn.lock'
+const yarnRegistryRe = /https:\/\/registry.yarnpkg.com/
+
+let ernDirs = fs.readdirSync(process.cwd()).filter(x => x.startsWith('ern'))
+ernDirs.push(yarnLock) // Root yarn.lock
+
+ernDirs.forEach((current) => {
```
belemaire: minor: parenthesis are not needed for an arrow function with one param (`(current)` can just be `current`)

yarn-lock.js
```diff
@@ -0,0 +1,62 @@
...
+  try {
+    let yarnLockPath = current === yarnLock
+      ? yarnLock
+      : `${current}/${yarnLock}`
```
belemaire: `path.join(current, yarnLock)`

yarn-lock.js
```diff
@@ -0,0 +1,62 @@
...
+      }
+
+      /*
+       yarn.lock contains registry other than https://registry.yarnpkg.com
+       http://npme.walmart.com/@yarnpkg/lockfile/-/lockfile-1.0.0.tgz#33d1dbb659a23b81f87f048762b35a446172add3
+       http://npme.walmart.com/colors/-/colors-1.1.2.tgz#168a4701756b6a7f51a12ce0c97bfa28c084ed63
+       http://npme.walmart.com/diff/-/diff-3.4.0.tgz#b1d85507daf3964828de54b37d0d73ba67dda56c
+       */
+      if (incorrectRegistries.length > 0) {
+        console.log(chalk.bold.red(`${yarnLock} contains registry other than https://registry.yarnpkg.com`))
```
belemaire: could you use `path.join(current, yarnLock)` here in the log so that we know which yarn.lock file the log is about. maybe keep `path.join(current, yarnLock)` in a variable `currentYarnLockPath` as you have to use the path earlier in the code as well (avoid duplication)

PR #623, opened 2018-02-22
Docs: platform api doc to link to commands

jomnius:
SUGGESTION:

I like links in documentation, but e.g. this doc is missing them. Maybe you left them out on purpose, because you believe it's better to guide users through "learn the platform basics first" pipeline, before they might get lost into command details and be even more confused.

Anyway, I like links. No problem, if you feel this PR does not improve anything and close it.


krunalsshah:
Links are more relevant for this page. 👍 may be we missed it.

PR #629, opened 2018-02-22
fix the system tests relating to container generation

krunalsshah:
fix the system tests relating to container generation


belemaire:
:+1: ty

PR #634, opened 2018-02-22
Fix cauldron add publisher command

belemaire:
- Fix the command in case no descriptor is explicitly provided as a command option. In that case, the command was failing after selecting the platform with error `An error occurred: Looks like there are no application defined in cauldron for ios`. This was due to naming the argument of `getNativeAppsForPlatform` function as `platform` which was also the name of a local variable used in the function, leading to the issue.

- Remove the break statement so that it does not add the publisher to the first native app returned but in case of multiple native apps present in Cauldron, it will properly ask the user to choose to which native app the publisher should be added to.


gmbharath12:
👍

PR #635, opened 2018-02-23
Fix cauldron add nativeapp

belemaire:
`cauldron add nativeapp` command was failing with error `An error occurred: Cannot read property 'name' of undefined` when the application name/platform was present in the cauldron but no versions were present. 

Could be reproduced through :

- `ern cauldron add nativeapp newapp:android:1.0.0`
- `ern cauldron del nativeapp newapp:android:1.0.0`
- `ern cauldron add nativeapp newapp:android:1.0.0`

Last command would fail with this error.

This PR fixes this fringe scenario.

Thanks @gmbharath12 for reporting this one.


gmbharath12:
Thanks for the fix ! Much needed.

PR #651, opened 2018-03-01
Add a list of all NS*UsageDescription to ErnRunner

Igor1201:
So it won't crash when a plugin tries to use some native feature. See [CocoaKeys](https://developer.apple.com/library/content/documentation/General/Reference/InfoPlistKeyReference/Articles/CocoaKeys.html) for more info.

As iOS doesn't have a manifest merger feature like Android (Info.plist modifications on framework won't reflect on app), we need to add `NS*UsageDescription` keys to ErnRunner Info.plist (and later, manually on the main app).


belemaire:
Thanks :+1:
@clairewli Could you review this PR and merge if no issue. Thanks !

PR #655, opened 2018-03-05
Improve system tests

belemaire:
System tests are taking a long time to run and with the recent introduction of fixtures (container/apis), it makes it more difficult to work with and ensure that one of the test is still working after a change to Container or APIs for example. Current approach of running all system tests on a developer workstation is way too heavy. With the addition of more system tests, it was due time for a proper refactoring.

This PR is moving the system tests into a dedicated directory `system-tests`, instead of keeping them at the root of the workspace.

All utilities functions used by `system-tests.js` have been moved to dedicated source files (one per function), now located in `system-tests/utils`. This way they can be reused across independent test suites.

This PR breaks down the system tests into smaller units (one can think of them as test suites).
These individual units can contain one test or multiple ones. 
They should be independent (i.e execution of one test suite shouldn't depend on another one or have side effects impacting other test suites).

All the test suites are to be stored in the `system-tests/test` directory. As of this PR, instead of a big system test suite (former `system-test.js`), there is now 8 test suites. Most of them contain a single test testing against a specific fixture (`create-container-ios-fixture.js`, `create-api-impl-native-fixture.js` ...). The former big test suite (`system-test.js`) is now named (`misc.js`). Ideally tests within this suite should be exploded to smaller test suites, this is going to be ongoing work.

With that being done, the last commit of this PR makes it possible to run test suites granularly. If no options is passed to `npm run system-tests`, then the command will enter an interactive prompt asking the user to select one or more test suite(s) to run. The `--all` flag can be used to skip interactive prompt and run all test suites (this flag will be used for nightly CI job).


gmbharath12:
This is cool 👍

PR #660, opened 2018-03-07
Fix Android container config passing

swashcap:
* **Problem:** `ElectrodeReactContainer.Config` settings aren’t honored
* **Solution:** Ensure the `isReactNativeDeveloperSupport` property is set _before_ `ElectrodeReactNativeHost` instantiation reads the value.
* **Testing:** Need help!

### DIFF
--- a/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
+++ b/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
@@ -127,7 +127,6 @@ public class ElectrodeReactContainer {
             {{/apiImplementations}}
      ) {
         if (sElectrodeReactNativeHost == null) {
-            sElectrodeReactNativeHost = new ElectrodeReactNativeHost(application);
 
             // ReactNative general config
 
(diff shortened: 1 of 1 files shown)


swashcap:
Race condition where the `isReactNativeDeveloperSupport` value is read here:

https://github.com/electrode-io/electrode-native/blob/bd83545558f5faa6efdaddf1f2c7bd541cea11e7/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java#L267-L270

before it's set here:

https://github.com/electrode-io/electrode-native/blob/bd83545558f5faa6efdaddf1f2c7bd541cea11e7/ern-container-gen/src/generators/android/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java#L130-L135

deepueg:
Good catch.! thanks.!

swashcap:
Should have noted, @krunalsshah and I are testing this in the Android app. I'll report back if the fix works.

swashcap:
This PR _did not_ address the bug: `ElectrodeReactContainer.Config`’s `isReactNativeDeveloperSupport` property doesn’t appear to have any effect, and in some cases the developer menu isn’t accessible when `isReactNativeDeveloperSupport == true`.

abdyer:
@krunalsshah @swashcap I also had a problem with the developer menu not working and solved it by moving the `onKeyUp()` code to `dispatchKeyEvent()`:

```kotlin
    override fun dispatchKeyEvent(event: KeyEvent?): Boolean {
        val isMenuKey = event?.keyCode == KeyEvent.KEYCODE_MENU

        if (isMenuKey
            && ElectrodeReactContainer.isReactNativeDeveloperSupport()
            && reactActivityDelegate.canShowDeveloperMenu()) {
            reactActivityDelegate.showDeveloperMenu()
            return true
        }

        return super.dispatchKeyEvent(event)
    }
```

PR #672, opened 2018-03-09
Show error if ERN platform version not found

jomnius:
Mainly for your information, took a long time to debug. Figured it out only, when writing this bug report.

**PROBLEM**: I'm trying to uninstall and reinstall Electrode Native, both seem successful. However something is referring to old install. What else do I need to reset...

My build script:

```bash
echo "=== STEP 0"
whoami
pwd

# === Setup ERN

# MUST have log level  trace to fix this error:
# runCauldronContainerGen failed: TypeError: this.stream.clearLine is not a function
# "It is due to Ora (the JS library we use to display spinners during commands execution)
# that doesn't properly work in certain CI envs."
export  ERN_LOG_LEVEL=trace

echo "=== STEP 1"
npm uninstall -g electrode-native && ern

echo "=== STEP 1a"
npm install -g electrode-native && ern

echo "=== STEP 2"
# --- electrode-native/system-tests/system-tests.js
# Trace log level should be set to trace to ensure that `ora` 
# gets disabled as it can lead to issues on CI env
#ern platform config logLevel trace

echo "=== STEP 3"
```

My build error:

```bash
17:28:05 === STEP 0
17:28:05 jenkins
17:28:05 /Users/jenkins/jenkins/workspace/bacon-ci-android-build
17:28:05 === STEP 1
17:28:06 removed 93 packages in 0.796s
17:28:06 /var/folders/95/xlfqlj515b1b_py1hry21sr00000gn/T/jenkins8917781206425641990.sh: line 15: ern: command not found
17:28:06 === STEP 1a
17:28:10 /usr/local/bin/ern -> /usr/local/lib/node_modules/electrode-native/bin/ern.js
17:28:10 + electrode-native@1.0.13
17:28:10 added 93 packages from 32 contributors in 3.57s
17:28:10 module.js:545
17:28:10     throw err;
17:28:10     ^
17:28:10 
17:28:10 Error: Cannot find module '/Users/jenkins/.ern/versions/0.12.2/node_modules/ern-local-cli/src/index.prod.js'
17:28:10     at Function.Module._resolveFilename (module.js:543:15)
17:28:10     at Function.Module._load (module.js:470:25)
17:28:10     at Module.require (module.js:593:17)
17:28:10     at require (internal/module.js:11:18)
17:28:10     at runLocalCli (/usr/local/lib/node_modules/electrode-native/src/index.js:155:5)
17:28:10     at Object.<anonymous> (/usr/local/lib/node_modules/electrode-native/src/index.js:45:3)
17:28:10     at Module._compile (module.js:649:30)
17:28:10     at Object.Module._extensions..js (module.js:660:10)
17:28:10     at Module.load (module.js:561:32)
17:28:10     at tryModuleLoad (module.js:501:12)
17:28:10 === STEP 2
```

I also tried this, no change:

```bash
watchman watch-del-all
rm -rf node_modules
rm -rf $TMPDIR/react-*
rm -rf $TMPDIR/haste-map-react-native-packager-*
yarn --reset-cache
```

**QUICK FIX:**

I have local .ernrc file under version control, which defined used platform as 0.12.2 - which obviously did not exist, since by default ern installs only latest version (currently 0.14.2).

Remove this row from .ernrc file (as soon as there's someone else at the office to approve my PR)

```bash
  "platformVersion": "0.12.2",
```

**SOLUTION:**

Maybe code in file https://github.com/electrode-io/electrode-native/blob/master/global-cli/src/index.js `function runLocalCli` could be patched with something like this...

### DIFF
--- a/global-cli/src/index.js
+++ b/global-cli/src/index.js
@@ -149,10 +149,16 @@ function runLocalCli () {
 
   const ernRc = JSON.parse(fs.readFileSync(ernRcPath, 'utf-8'))
 
+  let ernPlatformPath
   if ((ernRc.platformVersion === '1000') || (ernRc.platformVersion === '1000.0.0')) {
-    require(`${ERN_VERSIONS_CACHE_PATH}/${ernRc.platformVersion}/ern-local-cli/src/index.dev.js`)
+    ernPlatformPath = `${ERN_VERSIONS_CACHE_PATH}/${ernRc.platformVersion}/ern-local-cli/src/index.dev.js`
   } else {
-    require(`${ERN_VERSIONS_CACHE_PATH}/${ernRc.platformVersion}/node_modules/ern-local-cli/src/index.prod.js`)
+    ernPlatformPath = `${ERN_VERSIONS_CACHE_PATH}/${ernRc.platformVersion}/node_modules/ern-local-cli/src/index.prod.js`
+  }
+  try {
+    require(ernPlatformPath)
+  } catch (e) {
+    console.error(`cannot find ${ernPlatformPath}`)
   }
 }
 


belemaire:
@jomnius 
Thanks for your PR.
A few things here :

1) In Step 1A 
`npm uninstall -g electrode-native && ern`
You don't need `&& ern` there. The `electrode-native` package IS `ern` binary. So once you uninstall it globally, the `ern` command is gone from the machine (that's why you see `line 15: ern: command not found` in the log). Maybe you already noticed it, but just wanted to let you know.

2) In Step 1B
`npm install -g electrode-native && ern`
This is actually 2 steps in theory (depends what you consider steps thoug) because of the `&&` which run one command, then the other ;) On our CI we tend to break it into `npm install -g electrode-native` followed by `ern --debug` (Adding the debug flag here helps in some case of installation failure as it logs a bit more).

Now. `ern` command, the first time it is run (if Electrode Native is not installed on -or has been removed from- the machine) will install the latest available version of the platform (`0.14.2` as of now). 
If you want to use a different version you can run (after the installation with `ern --debug` is complete), `ern platform use 0.12.2` for example, which will download and switch platform version. This is OK for most users as the first time they install Electrode Native, they probably would like to have the latest version of it. Unfortunately it does not play super well with CI because on CI we want consistency, and therefore not a good thing plafrom one build to another the version of Electrode Native installed can suddenly change because we released a new version.
That's why in our CI env, our build jobs are parameterized and take the version of Electrode Native to use as a parameter. Then the CI job does what you're doing `npm install -g electrode-native` followed by `ern`, but it also follows with `ern platform use $ENVERSION` where `$ENVERSION` is the version of Electrode Native we want to be on consistently for the build job. `platform use` command download the version, install it and switch to it. This way we can control exactly the version of EN to be used by the build job, and increment it when needed.
This is not perfect as it consume extra time to install the initial version (latest) as a first step, which should not be needed, but we have currently no way to specify an initial version to install, it always install the latest. I will create an issue to answer that, will be helpful on CI envs. Will take it an extra mile, to fit your use case, being that if an `.ernrc` file is present at the time we do the first install of ElectrodeNative, it should then use the version specified in `platformVersion` of this file. This way you can control the version of ElectrodeNative to initially install and to use, directly from this file, no need to have a a parameterized build job. 

Please also note that you might want to add something like `rm -rf ~/.ern` to your cleanup / uninstallation step of electrode-native, as it keeps some data there. this might not be required depending on your setup, but that's what is needed for real full cleanup of electrode-native in addition to npm uninstalling it.

With that being said, let me write a review for your PR.

belemaire:
@jomnius 
I have created an issue for global cli improvement to more appropriately answer your needs (and ours as well). https://github.com/electrode-io/electrode-native/pull/672
Feel free to comment on it if you want to correct something of have more suggestions related to this issue.

jomnius:
Going on vacation, back Thursday. If you wish, you can close and fix this in the right way. Or not to fix: no problem! Otherwise this will take some time :D

belemaire:
@jomnius 
Sure, thanks for letting me know, will do. I always prefer to let contributors have a chance to get their code merged rather tank saying 'your PR require a few changes, I'll close it and do it myself' ;)
But this is a valid reason no to. Enjoy your vacation !

belemaire:
@jomnius 
Just to let you know that I have made this change for better error logging in https://github.com/electrode-io/electrode-native/pull/674.
Also I have done what we discussed on another ticket, being to use the local `.ernrc` from where `ern` commands are issued to infer version to use for initial platform installation instead of installing the latest available by default https://github.com/electrode-io/electrode-native/pull/675.
Cheers.

Issue #673, opened 2018-03-09
Offer a way to initial install a specific Electrode Native version


belemaire:
As of now, our `global-cli` (`ern` binary) will always install the latest version of Electrode Native upon first time install of Electrode Native. This is the intended behavior for most users. However it does not play nicely with CI as we might want to have control upon the initial version of Electrode Native being installed for the job and not pick the latest available, as it leads to inconsistency (from one build job run to another, the version of Electrode Native installed will change in case a new version is released).

Workaround as of now, is to let the binary install the latest, and then issue a command to download and install (and switch to) the intended version. This is not ideal as it takes some extra time to install the initial (latest) version, which is not going to be used anyway by the job.

`ern` binary should therefore accept an extra option `--installVersion` for example which when used will install the specified version instead of the latest.

In addition to this option, another way for Electrode Native to install a different version than latest upon initial install, should be to read the `.ernrc` file if one is present and use the `platformVersion` specified in this file. This way one could just attach this file to the build job and control the version from there. The advantage of this approach is also that all things specified in the `.ernrc` will be reused (like the cauldron url, the log level preference etc ...), this way instead of running commands such as `ern cauldron add repo ...`, `ern platform config logLevel trace` ... which ultimately write config to this file, one could just control the config from this file attached to the build job, instead of running all the commands in every build job run (thus also improving performance as all these commands take time to run).

belemaire:
Reference issue : https://github.com/electrode-io/electrode-native/pull/672

PR #675, opened 2018-03-12
Update global cli to support first time install of specific version instead of latest

belemaire:
If a local `.ernrc` file is present, use the platform version from this configuration file for first time install of Electrode Native, instead of the latest available version of Electrode Native.

Close https://github.com/electrode-io/electrode-native/issues/673


gmbharath12:
👍 
@belemaire : Please merge after conflict is resolved.

Issue #662, opened 2018-03-07
Improve cauldron regen-container command


belemaire:
`cauldron regen-container` command should check if there was any native dependencies changes in the MiniApps and if that is the case, properly update the native dependencies in Cauldron.
This was not done because before supporting git repos in the Cauldron for MiniApps (rather than NPM published version), it was not necessary (as there can be no code changes in the same version of a published npm package).
However when using git repos for miniapps, every time the regen-container command is triggered, the latest code from the git repos of the miniapps is retrieved, which might contain new native dependencies or updated native dependencies version, in that case the cauldron native dependencies should be properly updated to reflect these changes.

Issue #498, opened 2018-01-18
Update dependency(ies) version(s) in Cauldron if needed 


belemaire:
When adding a new MiniApp to a Container in Cauldron, or updating an existing MiniApp version, Electrode Native should properly update any native dependencies versions if needed.

Here is the behavior it should adopt, based on different scenarios :

### Scenario 1

In Cauldron :
```
DepA@1.0.0
DepB@1.0.0
MiniAppA@1.0.0 (DepA@1.0.0, DepB@1.0.0)
```

We then want to update to MiniAppA@1.0.1 which includes DepB@2.0.0

The current behavior denies this because of the diff in DepB version. For CodePush it should still be the behavior, however for Container as only MiniAppA is using DepB there is no reason to deny the update.

=> In that scenario, Electrode Native should update MiniAppA@1.0.1 and also update DepB@2.0.0 in the Cauldron/Container.

### Scenario 2

In Cauldron :
```
DepA@1.0.0
DepB@1.0.0
MiniAppA@1.0.0 (DepA@1.0.0, DepB@1.0.0)
MiniAppB@1.0.0 (DepA@1.0.0, DepB@1.0.0)
```

As in previous scenario, we want to update to MiniAppA@1.0.1 which includes DepB@2.0.0

=> In that scenario, Electrode Native should deny update to MiniAppA@1.0.1 as it would imply updating to DepB@2.0.0. But MiniAppB is still using DepB@1.0.0, thus updating to DepB@2.0.0 would propably break MiniAppB.

### Scenario 3

In Cauldron :
```
DepA@1.0.0
DepB@1.0.0
MiniAppA@1.0.0 (DepA@1.0.0, DepB@1.0.0)
MiniAppB@1.0.0 (DepA@1.0.0, DepB@1.0.0)
```

As in previous scenario, we want to update to MiniAppA@1.0.1 which includes DepB@2.0.0, but we'll also update MiniAppB@1.0.1 which also includes DepB@2.0.0 (using batch command to update both MiniApp versions in a single pass).

=> In that scenario, Electrode Native should update MiniAppA@1.0.1 and MiniAppB@1.0.1 and also update DepB@2.0.0 in Cauldron/Container.

-----

To put it in a nusthell, we should succeed updating/adding MiniApps which are containing updated native dependencies,  and update the native dependencies versions accordingly only if all MiniApps using this native dependency are targeting the new updated version (exact version match as this is a native module).

## For API / API impl dependencies

[...]

PR #676, opened 2018-03-12
Improve native dependencies versions sync in Cauldron

belemaire:
This PR introduces new utility functions and alter the algorithms of the following commands :

```
- cauldron add miniapps
- cauldron update miniapps
- cauldron batch
- cauldron regen-container
```

Without digging into implementation details, the high level change here is regarding proper synchronization of native dependencies stored in the Cauldron for the Container impacted by these commands.

The new algorithm is doing the following :
- List the native dependencies of all the MiniApps that are getting added or updated to/in the Cauldron Container. Also list the native dependencies of all the MiniApps currently in the Container. In a nutshell we just list the native dependencies of all the MiniApps that are going to be listed in the Container once the operation complete.
- For all these native dependencies, we check the one that are conflicting following the rules of Electrode Native (for example a major version diff in an API/API impl or a patch/minor/major diff in a native module). If we find any conflict, we just abort the operation (unless force flag is used, where it will result in just a warning).
- Given that there is no conflicting version (or force flag was used) we just retain the latest version of each dependency from the list of dependencies (we also include all current native dependencies listed in the Container Cauldron).
- This list of native dependencies is now the current list of native dependencies for the Cauldron Container. We then just synchronize the Cauldron, to ensure that it match the list.

This PR close the two following issues :

Close https://github.com/electrode-io/electrode-native/issues/662
Close https://github.com/electrode-io/electrode-native/issues/498

### DIFF
--- a/ern-core/src/resolveNativeDependenciesVersions.js
+++ b/ern-core/src/resolveNativeDependenciesVersions.js
@@ -1,0 +1,128 @@
+import semver from 'semver'
+import _ from 'lodash'
+import { PackagePath } from 'ern-core'
+import { NativeDependencies } from './nativeDependenciesLookup'
+import MiniApp from './MiniApp'
+
+export function containsVersionMismatch (
+  versions: Array<string>,
+  mismatchLevel: 'major' | 'minor' | 'patch') : boolean {
+  const minVersion = semver.minSatisfying(versions, '*')
+  const maxVersion = semver.maxSatisfying(versions, '*')
+  const majorMismatch = semver.major(maxVersion) !== semver.major(minVersion)
+  const minorMismatch = semver.minor(maxVersion) !== semver.minor(minVersion)
+  const patchMismatch = semver.patch(maxVersion) !== semver.patch(minVersion)
+  return majorMismatch ||
+        (minorMismatch && (mismatchLevel === 'minor' || mismatchLevel === 'patch')) ||
+        (patchMismatch && mismatchLevel === 'patch')
+}
+
+export function retainHighestVersions (
+  dependenciesA: Array<PackagePath>,
+  dependenciesB: Array<PackagePath>) : Array<PackagePath> {
+  const result : Array<PackagePath> = []
+  const groups = _.groupBy([...dependenciesA, ...dependenciesB], 'basePath')
+  for (const dependencyGroup of Object.values(groups)) {
+    let highestVersionDependency = dependencyGroup[0]
+    if ((dependencyGroup.length > 1) && (semver.gt(dependencyGroup[1].version, dependencyGroup[0].version))) {
+      highestVersionDependency = dependencyGroup[1]
+    }
+    result.push(highestVersionDependency)
+  }
+  return result
+}
+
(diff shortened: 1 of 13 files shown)


krunalsshah:
@belemaire sorry although I thought ```Given that there is no conflicting version (or force flag was used) we just retain the latest version of each dependency from the list of dependencies (we also include all current native dependencies listed in the Container Cauldron).``` was always true :)

ern-core/src/resolveNativeDependenciesVersions.js
```diff
@@ -0,0 +1,132 @@
...
+        (patchMismatch && mismatchLevel === 'patch')
+}
+
+export function retainHighestVersions (
+  dependenciesA: Array<PackagePath>,
+  dependenciesB: Array<PackagePath>) : Array<PackagePath> {
+  const result : Array<PackagePath> = []
+  const groups = _.groupBy([...dependenciesA, ...dependenciesB], 'basePath')
+  for (const dependencyGroup of Object.values(groups)) {
+    if (dependencyGroup.length > 1) {
```
krunalsshah:
```if(dependencyGroup.length > 1 && semver.lte(dependencyGroup[0].version, dependencyGroup[1].version)){
result.push(dependencyGroup[1])
} else {
result.push(dependencyGroup[0])
}```
We can change this to ^^^, just inverting the condition.

File: ern-local-cli/src/commands/cauldron/batch.js [hunk 2]
```diff
@@ -184,24 +185,48 @@ exports.handler = async function ({
...
+        nonUpdatedMiniAppsInCauldronObjs.push(m)
+      }
+
+      const nativeDependencies = await resolver.resolveNativeDependenciesVersionsOfMiniApps([...updateMiniAppsObjs, ...addMiniAppsObjs, ...nonUpdatedMiniAppsInCauldronObjs])
+      const cauldronDependencies = await cauldron.getNativeDependencies(napDescriptor)
+      const finalNativeDependencies = resolver.retainHighestVersions(nativeDependencies.resolved, cauldronDependencies)
+
+      const conflictingDependencies = nativeDependencies.pluginsWithMismatchingVersions
+      if (conflictingDependencies.length > 0) {
+        if (force) {
```
krunalsshah: minor: this block appears at multiple places, may be a contender for externalizing in utils? Again not absolute necessary, just if it makes sense.

belemaire:
@krunalsshah
Addressed your two comments regarding refactoring to make things leaner.
I don't really understand what you mean by `this is applicable for git instead of npm package`.
The resolution mecanism works for both for git and npm packages in term of miniapps, it can be mixed. Previous known behavior with npm packages remains similar.

Also please only review PR and do not merge, will battle test in dev version for next 2 weeks, planning for `0.16.0` release.

belemaire:
v0.15.0 was just released, removing `DO NOT MERGE` label for this PR.
Can be merged upon review approval, will be battle tested for the next 2 weeks on dev platform.
Scheduled for v0.16.0 release.

krunalsshah:
👍

PR #683, opened 2018-03-14
Code push custom server

deepueg:
This PR adds the ability to pass custom configs to codePush SDK. Allowing you to specify your serverUrl for the OTA.

### DIFF
--- a/ern-core/src/CodePushSdk.js
+++ b/ern-core/src/CodePushSdk.js
@@ -22,13 +22,20 @@ export type CodePushReleaseInfo = {
   /* generated */ uploadTime: number;
 };
 
+export type CodePushInitConfig = {
+  accessKey: string,
+  customHeaders?: {[headerName: string]: string},
+  customServerUrl?: string,
+  proxy?: string
+}
+
 export type CodePushPackage = CodePushPackageInfo & CodePushReleaseInfo;
 
 export default class CodePushSdk {
   _codePush: any
 
-  constructor (accessKey: string) {
-    this._codePush = new CodePush(accessKey)
+  constructor (initConfig: CodePushInitConfig) {
+    this._codePush = new CodePush(initConfig.accessKey, initConfig.customHeaders, initConfig.customServerUrl, initConfig.proxy)
   }
 
   async releaseReact (
--- a/ern-local-cli/src/lib/publication.js
+++ b/ern-local-cli/src/lib/publication.js
@@ -20,7 +20,8 @@ import {
   reactnative
 } from 'ern-core'
 import type {
-  CodePushPackage
+  CodePushPackage,
+  CodePushInitConfig
 } from 'ern-core'
 
 import inquirer from 'inquirer'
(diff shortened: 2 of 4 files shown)


ern-core/src/CodePushSdk.js
```diff
@@ -22,13 +22,20 @@ export type CodePushReleaseInfo = {
   /* generated */ uploadTime: number;
 };
 
+export type CodePushInitConfig = {
+  accessKey: string,
+  customHeaders?: {[headerName: string]: string},
```
krunalsshah: 👍

ern-core/src/CodePushSdk.js
```diff
@@ -22,13 +22,20 @@ export type CodePushReleaseInfo = {
...
+
 export type CodePushPackage = CodePushPackageInfo & CodePushReleaseInfo;
 
 export default class CodePushSdk {
   _codePush: any
 
-  constructor (accessKey: string) {
-    this._codePush = new CodePush(accessKey)
+  constructor (initConfig: CodePushInitConfig) {
+    this._codePush = new CodePush(initConfig.accessKey, initConfig.customHeaders, initConfig.customServerUrl, initConfig.proxy)
```
krunalsshah: is it possible that properties may be undefined/null? In that case , let MSFT code-push client take care of default value and restrict the properties sent.

belemaire: @krunalsshah 
this will happen anyway for any `undefined` as if let's say only access key is specified, the call will end up being `new CodePush('ACCESS_KEY', undefined, undefined, undefined)` which JS will translate anyway to `new CodePush('ACCESS_KEY')`. 
I would just add a check if `accessKey` is not undefined/null or empty string and throw here, as it is the only mandatory parameter.

ern-local-cli/src/lib/publication.js
```diff
@@ -546,21 +547,23 @@ async function getCodePushAppName (
...
-export function getCodePushAccessKey () {
+export function getCodePushInitConfig (): CodePushInitConfig {
   const codePushConfigFilePath = path.join(process.env.LOCALAPPDATA || process.env.HOME || '', '.code-push.config')
+  var codePushInitConfig: CodePushInitConfig
   if (fs.existsSync(codePushConfigFilePath)) {
-    return JSON.parse(fs.readFileSync(codePushConfigFilePath, 'utf-8')).accessKey
+    codePushInitConfig = JSON.parse(fs.readFileSync(codePushConfigFilePath, 'utf-8'))
   } else {
-    return config.getValue('codePushAccessKey')
+    codePushInitConfig = {'accessKey': config.getValue('codePushAccessKey')}
```
belemaire: could you also add headers/custom url/proxy here.
you can just assume the config properties to be prefixed by `codePush` as we did for access key. we will document that later that's ok, but at least the code is there. 
this can be needed in CI environments if we don't want to install the Code Push CLI on the box (and therefore don't end up having a `.code-push.config` on the box)

krunalsshah: +1

deepueg:
@krunalsshah @belemaire Done.

ern-core/src/CodePushSdk.js
deepueg: Access is validated here https://github.com/electrode-io/electrode-native/pull/683/files#diff-659728c901f8b033afe49b670ea0229aR568

krunalsshah:
👍

Issue #389, opened 2017-11-27
Git wrapper should support interactive prompts


belemaire:
Our git wrapper currently swallow all stdout output and does not redirect stdin. 

What it means is that when using HTTPs URLs for repositories (be it a new Cauldron, or even the manifest which is an HTTPS url), if the GIT client is not properly logged in (username/password not stored in keychain), the GIT client will display a prompt for user name / password to login. This prompt is not displayed to the user, thus the command will appear as stalling.

We should properly display this prompt and make sure stdin is properly handled, so that if that prompt is displayed, user can enter username/password and proceed with the command.

PR #684, opened 2018-03-14
Force authentication details to be part of an https cauldron url

belemaire:
Close https://github.com/electrode-io/electrode-native/issues/389

Initial idea was to redirect stdin to ask the user for credentials, but it complicates things a lot. Also it can lead to issue when people try to do that on CI. So to make things more robust, the simple solution is just to ensure that users provide the credentials as part of the HTTPS url. 

If users don't want their username/password to be used, they can use a token instead, and still if they don't want that, then they can use an SSH url ;)


gmbharath12:
👍 gotcha.

PR #701, opened 2018-03-28
force cleaning of package directory in node_modules after moving the …

deepueg:
…contents to root directory

**Issue**: When miniapp is looking for it's native dependencies the miniapp name is also returned in the list. 

**Root cause**: This was caused because of the `electrode-native-bridge` directory was still remaining inside `node_modules/<miniappname>` directory even after the [move](https://github.com/electrode-io/electrode-native/compare/master...deepueg:fix-create-container-issue?expand=1#diff-45142d217dea7e95458dcef2b4c18fb2R90) command execution. 

**Solution**: Ensure the `node_modules/<miniappname>` directory is emptied after the move.


krunalsshah:
@deepueg is it specific to the issue you were mentioning , w.r.t `yarn.lock` file missing etc?

PR #714, opened 2018-04-09
Store Container JS bundle and assets in Cauldron

belemaire:
Whenever a new Container is generated from a Cauldron, the resulting JS bundle and assets are now zipped and added to the Cauldron, using a filename convention based on the native application descriptor.

For example, when generating a new Container for the following native application descriptor `MyApp:android:1.0.0`, the JS bundle and assets will be zipped, resulting in a zip file named `MyApp:android:1.0.0.zip` which is then stored in the Cauldron, at the root, in the `bundles` directory. Any new Container generation for a given descriptor will generate a new zip file that will overwrite the one stored for this native application descriptor (if any).

This feature isn't really useful on its own but allows for future improvements that will make use of this zip files in the context of CodePush (to reduce the size of CodePush updates by diffing bundle and assets).

### DIFF
--- a/ern-core/src/ReactNativeCli.js
+++ b/ern-core/src/ReactNativeCli.js
@@ -12,6 +12,17 @@ import {
 } from './childProcess'
 const fetch = require('node-fetch')
 
+export type BundlingResult = {
+  // The root path to the assets
+  assetsPath: string;
+  // The target platform of the bundle
+  platform: string;
+  // Indicates whether this is a dev bundle or a production one
+  dev: boolean;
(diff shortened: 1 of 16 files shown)


ern-core/src/ReactNativeCli.js
```diff
@@ -12,6 +12,17 @@ import {
 } from './childProcess'
 const fetch = require('node-fetch')
 
+export type BundlingResult = {
```
krunalsshah: `BundleMetaData`?

PR #727, opened 2018-04-27
Add Visual Studio Code workspace settings

belemaire:
This way, developers using Visual Studio Code for electrode-native development, will have proper workspace settings applied.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,9 +21,6 @@ node_modules/
 npm-debug.log
 */node_modules
 
-# VSCode
-.vscode
-
 # distrib
 distrib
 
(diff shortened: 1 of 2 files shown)


krunalsshah:
@belemaire thanks for the PR. Although do we want to package `IDE` specific config files to the `en` project?

belemaire:
Yes, that's a valid concern. I thought about it for some time as well (that's why .vscode was in the .gitignore from the start).

But we are almost all using Visual Studio Code in the team, and it's one of the most popular free IDE out there so we can can expect the majority of contributors to use it as their IDE. It does not pollute the git repo too much as it is a single directory/file, and ensure that developers using .vscode will get the proper settings for electrode-native development (personally had to redo the settings manually on three different workstations for dev. its a pain).

People that are not using .vscode won't be bothered too much by this. It's a hidden directory anyway.

PR #728, opened 2018-05-01
Fix warning for ElectrodeContainer

clairewli:
Fix warnings inside ElectrodeContainer.

### DIFF
--- a/ern-container-gen/src/generators/ios/hull/ElectrodeContainer/ElectrodeReactNative.m
+++ b/ern-container-gen/src/generators/ios/hull/ElectrodeContainer/ElectrodeReactNative.m
@@ -1,12 +1,12 @@
 /*
  * Copyright 2017 WalmartLabs
- 
+
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- 
+
  * http://www.apache.org/licenses/LICENSE-2.0
- 
+
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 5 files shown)


ern-container-gen/src/generators/ios/hull/ElectrodeContainer/ElectrodeReactNative.m
```diff
@@ -190,7 +190,7 @@ - (void)startContainerWithConfiguration:(id<ElectrodePluginConfig>)reactContaine
 }
 
 - (void)loadCustomFonts {
-    NSMutableArray *fontPaths = [[NSBundle frameworkBundle] pathsForResourcesOfType:nil inDirectory:nil];
+    NSMutableArray *fontPaths = [[[NSBundle frameworkBundle] pathsForResourcesOfType:nil inDirectory:nil] mutableCopy];
```
gmbharath12: @clairewli 
Why do you want to `mutableCopy` here? Is the object not retained?
Doing `mutableCopy` increases the retain count and hence this is a memory leak..

gmbharath12: `pathsForResourcesOfType` returns an array that contains path names for all bundle resources (in this case). If nothing is matched it returns an empty array which is fine. 
So we shouldn't be doing `mutableCopy`.

ern-container-gen/src/generators/ios/hull/ElectrodeContainer/ElectrodeReactNative.m
```diff
@@ -190,7 +190,7 @@ - (void)startContainerWithConfiguration:(id<ElectrodePluginConfig>)reactContaine
 }
 
 - (void)loadCustomFonts {
-    NSMutableArray *fontPaths = [[NSBundle frameworkBundle] pathsForResourcesOfType:nil inDirectory:nil];
+    NSArray *fontPaths = [[NSBundle frameworkBundle] pathsForResourcesOfType:nil inDirectory:nil];
```
gmbharath12: good one 👍

PR #737, opened 2018-05-08
Update API gen and container for swift 4

clairewli:
Update API gen and container for swift 4


ern-api-gen/resources/swift/libraries/ern/api.mustache
```diff
@@ -67,3 +137,4 @@ public class {{classname}}Requests: NSObject {
     {{/operations}}
 }
 
+#endif
```
belemaire: Looking at this file, it seems like the only difference between swift3/swift4 is about prefixing the class declaration with `@objcMembers`. The rest of the code is exactly the same (might be wrong just had a quick look).
If that's the case no need to duplicate the whole source, as its harder to see where are the difference between swift3/4 in the file. It also complexifies maintenance as everything is duplicated.
What I would do, given the minimal differences is just surround the classes declarations with the preprocessor directive, i.e :

```
#if swift(>=4.0)
@objcMembers public class {{classname}}: NSObject  {
#else
public class {{classname}}: NSObject  {
#endif
```

ern-container-gen/src/generators/ios/hull/ElectrodeContainer.xcodeproj/project.pbxproj
```diff
@@ -416,16 +420,14 @@
...
 			};
 			name = QADeployment;
 		};
 		303268E61EA8139700A41A8F /* QADeployment */ = {
 			isa = XCBuildConfiguration;
 			buildSettings = {
-				FRAMEWORK_SEARCH_PATHS = (
-					"$(inherited)"
-				);
+				FRAMEWORK_SEARCH_PATHS = "$(inherited)";
```
belemaire: Please update all `FRAMEWORK_SEARCH_PATHS` in this file as they were before this commit (i.e surrounded by parenthesis). I think that our pbxproj parser is quite picky on this :(
Don't have time atm to look in the parser and ways to improve it, so just to be on the safe side let's just format these the way they were.

belemaire:
Thanks 👍

PR #740, opened 2018-05-09
Enable CodeLens for TypeScript implementations

belemaire:
Improve development experience in Visual Studio Code by enabling CodeLens for TypeScript implementations. It makes it easy to see if an interface has implementations, and to navigate to the implementations conveniently.

**BEFORE**
screen shot 2018-05-09 at 11 06 59 am

**AFTER**
screen shot 2018-05-09 at 11 06 46 am


gmbharath12:
wow 👍

PR #744, opened 2018-05-10
Fix TravisCI cron job for system tests

belemaire:


belemaire:
Self merging this one given that there is not much to review and CI cron job needs a fix asap.

PR #752, opened 2018-05-15
Fix issue in api-impl-gen related to bridge version range

belemaire:
This PR fix a bug that was reported by daily job system tests failure.

A bug was recently discovered and fixed in GitManifest (comparing versions as strings instead of using semver lib) : https://github.com/electrode-io/electrode-native/commit/2e7d23272db2c0783654b921850ce68c6c0f3300#diff-2fde8a8f6c7fa832045033cb61368189L203

Fixing this bug actually uncovered another issue/bug in api-impl-gen.
The problem is that api-impl-gen is installing the api dependency along with its dependencies. APIs have a dependency on bridge, but not on a fixed version of bridge, rather a range (`1.5.x`). `yarn add` call goes fine with such a range dependency but at some point in GitManifest, when asking for the plugin configuration, the manifest resolver fails on the call to `semver.gte` as this function is used to compare two versions (fixed), and a range well, it's not a version ...

Quick fix is that instead of providing the dependencies as such, with ranged versions, we just install all of them, and then, for all dependencies versions that were using a range, have a peek in the resulting yarn.lock to figure out the fixed versions that were actually installed. We then rebuild a dependency array (PackagePath array) matching the original one, at the difference that all dependencies with ranged versions now have a real fixed version (the one that was truly installed), making such an array ready to be fed to semver comparators.

GitManifest should have some guard clauses to ensure that all dependencies it is fed with are containing fixed versions and not range, should throw exception if not, in order to fail early. Will beef up Manifest with guards soon.

This is fixing the current bug as well as the impacted system test. That being said, system tests are still failing due to a recent new bridge version publication (1.5.13) and system test fixtures not having been regenerated. https://github.com/electrode-io/electrode-native/pull/753 is containing the regenerated fixtures.

### DIFF
--- a/ern-api-impl-gen/src/generators/ApiImplGen.ts
+++ b/ern-api-impl-gen/src/generators/ApiImplGen.ts
@@ -4,6 +4,8 @@ import _ from 'lodash'
 import chalk from 'chalk'
 import path from 'path'
 import fs from 'fs'
+import semver from 'semver'
+import * as lockfile from '@yarnpkg/lockfile'
 import ApiImplAndroidGenerator from './android/ApiImplAndroidGenerator'
 import ApiImplIosGenerator from './ios/ApiImplIosGenerator'
 import ApiImplJsGenerator from './js/ApiImplJsGenerator'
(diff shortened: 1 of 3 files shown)


ern-api-impl-gen/src/generators/ApiImplGen.ts
```diff
@@ -79,23 +81,66 @@ export default class ApiImplGen {
     reactNativeVersion: string
   ) {
     try {
-      shell.cd(downloadPath)
+      shell.pushd(downloadPath)
```
belemaire: Better to use `pushd` along with `popd` in `finally` block rather than `cd`. With `cd` we can easily lose track of which directory we are in if not properly rolled back in `finally`. If using `cd` with `finally` we still need to back up the current directory before running `cd` (so that we can roll back in `finally`) which is cumbersome and easy to forget.

PR #757, opened 2018-05-15
Update yarn-lock-check script

belemaire:
Given that we are now using [Yarn Workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) there is no longer a need to check for validity of modules lock files because due to the nature of yarn workspaces, there is just one top level lock file and no module lock files. 

From Yarn Workspaces documentation :
> Yarn will use a single lockfile rather than a different one for each project, which means less conflicts and easier reviews.

This PR updates yarn-lock-check script so that it doesn't look for lock files in modules directories and just check the validity of the top level project lock file.

### DIFF
--- a/yarn-lock-check.js
+++ b/yarn-lock-check.js
@@ -6,61 +6,42 @@ const path = require('path')
 const yarnLock = 'yarn.lock'
 const yarnRegistryRe = /https:\/\/registry.yarnpkg.com|http:\/\/registry.npmjs.org/
 
-let ernDirs = fs.readdirSync(process.cwd()).filter(x => x.startsWith('ern'))
-ernDirs.push(yarnLock) // Root yarn.lock
+const yarnLockContent = fs.readFileSync(yarnLock, 'utf8')
+const yarnLockParsed = lockfile.parse(yarnLockContent)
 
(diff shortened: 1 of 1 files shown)


yarn-lock-check.js
```diff
@@ -6,61 +6,49 @@ const path = require('path')
...
-  } catch (e) {
-    console.log(chalk.yellow(`Skip yarn-lock registry check for ${current}`))
   }
-})
+
+  /*
+    yarn.lock contains registry other than https://registry.yarnpkg.com
+    http://npme.walmart.com/@yarnpkg/lockfile/-/lockfile-1.0.0.tgz#33d1dbb659a23b81f87f048762b35a446172add3
+    http://npme.walmart.com/colors/-/colors-1.1.2.tgz#168a4701756b6a7f51a12ce0c97bfa28c084ed63
+    http://npme.walmart.com/diff/-/diff-3.4.0.tgz#b1d85507daf3964828de54b37d0d73ba67dda56c
```
deepueg: perhaps we should remove Walmart npme reference ?

belemaire:
@deepueg Addressed your comment by removing references in comments to internal npm walmart repo.

Issue #257, opened 2017-10-31
Container build configuration injection


belemaire:
Right now, the ios QA build config section is present in the container hull. This is a Walmart specific build config and should not be present in the generic container. Only Debug/Release (default standard build config) should be included in the generic container.

Any custom build config used for the target host application should be injected in some way, but not part of the generic container which should be company agnostic and just contain the Debug/Release default build configs.

belemaire:
This could be achieved by extracting the build configuration settings out of the iOS Container hull pbxproj into distinct xcconfig files. This was, native application can override some build settings with their own more easily.

Link to build setting extractor tool that could prove useful :
https://github.com/dempseyatgithub/BuildSettingExtractor

We will look into figuring out a proper solution for this, targeting v0.20.0 if all goes smooth

belemaire:
Build configuration has been extracted out of the Container and a [build configuration transformer](https://github.com/electrode-io/ern-container-transformer-build-config/blob/master/README.md) has been created. All that is left is to get rid of QA config and inject it in Walmart Container through the transformer

PR #760, opened 2018-05-16
Fix bundle naming ios

deepueg:


belemaire:
@nonsensery 
Sorry for late answer. We changed it to match the namespace we use on android for parity and coherence. That being said, we are currently working on https://github.com/electrode-io/electrode-native/issues/257 which once implemented will allow any user to customize (override) build settings of generated Containers based on their needs and wishes. Bundle identifier being part of the build settings, once we get this one in, you should be able to use a bundle identifier for your Container that is coherent with the rest of your native app, if that is what you'd like.
Will ping back on this issue once this feature is released.

PR #764, opened 2018-05-17
[code-push][container-gen] add the code push init code to JS side if …

### LINKED ISSUE #301: Extract CodePush logic out of MiniApps
As of now (internally), CodePush logic is directly embedded in the MiniApps. This is not a good practice as it makes it harder to reuse MiniApps with different CodePush configuration settings in different applications. 

MiniApps should not have to depend on CodePush. 
CodePush should be included at the top level if needed (composite MiniApp / Container) and the CodePush configuration logic should be moved out of MiniApps.

We then can offer a global CodePush configuration logic, while keeping CodePush settings per MiniApps if needed. This could be achieved either as making a base Component that MiniApps should have to inherit from (instead of the base React one), or we could also inject the necessary code at the composite MiniApp level / container generation, to gain more flexibility and not enforce MiniApp developers to have to inherit from a custom component.

More thinking / POC need to go into this one to figure out a proper solution and implement it.

deepueg:
…the generated container has code push plugin inside it. This helps to eliminate the steps to add the starter code to each mini app.

fixes #301

### DIFF
--- a/ern-container-gen/src/resources/CodePushStarterMiniApp.file
+++ b/ern-container-gen/src/resources/CodePushStarterMiniApp.file
@@ -1,0 +1,15 @@
+import { Component } from 'react'
+import codePush from 'react-native-code-push'
+
+export default class App extends Component<Props> {
+  render() {
+    return null
+  }
+}
+
+App = codePush({
+  checkFrequency: codePush.CheckFrequency.ON_APP_RESUME,
+  installMode: codePush.InstallMode.ON_NEXT_RESUME,
+  minimumBackgroundDuration: 60 * 2,
+})(App)
+codePush.sync()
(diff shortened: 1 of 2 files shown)


ern-container-gen/src/resources/CodePushStarterMiniApp.file
```diff
@@ -0,0 +1,15 @@
...
+    return null
+  }
+}
+
+App = codePush({
+  checkFrequency: codePush.CheckFrequency.ON_APP_RESUME,
+  installMode: codePush.InstallMode.ON_NEXT_RESUME,
+  minimumBackgroundDuration: 60 * 2,
+})(App)
+codePush.sync()
```
belemaire: Seems like we also need to pass the install mode here ... per the documentation : https://github.com/Microsoft/react-native-code-push/blob/master/docs/api-js.md#codepushsync

deepueg:
Closing this PR as I am testing this further now. Will reopen once ready.

PR #768, opened 2018-05-21
Disable Prettier for json files

belemaire:
Don't want Prettier to mess with json files formatting, especially package.json, so disabling it for json files.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,4 +1,5 @@
 **/*.md
+*.json
 **/coverage
 **/docs
 **/.coverage


krunalsshah:
@belemaire what about the files already formatted using prettier?

krunalsshah:
@belemaire is #769 part of it? clean up etc.?

belemaire:
@krunalsshah 
No a problem for json files already formatted with Prettier, it will be on a file per file basis if json formatting does not look right, but it's mostly ok with me as it is. was just a pain for package.json where it's better to have single liners for scripts.
#769 is partly made possible because prettier is disabled for json, otherwise I would have had no way to push package.json as it is formatted (actually I force commited with -n but still, if prettier is kept activated, it will reformat the package.json as it was before as soon as anyone commit, as it will run prettier on the json ... so better to merge it before this one, or at least both around similar time).

PR #776, opened 2018-05-31
Add NativePlatform type

belemaire:
We were using type `'android' | 'ios'` in multiple locations in the code to type variables representing a platform.

Multiple problems with this approach :
- Duplication
- Prone to error (mistyping)
- Painful if ever a new native platform is supported

Therefore, moving this one as a dedicated type `NativePlatform` in `ern-core`.

Let's make sure that everywhere we see a variable holding a native platform name in the code, we use this type rather than plain old `string` type (primitive type obsession anti-pattern). In this PR I just replaced all locations where `'android' | 'ios'` (or reverse) were used. I did not took care of the one variables that are using `string` (harder to locate).

### DIFF
--- a/ern-container-gen/src/utils.ts
+++ b/ern-container-gen/src/utils.ts
@@ -15,6 +15,7 @@ import {
   ModuleTypes,
   utils,
   log,
+  NativePlatform,
 } from 'ern-core'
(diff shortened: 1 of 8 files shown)


belemaire:
Closing in favor of https://github.com/electrode-io/electrode-native/pull/777 which contains this commit.

PR #805, opened 2018-06-14
Iterate babel resolver plugin config correctly for multiple miniapps

gmusick:
If multiple miniapps have a `package.json` babel config, only the first miniapp's config gets used. This is happening because of `for ... in` instead of `for ... of` on the `babelPlugin` array.


gmusick:
👍 

Question: do you have an idea when this will get released? I'm using `1000.0.0` right now, but would like to use a released version. Thanks!

belemaire:
@gmusick 
0.20 release is planned for Monday, would that work for you ?
Otherwise we can do a patch release of 0.19.1 just to include your fix, but given that it is not a bug with a wide impact to users (module resolvers are something very specific) and that the release date is close, I don't think we will go down that road.

gmusick:
@belemaire yep, that works. Thanks!

belemaire:
@gmusick 
FYI we've just released 0.20.0 which includes your fix.

Issue #811, opened 2018-06-18
Manifest should support iOS directive to declare target dependencies.


cihancimen:
Similar to adding a linked library, `addProject` directive should support adding target dependencies.
Following directive should add a target dependency, without changing `Link Binary with Libraries` build phase.
```
"addProject": [
 { 
          "path": "../../../../../Modules/WalmartIOSShared/WalmartIOSShared.xcodeproj", 
          "group": "TargetDependencies", 
          "frameworks": [ { "name": "WalmartIOSShared.framework", "target": "WalmartIOSShared", "productReference": "07C5A03F1C472F4600157A12" } ] 
        }
]
```

belemaire:
@cihancimen 
Just released v0.20.0 which includes the necessary in https://github.com/electrode-io/electrode-native/pull/813. Was a bit trickier than planned but its working. Got manifest changes ready, will follow up with you and @gmbharath12 tomorrow.

PR #813, opened 2018-06-19
Add new option addAsTargetDependency to iOS addProject plugin directive

belemaire:
- Bump `xcode-ern` dependency version to `1.0.3` which implements the logic for supporting this new option flag.


belemaire:
Auto approving this one as it needs to be shipped in v0.20.0 NOW.

belemaire:
Fix https://github.com/electrode-io/electrode-native/issues/811

PR #825, opened 2018-06-26
Cache packages to boost Container generator speed

belemaire:
This PR introduces a simple caching mechanism to cache packages used during Container generation (MiniApps, APIs & APIs implementations, native modules).

As a baseline, for a big sample production Container, if not changing anything in the Container but just calling `cauldron regen-container`, time to generate a Container goes from ~10min to ~6min (~40% decrease). This is the ideal scenario, with all packages versions being cached. 

Speed gains decrease linearly with the number of non-cached packages versions.

The cache has a default size of 2GB (for this big sample production Container, all packages at a specific version represent ~1GB, so this leaves us room to cache a few different versions). The package cache max size is configurable through `platform config max-package-cache-size`. Once max cache size is exceeded, the least recently accessed package(s) will be removed from the cache until total cache size gets below max cache size.

The package cache is disabled by default, and still experimental, so not yet documented. It will be shipped in Electrode Native 0.21.0, battle tested internally and will be moved out of experimental phase and documented in Electrode Native 0.22.0 most probably.

Enabling the package cache can be done through `platform config package-cache-enabled true`.

The packages cache directory is located inside the `.ern` folder in `packages-cache`.

Most of Container Generation time is now spent for composite MiniApp generation and bundling. This is a little bit trickier to cache but possible. That being said, we won't really be able to decrease bundling speed (if we stick to Metro) as it is out of our control. But caching composite prior to bundling could add a few minutes gain, probably making our reference container generation from ~10min down to ~3-4min.

### DIFF
--- a/ern-core/src/FsCache.ts
+++ b/ern-core/src/FsCache.ts
@@ -1,0 +1,220 @@
+import fs from 'fs'
+import path from 'path'
+import { readJSON, writeJSON, writeJSONSync } from './fileUtil'
+import shell from './shell'
+import uuidv4 from 'uuid/v4'
+import getSize from 'get-folder-size'
+
+/**
+ * Gets the total size (in bytes) of a given directory
+ * @param dirPath path to the directory for which to compute size
+ */
+const getSizeAsync: (dirPath: string) => Promise<number> = dirPath =>
+  new Promise((resolve, reject) => {
+    getSize(dirPath, (err, size) => {
+      err ? reject(err) : resolve(size)
+    })
+  })
+
+export interface CacheManifest {
+  /**
+   * Entries stored in the cache
+   */
+  entries: CacheManifestEntry[]
+  /**
+   * Total size of the cache
+   */
+  size: number
+}
+
+export interface CacheManifestEntry {
+  /**
+   * Id of the entry
+   * ex: my-miniapp@1.0.0
+   */
+  id: string
+  /**
+   * Absolute path to the entry
+   * ex : /Users/blemair/.ern/cache/78c40fda-d5bb-4681-9bd2-a5fba9df1f70
+   */
+  path: string
+  /**
+   * Size in bytes of the entry
+   */
+  size: number
+  /**
+   * Time (epoc) when the entry was last accessed
+   */
+  lastAccessed: number
+}
+
+const cacheManifestFileName = 'cache-manifest.json'
+export const maxDefaultCacheSize = 2 * 1024 * 1024 * 1024 /* 2GB */
+
+type AddObjectToCacheDirectory<T> = (obj: T, dirPath: string) => Promise<void>
+type ObjectToId<T> = (obj: T) => string
+
+/**
+ * Simple generic file system based LRU cache.
+ * This cache stores arbirtrary objects in unique directories.
+ * The root directory for the cache can be configured, as well as
+ * the maximum size allowed for the cache.
+ * Once cache reaches max size, the least recenctly accessed
+ * objects will be removed from the cache, until size of the cache
+ * falls below maximum size.
+ */
+export class FsCache<T> {
+  /**
+   * Root path to the directory holding the cache
+   */
+  public readonly rootCachePath: string
+  /**
+   * Maximum cache size (in bytes)
+   */
+  public readonly maxCacheSize: number
+  /**
+   * Custom function invoked to add an object entry to the cache
+   */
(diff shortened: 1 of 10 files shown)


ern-core/src/FsCache.ts
```diff
@@ -0,0 +1,220 @@
...
+      await writeJSON(this.cacheManifestPath, manifestObj)
+      return cacheEntry.path
+    }
+  }
+
+  /**
+   * Gets the cache manifest object
+   */
+  private async getManifestObj(): Promise<CacheManifest> {
+    return readJSON(this.cacheManifestPath)
```
krunalsshah: readJSONSync()?

ern-core/src/FsCache.ts
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * Time (epoc) when the entry was last accessed
+   */
+  lastAccessed: number
+}
+
+const cacheManifestFileName = 'cache-manifest.json'
+export const maxDefaultCacheSize = 2 * 1024 * 1024 * 1024 /* 2GB */
+
+type AddObjectToDirectoryFun<T> = (obj: T, dirPath: string) => Promise<void>
```
krunalsshah: `AddObjectToDirectoryFun` is not clear , sorry :(

ern-core/src/FsCache.ts
```diff
@@ -0,0 +1,220 @@
...
+    await writeJSON(this.cacheManifestPath, manifestObj)
+    return pathToObjectCacheDir
+  }
+
+  /**
+   * Gets path to the directory containing a cached object
+   * @param obj The object for which to retrieve cached directory path
+   */
+  public async getObjectCachePath(obj: T): Promise<string | void> {
+    const manifestObj = await this.getManifestObj()
```
krunalsshah: If it's possible to read `return readJSONSync(this.cacheManifestPath)`, we can `await writeJSONSync(this.cacheManifestPath, manifestObj)` (line 193) ? In which case we can avoid `await`

ern-core/src/FsCache.ts
```diff
@@ -0,0 +1,220 @@
...
+      writeJSONSync(this.cacheManifestPath, { entries: [], size: 0 })
+    }
+  }
+
+  /**
+   * Indicates whether a given object is currently stored in the cache
+   * @param obj The object to check for presence in the cache
+   */
+  public async isInCache(obj: T): Promise<boolean> {
+    const manifestObj = await this.getManifestObj()
```
krunalsshah: If it's possible for `getManifestObj` to `readJSONSync()` we an can eliminate need for `async`

ern-core/src/FsCache.ts
```diff
@@ -0,0 +1,220 @@
...
+import { readJSON, writeJSON, writeJSONSync } from './fileUtil'
+import shell from './shell'
+import uuidv4 from 'uuid/v4'
+import getSize from 'get-folder-size'
+
+/**
+ * Gets the total size (in bytes) of a given directory
+ * @param dirPath path to the directory for which to compute size
+ */
+const getSizeAsync: (dirPath: string) => Promise<number> = dirPath =>
```
krunalsshah: Treat this comment as informational, why did we have to read it into a `sync` block?

krunalsshah:
👍 thanks for the optimization.

ern-core/src/FsCache.ts
belemaire: This is an `async` method, it returns a `Promise`. `readJson` returns a Promise so nothing wrong here.

ern-core/src/FsCache.ts
belemaire: We should use asynchronous methods as much as possible in the code base. 
Async (promise based) is the way to go everywhere doable.
Unfortunately, constructors cannot be executed asynchronously, thus there was no other choice but to choose sync version in the constructor (unless change of design with lazy init, but wanted to keep things simple). That's not a big issue.

ern-core/src/FsCache.ts
belemaire: Agreed, will rename to `AddObjectToCacheDirectory`

ern-core/src/FsCache.ts
belemaire: Same as previous comment.

belemaire:
Pushed changes based on review :
- Rename `AddObjectToDirectoryFun` to `AddObjecToCacheDirectory`

PR #827, opened 2018-06-27
[container][android] set app properties on a react root view if a vie…

deepueg:
…w was already created for the given miniapp


deepueg:
the test fixture is already updated. 

Re the RN version, this feature is introduced in RN version 45. I will add a filter to the code.

deepueg:
@krunalsshah @belemaire Fixed.

PR #843, opened 2018-07-07
Introduce transform-container command

belemaire:


ern-local-cli/src/commands/transform-container.ts
```diff
@@ -0,0 +1,90 @@
...
+      describe: 'Local path to the Container to transform',
+      type: 'string',
+    })
+    .option('config', {
+      alias: 'c',
+      describe:
+        'Optional transformer configuration (json string or path to config file)',
+      type: 'string',
+    })
+    .option('platform', {
```
krunalsshah: alias missing `-p`

ern-local-cli/src/commands/transform-container.ts
```diff
@@ -0,0 +1,90 @@
...
+export const handler = async ({
+  containerPath,
+  config,
+  platform,
+  transformer,
+}: {
+  containerPath?: string
+  config?: string
+  platform: NativePlatform
+  transformer: string
```
krunalsshah: throw error if `transformer` is null or undefined

belemaire: It cannot be null or undefined because we set the `yargs` option `demandOption: true`.
With that option set, it is not possible for the user not to provide `-transformer/--t`. The command will fail with error thrown by `yargs` : `Missing required arguments:  transformer`

ern-local-cli/src/commands/transform-container.ts
belemaire: 👍 done

PR #857, opened 2018-07-17
Cauldron file system

belemaire:
Add support to store arbitrary files in Cauldron. 
This PR introduces three new cauldron subcommands :
- `ern cauldron add file`
- `ern cauldron update file`
- `ern cauldron delete file`

For example, running `ern cauldron add file /Users/username/test.json data/json/test.json` will add the `test.json` in the Cauldron in the `data/json` directory from the root of the Cauldron (creating the directori(es) if needed).

First use case for this will probably be to store publishers and/or transformers configurations (as transformers configurations can grow quite big and bloat main cauldron document, also better to keep them separate for easier management). Also for example the planned `script` transformer will allow the execution of arbitrary scripts which can get quite big (fat-binary for iOS being one). It will be a pain to store it directly embedded in the Cauldron document, and a pain also to keep such scripts in sync across different dev or build machines if not stored in a central repository such as the Cauldron.

Since a recent PR, `publish-container` / `transform-container` commands are allowing for the `publisher`/`transformer` configuration to be passed as a local file path to a file containing the configuration. This is a nice improvement but doesn't solve the sharing aspect. Idea is therefore to also allow to pass a reference to a file stored in Cauldron, using a scheme that we will introduce (most probably `cauldron://`). This way config (complex or simple) or supporting files could be kept separate from the Cauldron document (thus not bloating it) while being shared across users of a Cauldron and allow easy maintenance.

This PR contains some tests, but no documentation for new commands yet. A specific PR for documentation regarding Cauldron file storage will be done prior to 0.22 release, including transformers documentation as well as updated publishers documentation, as they will be closely related.

### DIFF
--- a/ern-cauldron-api/src/CauldronApi.ts
+++ b/ern-cauldron-api/src/CauldronApi.ts
@@ -804,6 +804,70 @@ export default class CauldronApi {
   // =====================================================================================
 
   // -------------------------------------------------------------------------------------
+  // ARBITRARY FILE ACCESS
+  // -------------------------------------------------------------------------------------
+
+  public async addFile({
+    cauldronFilePath,
+    fileContent,
+    fileMode,
+  }: {
+    cauldronFilePath: string
+    fileContent: string | Buffer
+    fileMode?: string
+  }) {
+    if (!cauldronFilePath) {
+      throw new Error('[addFile] cauldronFilePath is required')
+    }
+    if (!fileContent) {
+      throw new Error('[addFile] fileContent is required')
+    }
+    if (await this.hasFile({ cauldronFilePath })) {
+      throw new Error(
+        `[addFile] ${cauldronFilePath} already exists. Use updateFile instead.`
+      )
+    }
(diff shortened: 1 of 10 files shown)


ern-cauldron-api/src/CauldronApi.ts
```diff
@@ -803,6 +803,70 @@ export default class CauldronApi {
...
+    fileContent: string | Buffer
+    fileMode?: string
+  }) {
+    if (!cauldronFilePath) {
+      throw new Error('[updateFile] cauldronFilePath is required')
+    }
+    if (!fileContent) {
+      throw new Error('[updateFile] fileContent is required')
+    }
+    if (!(await this.hasFile({ cauldronFilePath }))) {
```
krunalsshah: :minor: extra ()

belemaire: That's the way it is across the code base. Not intended but no real choice, this is due to `prettier` auto formatting of the source code. Someone reported this in https://github.com/prettier/prettier/issues/746 but in the end I think decision was to keep them.

PR #859, opened 2018-07-17
add system test for supported platform config

krunalsshah:
add system test for supported platform config


belemaire:
Thanks !

One gotcha here is for developers running system tests locally, as it will update their Electrode Native local configuration during the tests and not roll it back to previous config once done (for ex, in my local Electrode Native configuration, my tmp-dir is set to `/Users/blemair/ern-tmp` and this test will set it to `~/dir/to/command/exec`). We had the same issue with Cauldron (it was switching to a Cauldron for tests but not switching back to user activated Cauldron after tests .. this was fixed recently).

What I would suggest is to just do a copy of the `.ernrc` file, next to its location on disk (for ex `.ernrc.orig`) just before running the `ern platform config` commands, and then after the test commands are executed, just restore the original `.ernrc.orig` to `.ernrc`.

I would recommend surrounding whole of the `ern platform config` commands in a `try` block and restore the `.ernrc` in the `finally` block so that even in case of crash of one of these commands, the original `.ernrc` is properly restored.

Also the path to the `.ernrc` file might not be the default `~/.ern/.ernrc` one, but might be a different one if env variable `ERN_HOME` is set. Path to `.ernrc` can be obtained as follow :

```
const ERN_ROOT_PATH = process.env.ERN_HOME || path.join(os.homedir(), '.ern')
const ERN_RC_FILE_PATH = path.join(ERN_ROOT_PATH, '.ernrc')
```

krunalsshah:
yes good point. For some reason I thought it uses a tmp folder. My bad, should have checked.

krunalsshah:
@belemaire yes it makes sense. Initially thought process was to use `fs.copyFile` & changed my mind to handle it by keeping a copy in the object :) Thanks for spotting this.

PR #868, opened 2018-07-18
Add ECDevice scheme to iOS Container

belemaire:
This scheme is needed for the Fat Binary script which will be moved to a transformer soon.
It is duplicated from the ElectrodeContainer scheme with the difference that the default run configuration is release (instead of debug).
iOS system tests fixture donot need to be regenerated (ran iOS Container system tests locally without any issue).


krunalsshah:
@ghostatron @gmbharath12

PR #869, opened 2018-07-19
Expose API swift files as public headers in ElectrodeContainer

belemaire:
This is needed for Fat Binary proper support and is also more handy for some other use cases.
This feature requires `xcode-ern` version `1.0.8`, so the version of this dependency has been adequately bumped.


krunalsshah:
@ghostatron @gmbharath12

PR #883, opened 2018-08-05
Update iOS Container for third party Frameworks support

belemaire:


belemaire:
Auto approving as this needs to be included in 0.23 release.

PR #884, opened 2018-08-05
Add time taken by tasks for debug and trace log levels

belemaire:
Time taken by tasks was only logged for non `debug`/`trace` log levels (the ones using spinners).
This PR updates `kax` which now allows to log time taken by tasks when using `KaxSimpleRenderer` (the renderer used for `debug`/trace` levels.


belemaire:
Auto approving as this needs to be included in 0.23 release.

PR #891, opened 2018-08-17
Introduce Electrode Native version enforcement by Cauldron

belemaire:
Add a way to enforce a specific Electrode Native version to be used by all clients of a Cauldron.

Enforcement will be active by default on newly created Cauldrons. Existing Cauldrons will have to explicitly opt-in for this feature by using the new `ern cauldron update ernversion` command, to set an initial version to be enforced.

A new command is introduced, part of this feature : `ern cauldron update ernversion`. This command allows to update the version of Electrode Native that is enforced by the Cauldron, thus allowing a Cauldron owner to control the EN version of all clients connecting to the Cauldron.

If Cauldron Electrode Native version enforcement is active, any user running a different version of Electrode Native will be denied any operation involving the Cauldron. They will be prompted to either update their Electrode Native version or to update the Electrode Native version enforced by the Cauldron.

To be noted that version `1000.0.0` of Electrode Native (development version) is bypassing enforcement (meaning that it can access any Cauldron, whatever Electrode Native version enforced by the target Cauldron).

Also to be noted that this enforcement will only work for Electrode Native versions >= 0.24.0, as this feature will be released in this upcoming version. For example, if the Cauldron is enforcing version 0.24.0 but the client is using 0.21.0, the enforcement will be bypassed given that 0.21.0 does not contain the enforcement code.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -106,6 +106,7 @@
   - [jsapiimpls](cli/cauldron/update/jsapiimpls.md)
   - [miniapps](cli/cauldron/update/miniapps.md)
   - [nativeapp](cli/cauldron/update/nativeapp.md)
+  - [ernversion](cli/cauldron/update/ernversion.md)
 - [cauldron batch](cli/cauldron/batch.md)
 - [cauldron regen-container](cli/cauldron/regen-container.md)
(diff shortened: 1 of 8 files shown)


belemaire:
Closing this one, will reopen later on with a different design approach to hit two birds in one stone

PR #899, opened 2018-08-22
Update iOS API Impl Generator for Swift 4 compatibility

belemaire:
For Swift 4 compatibility, all functions that needs to be invoked from objc, needs to be marked with 
mark with `@objc`, or it will not generate class function in Swift bridge header.

The `register` func in the generated `Controller` class does not have the `@objc` so it leads to an issue when trying to invoke it (with XCode 9.3) : 
screen shot 2018-08-22 at 1 25 11 pm
screen shot 2018-08-22 at 1 24 57 pm

Fix is just to add the `@objc` to the func.

screen shot 2018-08-22 at 1 27 47 pm

Just updated the template accordingly. Also compatible with Swift < 4.

### DIFF
--- a/ern-api-impl-gen/resources/ios/apiController.mustache
+++ b/ern-api-impl-gen/resources/ios/apiController.mustache
@@ -26,6 +26,7 @@ import Foundation
     
     private var requestHandler: {{apiName}}ApiRequestHandlerProvider?
     
+    @objc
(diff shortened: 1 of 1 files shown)


gmbharath12:
yes that's right 👍 Thanks for the fix!

Issue #896, opened 2018-08-21
iOS Container Build Issue


belemaire:
A few users reported an issue that leads to failure to build an iOS Container (One of the errors being reported in the build error log, amongst others is `Could not build module Darwin`)

We haven't yet precisely identified the root cause of this issue, even though it seems related to the scripts being run by React Native during XCode build, to retrieve third party dependencies.

A few workarounds have been identified, though they work for some users but not for others.
We don't have necessary resources at the moment to dig deep into this issue, any help would be appreciated, until we can allocate proper resources.

All these issues are related to this root issue, and contain more information :

https://github.com/electrode-io/electrode-native/issues/575
https://github.com/electrode-io/electrode-native/issues/881
https://github.com/electrode-io/electrode-native/issues/226

belemaire:
For users experiencing this issue, the root cause seems to be related to corrupted cached dependencies retrieved by React Native framework during the iOS Container build.

The following steps fixed this issue for our team :

Before running an iOS Container build (when building the Container on its own, or through a native application build) :

1.  Remove the `third-party` directory located in`ElectrodeContainer/Libraries/ReactNative/third-party`
2. Remove the React Native cache directory `rm -rf ~/.rncache`

Ideally these two steps should be done before each build.

If you still experience this issue after performing these steps, please report it to us.

gmbharath12:
For Xcode 10/10.1 please follow these steps. 

https://github.com/electrode-io/electrode-native/wiki/Prerequisites-Xcode

belemaire:
It's been a while now that we had any report of this issue. Closing for now.

PR #903, opened 2018-08-24
Use non recursive headers search paths for iOS Container

belemaire:
It appears that recursive headers search paths in iOS Container are leading to some issues under certain non precisely identified circumstances. 

For example, it can lead to the know issue https://github.com/electrode-io/electrode-native/issues/896

This PR turn all headers search paths defined in the iOS Container as non recursive.


gmbharath12:
👍

PR #910, opened 2018-08-29
Improve setup-dev script

belemaire:
There is no need to run `npm run rebuild` (which transpile every ern module) anymore for running Electrode Native development version (1000.0.0).

Building (transpiling) is now only needed to perform Electrode Native releases.

Removing this extra unecessary step from `setup-dev` as it was taking unnecessary time to run.


belemaire:
Merging this one now to speed up cron ci.

PR #919, opened 2018-09-05
add method to remove a mini app view from electrode activity delegate…

deepueg:
…. This helps activities with multiple fragments to easily destroy views when a fragment is destroyed.


belemaire:
@deepueg : Raaah once more forgot about system test fixture of Android Container :( Should be part of the PR as Container code was merged. Will do it separately and issue a new PR.

PR #923, opened 2018-09-14
fix toBundle logic in apiGen.

deepueg:


krunalsshah:
pls regen the system test fixture

deepueg:
regenerated.

PR #938, opened 2018-10-11
Miscellaneous code improvements

belemaire:
Mostly in `ern-core` and `ern-local-cli`, targeting the use of `yargs` coercion through `coerce` to simplify cli command handlers. Includes other smallish beautifications / type enforcements here and there.


ern-core/src/ErnBinaryStore.ts
```diff
@@ -27,6 +27,11 @@ export default class ErnBinaryStore implements BinaryStore {
   public async removeBinary(
     descriptor: NativeApplicationDescriptor
   ): Promise<string | Buffer> {
+    if (!(await this.hasBinary(descriptor))) {
```
krunalsshah: can be extracted to a new function.

krunalsshah:
37e7c24
f503d5a
37e7c24
is there an opportunity to squash these commits?

PR #975, opened 2018-11-12
Add detachContainerVersionFromRoot flag to Cauldron nap version

belemaire:
Adds support of a new flag `detachContainerVersionFromRoot` to the `CauldronNativeAppVersion` type.

If this flag is present and set (`"detachContainerVersionFromRoot" : true`) for a given Native Application Version in Cauldron, then, Container versioning will be 'detached' (independent) from top level Container versioning. What this means in practice is that whenever the Container version of the native application is updated, it won't update the top level Container version.

For example, assuming we have a Cauldron with current top level Container version set to `1.0.5` and 3 versions of a native application defined, with only one having the `detachContainerVersionFromRoot ` flag set, as follow :

| Descriptor  | Container Version | detachContainerVersionFromRoot |
| ------------- | ------------- | ------------- |
| test:android:1.0.0  | 1.0.4  | false  |
| test:android:2.0.0 |  1.0.5  | false  |
| test:android:1000.0.0 |  2.0.1  | true  |

If one regenerate a new Container for `test:android:1.0.0`, then the container version associated to `test:android:1.0.0` will be patch bumped to `1.0.5` and the top level Container version will be also set to `1.0.5`.

However, because `detachContainerVersionFromRoot ` flag is set for `test:android:1000.0.0` application, if one was to regenerate a new Container for `test:android:1000.0.0`, the Container version of this descriptor would be bumped to `2.0.2` but the top level container version will remain untouched (thus the container version associated to the native application version is detached from other native application versions).

This is not documented yet as we'll have come up with better flexibility for Container versioning in the future. That being said, this feature is needed for one internal team in relation the CI workflow and will be needed by other internal teams very soon, therefore the need to have a working -temporary- solution in place as of now.

### DIFF
--- a/ern-cauldron-api/src/CauldronHelper.ts
+++ b/ern-cauldron-api/src/CauldronHelper.ts
@@ -1133,19 +1133,25 @@ export class CauldronHelper {
     containerVersion: string
   ): Promise<void> {
     await this.cauldron.updateContainerVersion(napDescriptor, containerVersion)
-    const topLevelContainerVersion = await this.getTopLevelContainerVersion(
+    const nativeApplicationVersion: CauldronNativeAppVersion = await this.getDescriptor(
       napDescriptor
     )
-    if (
-      semver.valid(containerVersion) &&
-      topLevelContainerVersion &&
-      semver.valid(topLevelContainerVersion) &&
-      semver.gt(containerVersion, topLevelContainerVersion)
-    ) {
-      await this.cauldron.updateTopLevelContainerVersion(
-        napDescriptor,
-        containerVersion
+    // Update top level Container version only for non detached container versions
+    if (!nativeApplicationVersion.detachContainerVersionFromRoot) {
+      const topLevelContainerVersion = await this.getTopLevelContainerVersion(
+        napDescriptor
       )
+      if (
+        semver.valid(containerVersion) &&
+        topLevelContainerVersion &&
+        semver.valid(topLevelContainerVersion) &&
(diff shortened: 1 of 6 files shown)


krunalsshah:
@belemaire can we name the flag as `detachContainerVersionFromRoot `

belemaire:
@krunalsshah Sure. Just did it. Updated PR name and description accordingly as well as commit description.

krunalsshah:
@belemaire what happens if the container version collide, lets say `topLevel` container is at `1.1.1` and the detached container is requesting to be at `1.1.1`
Other situations when , 2 native descriptors are both trying to use the same `1.11.1` & `1.11.1` with detached true, wouldn't it fail to publish on the artifact?

belemaire:
@krunalsshah You are correct, it might collide. 
There is no version collision safeguards with detached container version, and cannot really be while keeping this feature dumb simple as it is desired now.
User has to properly manage its versioning ranges manually. 
But for now it's really meant to be used only for some of our teams, for the CI workflow, just for one specific native app version (the dev one) of the native app that we might create (maybe version 1000.0.0 in Cauldron for example). We will detach container version for this specific native app version and only use branches for MiniApps. Container version might be starting with 1000.0.0 for this native app version so a very different version range than root container versioning, with no collision risk. Will be used only for dev CI workflow where placeholder 1000.0.+ version will be used in Android build.gradle for ex.
This quick implementation was just meant to unlock this specific CI dev workflow, and keep a container versioning range for dev containers distinct from release containers (otherwise workflow wouldn't work).

PR #977, opened 2018-11-13
refactor definition of updateCodePushEntry & patch method

krunalsshah:
refactor definition of updateCodePushEntry to look consistent with addCodePushEntry, patch method to look consistent with promote and release


belemaire:
Thanks 👍 
Waiting for tests to clear and will merge.

PR #984, opened 2018-11-16
Fix cauldron repo use command for invalid alias

belemaire:
`cauldron repo use` command was incorrectly working when provided with an invalid alias (alias that does not exist). This PR fixes this issue. `cauldron repo use` command will now properly fail in case it is used with an invalid cauldron alias.

Thanks @gmbharath12 for noticing this issue.


gmbharath12:
👍

PR #993, opened 2018-11-21
Fix queued requests issue in Bridge

gmbharath12:
- When a request is made from iOS to Javascript, Electrode Bridge should wait for React Native initialization and queue up the requests. It was improperly handled in Bridge. Time `dispatch_time_t` was incorrectly set when dispatching semaphore that caused this issue and couldn't wait until `dispatch_semaphore_signal` was fired.

- Detaching these operations in global queue instead of the main queue. Because we don't want to block the main thread.

- After React Native is initialized we dequeue all requests (FIFO). 

 - Example showing iOS request ---> to JavaScript 
`[movieAPI.requests getMovieDetailWithMovieId:@"1" responseCompletionHandler:^(id  _Nullable data, id<ElectrodeFailureMessage>  _Nullable message) {
        NSLog(@"Data %@", data);
    }]` 

Credits to @belemaire 🥇 for the help 👍

### DIFF
--- a/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
+++ b/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
@@ -229,9 +229,16 @@ static dispatch_semaphore_t semaphore;
             SEL selector = NSSelectorFromString(@"onReactNativeInitialized");
             SEL transceiverReadySelector = NSSelectorFromString(@"onTransceiverModuleInitialized");
             if ([localModuleInstance  respondsToSelector:selector]) {
-                dispatch_async(dispatch_get_main_queue(), ^{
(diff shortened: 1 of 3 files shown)


krunalsshah:
@belemaire , look's we'd need the regen of system test fixtures before merging this.

belemaire:
@krunalsshah Good point indeed. 
The script should have warned @gmbharath12 upon commit to take a closer look to eventually regen the fixtures as it would have detected a modification in one of the container hulls.

@gmbharath12 Could you please update the code as discussed ? 
i.e : timeout of 60 seconds instead of 10, and if the timeout occurs, don't call the selector (`onReactNativeInitialized`) but rather log an error message such as (`Timeout: RCTJavaScriptDidLoadNotification was not raised`)

PR #997, opened 2018-11-29
Update yarn locks copy over mecanism

belemaire:
Two things in this PR to improve yarn locks copy over (when creating a new native application version in Cauldron and copying data over from a source native application version).

- Copy the source yarn.lock file instead of reusing the same file
Reusing the same yarn.lock was preventing proper archiving of yarn locks and could lead to some issues. 
For illustration, when copying over the Container yarn.lock of version 1.0 of an app to version 2.0 and reusing the same yarn.lock file -as of now- (copying the same yarn.lock filename over to the new version), as soon as a new Container gets regenerated for version 2.0, it will recreate a new yarn.lock and delete the previous one ... impact is that because the previous one was also the one used by the Container of the native app 1.0, this yarn lock file would not be available anymore for new Container generation of native app 1.0. 

- Update yarn.lock filename generation
By using uuids generated on the fly, instead of using the SHA of the yarn.lock.
This is in direct relation with previous bullet point. If using SHA as the filename, there is no way to have two copies of the same file (yarn.lock) with different filenames, as same content produces same SHA. This is why we're switching to a uuid base filename, to guarantee distinct filenames even for same file contents.


ern-cauldron-api/src/CauldronHelper.ts
```diff
@@ -408,9 +410,9 @@ export class CauldronHelper {
     napDescriptor: NativeApplicationDescriptor,
     key: string,
     yarnlockPath: string
-  ): Promise<void> {
+  ): Promise<string> {
     const yarnLockFile = await fileUtils.readFile(yarnlockPath)
-    return this.cauldron.addYarnLock(napDescriptor, key, yarnLockFile)
+    return await this.cauldron.addYarnLock(napDescriptor, key, yarnLockFile)
```
krunalsshah: pls remove await.

krunalsshah:
lgtm

ern-cauldron-api/src/CauldronHelper.ts
belemaire: done 👍

PR #1001, opened 2018-11-29
Add json option flag to some commands

belemaire:
Add `--json` option flag to the following commands :

- `cauldron config get`
- `cauldron get dependency`
- `cauldron get nativeapp`
- `list dependencies`

There is some special handling in the top level CLI implementation to disable all other console output in case this flag is set (no banner / no header and no logs what so ever), so that only the resulting formatted JSON output is written to the console. 

This is done because common use of the `--json` flag in CLI commands is to feed the result of the command to some other program. Having some banner / logs before the formatted JSON makes it more difficult to be parsed for consuming app.

### DIFF
--- a/ern-local-cli/src/commands/cauldron/config/get.ts
+++ b/ern-local-cli/src/commands/cauldron/config/get.ts
@@ -14,6 +14,10 @@ export const builder = (argv: Argv) => {
       type: 'string',
     })
     .coerce('descriptor', NativeApplicationDescriptor.fromString)
+    .option('json', {
+      describe: 'Output config as a single line JSON record',
+      type: 'boolean',
+    })
     .option('key', {
       describe:
         'The config key (echoes the whole config object if not specified)',
@@ -29,10 +33,12 @@ export const builder = (argv: Argv) => {
 
 export const commandHandler = async ({
   descriptor,
+  json,
   key,
   strict,
 }: {
   descriptor?: NativeApplicationDescriptor
+  json?: boolean
   key?: string
   strict: boolean
 }) => {
@@ -49,8 +55,10 @@ export const commandHandler = async ({
   } else if (!key && !strict) {
     result = await cauldron.getConfig(descriptor)
   }
-  const jsonConfig = JSON.stringify(result, null, 2)
-  console.log(jsonConfig)
+
+  process.stdout.write(
+    json ? JSON.stringify(result) : JSON.stringify(result, null, 2)
+  )
 }
 
 export const handler = tryCatchWrap(commandHandler)
--- a/ern-local-cli/src/commands/cauldron/get/dependency.ts
+++ b/ern-local-cli/src/commands/cauldron/get/dependency.ts
@@ -16,13 +16,19 @@ export const builder = (argv: Argv) => {
     .coerce('descriptor', d =>
       NativeApplicationDescriptor.fromString(d, { throwIfNotComplete: true })
     )
+    .option('json', {
+      describe: 'Output dependencies as a single line JSON array',
+      type: 'boolean',
+    })
     .epilog(epilog(exports))
 }
 
 export const commandHandler = async ({
   descriptor,
+  json,
 }: {
   descriptor: NativeApplicationDescriptor
+  json?: boolean
 }) => {
   await logErrorAndExitIfNotSatisfied({
     napDescriptorExistInCauldron: {
@@ -34,9 +40,9 @@ export const commandHandler = async ({
 
   const cauldron = await getActiveCauldron()
   const dependencies = await cauldron.getNativeDependencies(descriptor)
-  for (const dependency of dependencies) {
-    log.info(dependency.toString())
-  }
+  json
+    ? process.stdout.write(JSON.stringify(dependencies))
+    : dependencies.forEach(d => log.info(d.toString()))
 }
 
 export const handler = tryCatchWrap(commandHandler)
--- a/ern-local-cli/src/commands/cauldron/get/nativeapp.ts
+++ b/ern-local-cli/src/commands/cauldron/get/nativeapp.ts
@@ -9,21 +9,31 @@ export const desc = 'Get a native application from the cauldron'
 export const builder = (argv: Argv) => {
   return argv
     .coerce('descriptor', NativeApplicationDescriptor.fromString)
+    .option('json', {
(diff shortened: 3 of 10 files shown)


ern-local-cli/src/commands/cauldron/config/get.ts
```diff
@@ -14,6 +14,14 @@ export const builder = (argv: Argv) => {
       type: 'string',
     })
     .coerce('descriptor', NativeApplicationDescriptor.fromString)
+    .option('json', {
+      describe: 'json ouput',
+      type: 'boolean',
+    })
+    .option('json', {
```
krunalsshah: duplicate

ern-local-cli/src/commands/cauldron/get/dependency.ts
```diff
@@ -16,13 +16,19 @@ export const builder = (argv: Argv) => {
     .coerce('descriptor', d =>
       NativeApplicationDescriptor.fromString(d, { throwIfNotComplete: true })
     )
+    .option('json', {
+      describe: 'json output',
```
krunalsshah: format output as lines of JSON (see jsonlines.org), please change it all description placeholders. thanks

ern-local-cli/src/commands/cauldron/get/dependency.ts
```diff
@@ -34,9 +40,9 @@ export const commandHandler = async ({
 
   const cauldron = await getActiveCauldron()
   const dependencies = await cauldron.getNativeDependencies(descriptor)
-  for (const dependency of dependencies) {
-    log.info(dependency.toString())
-  }
+  json
+    ? console.log(JSON.stringify(dependencies, null, 2))
```
krunalsshah: console.log?

ern-local-cli/src/commands/cauldron/get/nativeapp.ts
```diff
@@ -9,18 +9,26 @@ export const desc = 'Get a native application from the cauldron'
...
 }: {
   descriptor?: NativeApplicationDescriptor
+  json?: boolean
 }) => {
   const cauldron = await getActiveCauldron()
   if (!descriptor) {
     const napDescriptors = await cauldron.getNapDescriptorStrings()
-    napDescriptors.forEach(n => log.info(n))
+    json
+      ? console.log(JSON.stringify(napDescriptors, null, 1))
```
krunalsshah: console.log?

ern-local-cli/src/index.ts
```diff
@@ -115,12 +139,16 @@ export default function run() {
   kax.renderer =
     logLevel === 'trace' || logLevel === 'debug'
       ? new KaxSimpleRenderer(kaxRendererConfig)
-      : new KaxAdvancedRenderer(kaxRendererConfig)
+      : logLevel === 'off'
```
krunalsshah: do we really want to complicate here? let's keep default render as the option. as documented in the packages wiki

ern-local-cli/src/commands/list/dependencies.ts
```diff
@@ -34,17 +45,21 @@ export const commandHandler = async ({ module }: { module?: string }) => {
...
-  logDependencies(
-    dependencies.thirdPartyInManifest,
-    'Third party declared in Manifest'
-  )
-  logDependencies(
-    dependencies.thirdPartyNotInManifest,
-    'Third party not declared in Manifest'
-  )
+  if (json) {
+    console.log(JSON.stringify(dependencies, null, 2))
```
krunalsshah: console.log?

ern-local-cli/src/commands/cauldron/config/get.ts
belemaire: woops. will remove

ern-local-cli/src/commands/cauldron/get/dependency.ts
belemaire: why not ? ;)

ern-local-cli/src/commands/cauldron/get/nativeapp.ts
belemaire: -duplicate-

ern-local-cli/src/commands/list/dependencies.ts
belemaire: -duplicate-

ern-local-cli/src/index.ts
belemaire: Not sure that I get what you mean here ?
Basically if the `--json` option is set, we want to disable all kind of logs (thus setting logLevel to 'off' level). 
This includes any log done through `kax`. I don't see another simpler way (or maybe avoid the ternary and use if/else blocks instead which would make the code more readable ?)

ern-local-cli/src/commands/cauldron/get/dependency.ts
belemaire: We don't output multiple lines of json here (i.e not one json "record" per line as described in jsonlines.org), but a single json object/array (a single "record"). that being said, that brings a good point, its somewhat easier for third party programs to parse if outputing json object as single line instead of formatted multiline json object. will update the code accordingly.

belemaire: Will also update descriptions to be more meaningful

belemaire:
Updated PR per review comments.

ern-local-cli/src/commands/cauldron/get/dependency.ts
krunalsshah: I remember we had PR long back to get rid of `console.log`, fail to gather context though.

belemaire: Yes that's correct, `console.log` should not be used indeed, but rather use `log.` functions. 
That being said, as noted, this is a special case due to the fact that `log.` is turned off when `--json` flag is on.
But I can see the confusion, and that's fair, so I'll use `process.stdout.write` instead, so that we can keep the rule of no `console.log` in the code base.

belemaire: done

ern-local-cli/src/commands/cauldron/get/dependency.ts
belemaire: done

PR #1016, opened 2018-12-06
Improve same MiniApps inference

belemaire:
Improve inference of same MiniApps versions used in `performContainerStateUpdateInCauldron` function. An optimized path is followed in case all MiniApps are using the same versions compared to previous Container, but this similar version inference was not considering git branches/tags.

This PR adds a couple utility functions to `ern-core`, especially `areSamePackagePathsVersions` function, which given two PackagePath arrays will return `true` if the arrays contain identical MiniApps with identical versions, and properly consider MiniApps that are using git branch/tags.
It replaces `performContainerStateUpdateInCauldron` basic algorithm used for same MiniApps versions inference with a call to `areSamePackagePathsVersions` to extend optimization reach.


ern-core/test/Utils-test.ts
```diff
@@ -302,4 +302,100 @@ d9fa903349bbb9e7f86535cb69256e064d0fba65        refs/tags/v0.1.2`
...
+      ]
+      const b = [
+        PackagePath.fromString('packA@1.2.3'),
+        PackagePath.fromString('packB@1.0.0'),
+      ]
+      const result = await coreUtils.areSamePackagePathsAndVersions(a, b)
+      expect(result).true
+    })
+
+    it('should return true if the packages arrays are identical', async () => {
```
krunalsshah: duplicate oops

PR #1021, opened 2018-12-10
Use uuid for yarn lock instead of file shasum

belemaire:
Use uuid instead of file shasum as yarn lock filename (missing from #997).


ern-cauldron-api/src/CauldronApi.ts
```diff
@@ -1173,7 +1173,7 @@ export default class CauldronApi {
     if (fileName) {
       const pathToOldYarnLock = this.getRelativePathToYarnLock(fileName)
       await this.fileStore.removeFile(pathToOldYarnLock)
-      const newYarnLockFileName = shasum(yarnlock)
+      const newYarnLockFileName = uuidv4()
```
krunalsshah: if we are not using `shasum` we can remove the code from the `uitls.js`

belemaire: yes, good call, was the only use left. 
let me remove it.

PR #1030, opened 2018-12-20
allow adding file dependencies for react native plugins that has them.

deepueg:


krunalsshah:
👍 what is this PR about? Also does it need any documentation to illustrate the usage?

PR #1038, opened 2019-01-03
Enable bridge logger for react native dev mode

krunalsshah:
Enable bridge logger for react native dev mode


deepueg:
Very nice.! 👏

Issue #770, opened 2018-05-23
MiniApp Typescript Support?


gmusick:
I'm trying to use typescript to create a MiniApp and am getting the following when running `ern run-android`:

```
✔ Starting React Native packager
⠧ Generating Container[bundleMiniApps] Something went wrong: Error: Command failed: /Users/greg/.ern/versions/0.18.2/node_modules/ern-core/node_modules/.bin/react-native bundle --entry-file=index.android.js --dev=false --platform=android --bundle-output=/Users/greg/.ern/containergen/out/android/lib/src/main/assets/index.android.bundle --assets-dest=/Users/greg/.ern/containergen/out/android/lib/src/main/res

Unable to resolve module `./App` from `/private/var/folders/1c/m9p72__929d5krj03_p9jkk40000gn/T/tmp-41925fpDeN0iUAuL7/node_modules/native-test-miniapp/index.android.js`: The module `./App` could not be found from `/private/var/folders/1c/m9p72__929d5krj03_p9jkk40000gn/T/tmp-41925fpDeN0iUAuL7/node_modules/native-test-miniapp/index.android.js`. Indeed, none of these files exist:

  * `/private/var/folders/1c/m9p72__929d5krj03_p9jkk40000gn/T/tmp-41925fpDeN0iUAuL7/node_modules/native-test-miniapp/App(.native||.android.js|.native.js|.js|.android.json|.native.json|.json)`
  * `/private/var/folders/1c/m9p72__929d5krj03_p9jkk40000gn/T/tmp-41925fpDeN0iUAuL7/node_modules/native-test-miniapp/App/index(.native||.android.js|.native.js|.js|.android.json|.native.json|.json)`


[generateContainer] Something went wrong. Aborting Container Generation
✖ Command failed: /Users/greg/.ern/versions/0.18.2/node_modules/ern-core/node_modules/.bin/react-native bundle --entry-file=index.android.js --dev=false --platform=android --bundle-output=/Users/greg/.ern/containergen/out/android/lib/src/main/assets/index.android.bundle --assets-dest=/Users/greg/.ern/containergen/out/android/lib/src/main/res

[...]

PR #1059, opened 2019-01-28
Typescript file support

davidgovea:
Recent versions of react-native support typescript source files (`.ts`, `.tsx`) out of the box.

On recent versions of ern (I started with `v0.26`), typescript files worked fine, but they do not on the most recent version.

[This commit](https://github.com/electrode-io/electrode-native/commit/9c30e71190994e74b445fe6004e0049cb0bc609c) added overrides to the RN config, and caused typescript files to not be resolved.

So, this commit just adds ts/tsx files back into the resolved extensions.

Looks like 'backlog' ticket #770 could be closed out due to the work inside RN.

---

Thanks for this fantastic project -- it's incredible just how much surface area is covered by ERN.
Let me know if you have any comments -- happy to refactor/document/test as needed to get this change incorporated :smiley:

### DIFF
--- a/ern-container-gen/src/generateMiniAppsComposite.ts
+++ b/ern-container-gen/src/generateMiniAppsComposite.ts
@@ -240,10 +240,10 @@ export async function generateMiniAppsComposite(
     let sourceExts
     if (semver.gte(compositeReactNativeVersion, '0.57.0')) {
       sourceExts =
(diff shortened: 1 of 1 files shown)


deepueg:
@davidgovea Thanks for the PR and feedback, makes us happy to see it's helping developers. PR looks good to me.

PR #1070, opened 2019-02-27
Add initial Babel support

belemaire:
This PR adds initial Babel support for MiniApps.

Prior to this PR, any `.babelrc` file present in the root of a MiniApp was not considered during bundling. This was problematic for MiniApps that wanted to make use of some Babel plugins.

Unfortunately, proper support for Babel 7 requires a bit of hacking, because when we generate 
a `composite` bundle of multiple MiniApps, we can consider the composite as a 'virtual monorepo' composed of different MiniApps, therefore the .babelrc is not at the top level / root of the bundling working directory. With Babel 7 (RN >= 0.56.0) we need to make use of the [babelrcRoots](https://babeljs.io/docs/en/options#babelrcroots) option of Babel to specify 'whitelisted' directories containing `.babelrc` files to be processed (MiniApps typically), otherwise they are ignored. 
The problem is that as of now, the [metro](https://github.com/facebook/metro) bundler does not offer a way to provide this option. 
We will open a pull request to metro soon, but meanwhile, and to maintain backward compatibility until metro offer this option, we rely on patching a metro source file prior to bundling, to add this option.

With this PR, any MiniApp will be able to make use of Babel plugins by using a valid `.babelrc` file at the root of the MiniApp, and setting the `useBabelRc` flag in the `ern` section of their `package.json`, as follow 

```json
"ern": {
  "useBabelRc": true
}
```

 to let Electrode Native know that this MiniApp `.babelrc` should be processed during bundling.

### DIFF
--- a/ern-container-gen/src/generateMiniAppsComposite.ts
+++ b/ern-container-gen/src/generateMiniAppsComposite.ts
@@ -5,6 +5,7 @@ import {
   yarn,
   readPackageJson,
   writePackageJson,
+  fileUtils,
 } from 'ern-core'
 import { cleanupMiniAppsCompositeDir } from './cleanupMiniAppsCompositeDir'
 import {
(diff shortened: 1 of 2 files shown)


ern-container-gen/src/generateMiniAppsComposite.ts
```diff
@@ -122,15 +123,142 @@ export async function generateMiniAppsComposite(
...
+
+    const pathToMetroNodeModuleDir = path.join(
+      compositeNodeModulesPath,
+      'metro'
+    )
+    let metroPackageJson
+    if (fs.existsSync(pathToMetroNodeModuleDir)) {
+      metroPackageJson = await readPackageJson(pathToMetroNodeModuleDir)
+    }
+    const metroVersion = metroPackageJson ? metroPackageJson.version : '0.0.0'
```
belemaire: check of existence of metro directory and falling back to 0.0.0 version otherwise is only done to support the tests. that's pretty ugly but anyway this whole function should be refactored asap to break it down into smaller units along with the tests.

PR #1071, opened 2019-02-28
Add a new Cauldron config flag to bypass yarn.lock

belemaire:
Add a new flag `bypassYarnLock` for both `containerGenerator` and `codePush` config objects in Cauldron.

Setting this flag to `true` will bypass `yarn.lock` usage for JS composite generation. 
It will still publish the composite `yarn.lock` to the Cauldron (for tracking purposes and to have a way to know all of the JS packages and versions that are included in any bundle), but won't make use of it to actually generate the composite, recreating a new one upon every generation.

If this flag is not set, it will default to `false` (current behavior).

### DIFF
--- a/ern-orchestrator/src/codepush.ts
+++ b/ern-orchestrator/src/codepush.ts
@@ -333,6 +333,14 @@ export async function performCodePushOtaUpdate(
       PackagePath.fromString(j.toString())
     )
 
+    const codePushConfig = await cauldron.getCodePushConfig(napDescriptor)
+    if (codePushConfig && codePushConfig.bypassYarnLock) {
+      pathToYarnLock = undefined
+      log.debug(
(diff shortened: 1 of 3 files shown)


ern-orchestrator/src/codepush.ts
```diff
@@ -333,6 +333,14 @@ export async function performCodePushOtaUpdate(
       PackagePath.fromString(j.toString())
     )
 
+    const codePushConfig = await cauldron.getCodePushConfig(napDescriptor)
+    if (codePushConfig && codePushConfig.bypassYarnLock) {
+      pathToYarnLock = undefined
+      log.debug(
```
krunalsshah: repetitive , message from ern-local-cli/src/commands/create-container.ts will take care of it

PR #1075, opened 2019-03-07
Expose a new method for getting the view of a Mini App

regnerjr:
Users often want to use only the RCTView from a mini app.
We should allow them to do so.

### DIFF
--- a/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
+++ b/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
@@ -89,13 +89,22 @@ NS_ASSUME_NONNULL_BEGIN
 
 /**
  Returns a react native miniapp (from a JSBundle) inside a view controller.
- 
- @param name The name of the mini app, preferably the same name as the jsbundle
- without the extension.
+
+ @param name The name of the mini app, preferably the same name as the jsbundle without the extension.
  @param properties Any configuration to set up the mini app with.
  @return A UIViewController containing the view of the miniapp.
  */
 - (UIViewController *)miniAppWithName:(NSString *)name
                            properties:(NSDictionary * _Nullable)properties;
+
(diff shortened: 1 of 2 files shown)


ern-api-impl-gen/hull/ios/ElectrodeApiImpl/ElectrodeReactNative.h
```diff
@@ -100,4 +100,15 @@ extern NSString * const ERNDebugEnabledConfig;
...
+   Returns a react native miniapp (from a JSBundle) inside a view.
+
+   @param name The name of the mini app, preferably the same name as the jsbundle
+   without the extension.
+   @param properties Any configuration to set up the mini app with.
+   @return A UIView containing the view of the miniapp.
+*/
+- (UIView *)miniAppViewWithName:(NSString *)name
+                     properties:(NSDictionary *)properties;
 @end
```
gmbharath12: Thanks John for the PR. 
These changes should be in `ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h`. 
Could you please remove these changes in `ern-api-impl-gen/hull/ios/ElectrodeApiImpl/ElectrodeReactNative.h`

gmbharath12: @regnerjr Did you test these changes on any app?

regnerjr: I did not get a chance to try these changes on any app. 
On Walmart Grocery, they are calling the existing API to get the View controller, and then grabbing the view out of that returned View Controller and just embedding the view directly. Which made me think maybe there was room for a dedicated API. 

***************
Thanks for the recommendation. I'll get these changes moved over to the correct files.

deepueg:
Closing since there is no further update on this for a while.

PR #1076, opened 2019-03-07
Ios container update

deepueg:
This adds two new convenience method for iOS container. 

```
/**
 Returns a react native miniapp (from a JSBundle) inside a view controller.

 @param name The name of the mini app, preferably the same name as the jsbundle
 without the extension.
 @param properties Any configuration to set up the mini app with.
 @return A UIViewController containing the view of the miniapp.
 */
                           properties:(NSDictionary * _Nullable)properties;

/**
 Call this to update an RCTRootView with new props. Calling this with new props will cause the view to be rerendered.
 Request will be ignored if the returned view is not an RCTRootView instance.
 */
- (void) updateView: (UIView *) view withProps:(NSDictionary *) newProps;
```

### DIFF
--- a/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
+++ b/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
@@ -97,5 +97,23 @@ NS_ASSUME_NONNULL_BEGIN
  */
 - (UIViewController *)miniAppWithName:(NSString *)name
                            properties:(NSDictionary * _Nullable)properties;
+
+/**
+ Returns a react native miniapp (from a JSBundle) inside a view controller.
+
+ @param name The name of the mini app, preferably the same name as the jsbundle
+ without the extension.
+ @param properties Any configuration to set up the mini app with.
+ @return A UIViewController containing the view of the miniapp.
+ */
+- (UIView *)miniAppViewWithName:(NSString *)name
+                     properties:(NSDictionary *_Nullable)properties;
+
+/**
+ Call this to update an RCTRootView with new props. Calling this with new props will cause the view to be rerendered.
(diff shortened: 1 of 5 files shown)


deepueg:
Just noticed that #1075 is also trying to do this. This PR has another method that allows to rerender a miniApp view. Perhaps we can close #1075 and proceed with this PR.

belemaire:
@gmbharath12 
It's been a while since this PR was initially opened. 
Could you please retry your testing to see if the crash still happens and sync with @deepueg if that's the case. Otherwise please give a :+1: to proceed with the merge. Only thing that keeps me from merging is your last comment referring to a crash.

gmbharath12:
@belemaire Yeah I'll test it by tomorrow and I'll let you know

gmbharath12:
@belemaire: Tested this one and it still breaks. Problem with this approach is that, UIView (say testView) that is returned by the function don't have any constraints set to it. So, it's the responsibility of the caller to make sure view's constraint is set properly. Also to be noted that, crash happens when we try adding this `testView` as a subview. Eg: `[self.view addSubview:testView]` With this as an example: we need to make sure both `self.view` and `testView` have the right constraints set. 

There's another way to add `testView` to parentView (self.view in our case): `self.view = testView;` We don't need to set the constraints of testView but we must set the right constraints for the parent view (which is self.view in this case). 

In brief, caller calling this function:
 - (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *)properties is responsible to have the right constraints set. 

I have made the changes on my end, tested and it looks OK to me. 

I'll approve this PR 👍

deepueg:
sorry for the delay, I will have the system tests regenerated tomorrow.

deepueg:
@belemaire merge conflict resolved.

PR #1084, opened 2019-03-12
Allow to selectively disable transformers/publishers

belemaire:
Add support of a new `disabled` boolean property in `transformers`/`publishers` objects configured in Cauldron.

Setting this property to `true` will disable (bypass) the transformer or publisher, as illustrated below

```json
{
  "name":"git",
  "url":"git@github.com:user/myweatherapp-android-container.git",
  "disabled": true
}
```

This allows for selective disablement of publishers and/or transformers in Cauldron, without having to remove the whole publisher/transformer configuration from Cauldron.

If the property is not present (default) or set to false, the transformer/publisher will be enabled.


gmbharath12:
awesome. Thanks 👍

Issue #1082, opened 2019-03-11
Invalid version: null when adding a new version of a nativeapp


LuisRizo:
I already have three versions of the miniapp on the cauldron, but they are all using only major and minor versions. For example, 1.00 instead of 1.00.02.

I'm running `ern cauldron add nativeapp NativeAppName:ios:1.00.02` and the command results in the following output:

```
✔ Ensuring that native application descriptor does not already exist in Cauldron 1s
✖ An error occurred: Invalid Version: null
```

I also tried using the --copyFromVersion flag and pointing to the last version and it now it says "Invalid Version: 0.50" which was the first native version I ever uploaded to the cauldron.

```
✔ Ensuring that native application descriptor does not already exist in Cauldron 1s
✖ An error occurred: Invalid Version: 0.50
```

I'm using ern platform version `v0.29.6`

belemaire:
Hi @LuisRizo 

The versions you are using are not valid semver compliant versions.
For MiniApps or Native Application Versions in Cauldron, we except only semver compliant versions (non compliant versions will fail publication to npm anyway, as npm registry is also excepting such versions).

Semantic versioning specifications are widely known and used by developers, you can see them here
https://semver.org/

You version `1.00.02` is not valid, because a version [cannot include leading zeros](https://semver.org/#spec-item-2), also you version `0.50` is not valid because a [valid version must include a patch version](https://semver.org/#spec-item-2)

LuisRizo:
Thanks for your reply @belemaire.

Were these checks introduced recently? My teammates seem to have been able to add these nativeapps using the CLI with an invalid semantic version more than once. Now that I'm trying to add a new version, even if I try adding a proper version, it won't work because it performs checks on the old versions.

Does that mean that my only solution is to start the cauldron from scratch or manually get rid of the previously added versions with invalid semantic versioning on the cauldron?

belemaire:
Hi @LuisRizo 
You are correct, sorry I was under the assumption that this was enforced. 
After looking into this, this restriction was introduced in 0.28.0 apparently, as a side effect of another commit. That being said, we never really ran into this issue because we've been using semver versioning for native app versions in cauldron for a while now.

[...]

PR #1089, opened 2019-03-15
Fix app versioning for cauldron add nativeapp command

belemaire:
Do not enforce a semver compliant version for native app versions in Cauldron.

Fixes https://github.com/electrode-io/electrode-native/issues/1082


belemaire:
Merging for 0.29.9 release

PR #1103, opened 2019-03-26
Bump node-simctl version

belemaire:
Fix `run-ios` command with`XCode 10.2`, not finding any device.

After investigation this is due to the `node-simctl` version used. Using version `4.0.2` is properly listing devices with `XCode 10.2`, also tested it with `XCode 10.1`.


belemaire:
It seems like this `run-ios` ios is not specifically due to XCode 10.2, but appeared with a specific version of node-simctl, independently of the XCode version installed.

PR #1156, opened 2019-05-14
Fix compile error for annotationProcessor native dependency on android

tianling536:


belemaire:
Thanks for your PR,

However I don't think this setting should be added by default inside the Container/Runner. 

From Android [build dependencies documentation](https://developer.android.com/studio/build/dependencies?utm_source=android-studio#annotation_processor), it is said that setting this one to true is not recommended and will be deprecated soon.

> If you experience issues after migrating your project's annotation processors to the processor classpath, you can allow annotation processors on the compile classpath by setting includeCompileClasspath to true. However, setting this property to true is not recommended, and the option to do so will be removed in a future update of the Android plugin.

We don't really want to set such settings for all users (especially ones that are to be deprecated). Also what if a user need to set it to `false` for some reason ? We cannot unfortunately accommodate every use cases through hard coded config.

There might be another way around the compilation error you are facing, but in case there is not, you could use the [git patch transformer](https://github.com/electrode-io/ern-container-transformer-git-patch) which allows for patching code and config files in the Container to fit any custom needs. In that case your patch file would be to add 

```
javaCompileOptions {
  annotationProcessorOptions {
    includeCompileClasspath true
  }
}
```

in the `build.gradle` of the Container.

For the runner you could do the change manually to the Android project, as the Runner is just generated once.

If you need help please let us know.

tianling536:
I resolved the issue by modify the manifest config file correctly,thanks a lot!

PR #1171, opened 2019-05-22
Do not mustache jniLibs when generating Android Container

belemaire:
Some Android native modules are including native libraries (`.so` files) to be stored in the `jniLibs` directory of the Container.

Because this directory is not included in the directories to be excluded from Mustache processing, this is leading to Container generation failure when such native modules are included in the Container.

This PR adds `lib/src/main/jniLibs` directory to the list of Mustache exclusions.


krunalsshah:
Interesting

PR #1173, opened 2019-05-22
Convert leftover js file in ern-api-gen to ts

belemaire:


ern-api-gen/src/java/MustacheWriter.ts
```diff
@@ -0,0 +1,70 @@
+import Mustache from 'mustache'
```
krunalsshah: import Mustache from './java/Mustache'

PR #1175, opened 2019-05-28
Add cwd param to runMiniApp function

belemaire:
To allow client to provide a custom `cwd`.


krunalsshah:
The caller of these methods may not be passing the attributes yet.

Issue #1060, opened 2019-01-28
Android missing `replaceInFile` directive


davidgovea:
`replaceInFile` is [listed as a platform-independent directive](https://native.electrode.io/reference/index-2/native-modules#platform-agnostic-directives), but it looks like it's not implemented on Android.

I'm working on a manifest entry for a more involved RN module: `react-native-gesture-handler`, which needs edits to the `ReactActivityDelegate` class definition (https://kmagiera.github.io/react-native-gesture-handler/docs/getting-started.html).

File replacement is not necessarily the best way to do this, but it is a very powerful directive that unlocks a lot of possibilities.

I have an initial implementation here: https://github.com/electrode-io/electrode-native/compare/master...runtrizapps:feature/android-replace-in-file?expand=1

---

Worth noting: I'm performing the replacements *after* the mustache templates are evaluated, which may differ from the approach on iOS (it seems like `replaceInFile` is designed to process incoming plugin files, not the final container files).

Perhaps a container transformer would be more appropriate? Not sure if there is a way to link a manifest-plugin entry with a container transformation..

Not submitting a PR at this time: just want to hear thoughts/rationale from the team. Happy to refactor or try out different approaches.

Thank you!

deepueg:
@davidgovea Thanks for opening this issue and providing suggestions. We will have a discussion around this to consider various options and let you the path forward. Thanks again.

davidgovea:
Live in `v0.34.0`, thank you!

PR #1197, opened 2019-06-04
WIP: Android `replaceInFile`

davidgovea:
addresses #1060 

Rebased the old POC up to master -- figured I'd create the PR to help discussion.

With [iOS's plugin processing](https://github.com/electrode-io/electrode-native/blob/8756615e1fae55b10a8de47c079a9db72de4f11f/ern-core/src/iosUtil.ts#L37), the 'hull filling' happens first. In android, it seems to happen [after the plugin processing](https://github.com/electrode-io/electrode-native/blob/master/ern-container-gen-android/src/AndroidGenerator.ts#L246), so we need to defer the replacements.

The `fs` use and replacement function was lifted from the iOS implementation, but was then wrapped in an `async` function side-effect creator for later processing. **This is probably inefficient**, but I held back on optimizations, pending discussion.

### DIFF
--- a/ern-container-gen-android/src/AndroidGenerator.ts
+++ b/ern-container-gen-android/src/AndroidGenerator.ts
@@ -23,6 +23,7 @@ import {
 
 import _ from 'lodash'
 import path from 'path'
+import fs from 'fs'
 import readDir from 'fs-readdir-recursive'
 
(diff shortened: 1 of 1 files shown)


davidgovea:
@belemaire updated & squashed

belemaire:
Thanks 👍

PR #1228, opened 2019-06-24
Update gradle config for api-impl-native android.

deepueg:
This PR provides few improvements to the native api implementation project for android.

### DIFF
--- a/ern-api-impl-gen/hull/android/build.gradle
+++ b/ern-api-impl-gen/hull/android/build.gradle
@@ -6,7 +6,7 @@ buildscript {
         google()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.0.1'
+        classpath 'com.android.tools.build:gradle:{{{androidGradlePlugin}}}'
 
         // NOTE: Do not place your application dependencies here; they belong
         // in the individual module build.gradle files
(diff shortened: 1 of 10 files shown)


belemaire:
@deepueg 
We should probably use same mustaching as for other projects for android versions (as can be seen in [Android Generator Hull](https://github.com/electrode-io/electrode-native/blob/v0.34/ern-container-gen-android/src/hull/lib/build.gradle#L15-L19) for example). This way with each release of Electrode Native we can bump all versions to latest in [one location](https://github.com/electrode-io/electrode-native/blob/55949decf15791a53feaf0792bb5ccac270c8a7c/ern-core/src/android.ts#L14-L20) only, while still allowing users to control versions on their own in Cauldron or through command line config. For now this is OK, but just FYI, we should make the move asap.

Also 

```
implementation 'com.android.support:appcompat-v7:28.0.0'
testCompile 'junit:junit:4.12'	    testImplementation 'junit:junit:4.12'
implementation 'androidx.appcompat:appcompat:1.0.2'
implementation 'androidx.lifecycle:lifecycle-extensions:2.0.0'
```

So we are moving Electrode Native to AndroidX ? At least starting with Native API Implementations it seems. Would have liked to make the move at once with 0.60 release of React Native, but that's OK, I don't think native api implementations are that widely used, but sounds like a breaking change for release notes. Also is that OK to have legacy `appcompat` along with `androidx.appcompat` ? I had the impression that they were somehow mutually exclusive, but I'm probably wrong.
Please let me know, if that's fine I'll just approve / merge.

deepueg:
@belemaire thanks for pointing out the common version mustaching. Updated. 

Regarding androidX, Only impl will move to androidX for now since we don't have many native implementations out there. Having said that, any project that wants to use a native impl library will need to add these two entries into their gradle.properies. This will only be the breaking change for the native side. 
```
android.useAndroidX=true
android.enableJetifier=true
```

belemaire:
Thanks for the details and mustaching versions :+1:

PR #1232, opened 2019-06-25
add noDuplicateReleaseError option to code push release and promote

### LINKED ISSUE #1230: [Feature] Support `--disable-duplicate-release-error` for CodePush release and promote
It would be helpful if we were able to forward the `--disable-duplicate-release-error` flag to the CodePush CLI that would allow the `promote` or `release` command to not fail if the same code is attempted to be released.

More info on the feature:
https://docs.microsoft.com/en-us/appcenter/distribution/codepush/cli#no-duplicate-release-error-parameter

krunalsshah:
add noDuplicateReleaseError option to code push release and promote

Fixes #1230

### DIFF
--- a/docs/cli/code-push/promote.md
+++ b/docs/cli/code-push/promote.md
@@ -85,6 +85,12 @@ If no `targetDescriptors` nor a `targetSemVerDescriptor` is specified, the comma
  
 * **Default** Empty string
 
+`--noDuplicateReleaseError`
+
+* When this flag is set, releasing a package that is identical to the latest release will produce a warning instead of an error
+
+* **Default** false
+
 #### Related commands
 
 [code-push release] | Issue a CodePush release 
--- a/docs/cli/code-push/promote.md
+++ b/docs/cli/code-push/promote.md
@@ -85,9 +85,9 @@ If no `targetDescriptors` nor a `targetSemVerDescriptor` is specified, the comma
  
 * **Default** Empty string
 
-`--noDuplicateReleaseError`
+`--disableDuplicateReleaseError`
 
-* When this flag is set, releasing a package that is identical to the latest release will produce a warning instead of an error
+* When this flag is set, promoting a package that is identical to the latest release on the target deployment will produce a warning instead of an error
 
 * **Default** false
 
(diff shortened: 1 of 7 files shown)


krunalsshah:
https://github.com/microsoft/code-push/blob/064d70ddaff16b131836ece3723505e01844b5fa/cli/definitions/cli.ts#L173

belemaire:
wow that was quick :)

Unfortunately from what I can see in sources of CodePush SDK / CLI, your changes won't have any effect :(

You pasted link to type definition of `code-push promote` CLI command, which accept this parameter. But we are not using the code-push CLI in Electrode Native, we are directly calling in the SDK.

SDK `promote` and `release` method, that we call, are getting a `PackageInfo` type as last parameter (as can be seen here in [release](https://github.com/microsoft/code-push/blob/master/sdk/script/management-sdk.ts#L302) method implementation). This `PackageInfo` 
does not contain the `noDuplicateReleaseError` property. As can be seen in the `PackageInfo` definition, as well as in the `release`/`promote` methods of the SDK there is no use what-so-ever of this property name.

It seems like the `noDuplicateReleaseError` flag, as stated by the documentation is done to only transform the error as a warning only. It will still not update the package on the server side, but it won't throw an error on the CLI side, but rather log a warning.

From what I can see, for `release` and `promote` CLI commands, code-push is just attaching an error handler when it invokes these SDK functions (for example [here for `promote` command](https://github.com/microsoft/code-push/blob/064d70ddaff16b131836ece3723505e01844b5fa/cli/script/command-executor.ts#L1145))

And the CLI error handler, [as implemented here](https://github.com/microsoft/code-push/blob/064d70ddaff16b131836ece3723505e01844b5fa/cli/script/command-executor.ts#L1530), just swallows the exception and log a warning instead, in case the `noDuplicateReleaseError` flag is set and the error is `AccountManager.ERROR_CONFLICT` indicating a duplicating package I would assume.

We should prob do the same, meaning that we should hook up an error handler to our calls of sdk `promote` and `release` methods, and swallow this error / log warning instead in case the `--disable-duplicate-release-error` is set.

krunalsshah:
Yup, that makes sense. Will amend the changes :) thanks for the analysis, it totally makes sense.

PR #1240, opened 2019-07-08
Fix yarn resolutions

belemaire:
Fix an issue with new `yarn resolutions` feature, that would overwrite patched files in `node_modules` as it runs `yarn install` after the files are patched :(


belemaire:
Merging for imminent `0.34.3` release

PR #1248, opened 2019-07-12
Auto resolve host ip for run-ios command

belemaire:
Auto resolve local machine IP for `run-ios` command so that it is not necessary to manually find IP of machine to provide it to the `--host` option of `run-ios`.
If `--host` option is set on the command line, the value provided will still be used rather than the auto resolved one.

Also adding `host`/`port` options support to `ern start`


belemaire:
Merging this one for `0.35.0` imminent release

PR #1250, opened 2019-07-12
Update Container and Runner project to use IP Address and Port

gmbharath12:
- Please refer https://github.com/electrode-io/electrode-native/pull/1248 for more info.

### DIFF
--- a/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
+++ b/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
@@ -47,12 +47,24 @@ static dispatch_semaphore_t semaphore;
 
 @implementation ElectrodeContainerConfig
 
-- (void) setupConfigWithDelegate:(id<RCTBridgeDelegate>)delegate {
+- (instancetype)init
+{
+    self = [super init];
+    if (self) {
+        //default values
+        self.packagerHost = @"localhost";
+        self.packagerPort = @"8081";
+    }
+    return self;
+}
+
+- (void)setupConfigWithDelegate:(id<RCTBridgeDelegate>)delegate {
     if ([delegate respondsToSelector:@selector(setJsBundleURL:)]) {
         NSURL *url;
         if (self.debugEnabled) {
-            url = [NSURL URLWithString:@"http://localhost:8081/{{{jsMainModuleName}}}.bundle?platform=ios&dev=true"];
-            NSLog(@"using local port to debug");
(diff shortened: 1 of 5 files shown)


ern-runner-gen-ios/src/hull/ErnRunner/AppDelegate.m
```diff
@@ -27,11 +27,12 @@ @implementation AppDelegate
 
 
 - (BOOL)application:(UIApplication *)application didFinishLaunchingWithOptions:(NSDictionary *)launchOptions {
-    // Start the container
+    // Configure the container
     ElectrodeContainerConfig *containerConfig = [[ElectrodeContainerConfig alloc] init];
     containerConfig.debugEnabled = RnDevSupportEnabled;
+    containerConfig.packagerHost = (PackagerHost)? PackagerHost : containerConfig.packagerHost;
+    containerConfig.packagerPort = (PackagerPort)? PackagerPort : containerConfig.packagerPort;
```
belemaire: I don't really understand these two lines logic.
Because we are inside the Runner and you import `RunnerConfig.h`, the `PackagerHost` and `PackagerPort` will always be defined, so I don't see the need of ternary here. 
And if they are not defined (won't be possible) then it does `containerConfig.packagerPort = containerConfig.packagerPort` which makes no real sense.
Could you clarify the need or just remove the ternary to just affect these variables with the values of `PackagerHost`/`PackagerPort` ?

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
```diff
@@ -47,12 +47,24 @@
...
+}
+
+- (void)setupConfigWithDelegate:(id<RCTBridgeDelegate>)delegate {
     if ([delegate respondsToSelector:@selector(setJsBundleURL:)]) {
         NSURL *url;
         if (self.debugEnabled) {
-            url = [NSURL URLWithString:@"http://localhost:8081/{{{jsMainModuleName}}}.bundle?platform=ios&dev=true"];
-            NSLog(@"using local port to debug");
+            // iOS device and Macbook must be on the same Wi-fi & metro packager (bundler) by default runs on 8081 port.
+            NSString *urlString = [NSString stringWithFormat:@"http://%@:%@/index.bundle?platform=ios&dev=true",self.packagerHost,self.packagerPort];
```
belemaire: Please keep mustache template placeholder `{{{jsMainModuleName}}}` rather than setting it to `index.bundle`. This is very important.

gmbharath12:
Refer https://github.com/electrode-io/electrode-native/pull/1251 Addressed all comments

PR #1256, opened 2019-07-23
Update native dependencies management in Cauldron

belemaire:
**Breaking Changes**

This PR removes the following commands : 
- cauldron update dependencies
- cauldron add dependencies
- cauldron del dependencies

It also removes options to add/update/del native dependencies through `cauldron batch` command.

Since version `0.31.0` of Electrode Native, it was only possible to use these commands for `orphaned` dependencies (i.e native dependencies that are not directly or transitively depended upon by any MiniApp or JS API implementation). 

In practice, 99% of applications won't have such `orphaned` native dependencies, making the commands quite useless and adding confusion around native dependencies management.

As a matter of fact, this PR gets rid of the concept of `orphaned` dependencies. Now all native dependencies need to be either a direct or transitive dependency of a MiniApp of JS API Implementation to be included in a Container.

Given this, this PR also modifies the way native dependencies are synced in Cauldron whenever a Container is regenerated. Prior to this PR, Electrode Native was taking a snapshot of the native dependencies prior to the Container generation, and after Container generation, and was doing a kind of diff between the two. The algorithm was mostly done to support the fact that prior to `0.31.0` it was possible to update native dependencies versions directly in Cauldron, but also to make sure that orphaned dependencies would remain in the native dependencies list in the Cauldron.
One major issue with this algorithm was that it was keeping native dependencies in Container around even if no MiniApps were depending on it (because orphaned dependencies are allowed) which led to some difficulties and issues in our mobile application.
Because we don't support orphaned dependencies anymore, the algorithm is now much simpler, and just set the native dependencies in Cauldron as the one present inside a Container after its generation. It won't keep any dependencies lingering around if they are not used in the current Container.


gmbharath12:
Sad to see this go. Love this `batch` command

PR #1260, opened 2019-07-25
Fix ern start .babelrc resolution

belemaire:
Fix an issue related to `.babelrc` proper resolution when using `ern start` with linked MiniApps.

Issue was due to `.babelrc` files from node modules to be processed when they shouldn't be.

For example, if a MiniApp `MiniAppA` is depending on a dependency, `DependencyA`, each of them having their own `.babelrc`, then the file structure would end up being something like this inside a composite :

```
.
├── index.js
├── node_modules
│   ├── DependencyA
│   │   ├── .babelrc
│   ├── MiniAppA
│   │   ├── .babelrc
```

However, using `ern start` with linked `MiniAppA`, the structure would look as follow, because we copy over the whole content of `MiniAppA` working directory to the `MiniAppA` directory in `node_modules` of composite:

```
.
├── index.js
├── node_modules
│   ├── MiniAppA
│   │   ├── .babelrc
│   │   ├── node_modules
│   │   │   ├── DependencyA
│   │   │   │   ├── .babelrc
```

The problem here is that when a MiniApp is setting `useBabelRc` we add an entry in `babelRcRoots` being the path of the MiniApp (for example here it would be `./node_modules/MiniAppA`). Unfortunately, Babel is resolving directories in babelRcRoots, recursively, meaning that it will process `node_modules/MiniAppA/.babelrc` but also `node_modules/MiniAppA/node_modules/DependencyA/.babelrc`. We don't want that, as we just want it to process the `.babelrc` of the MiniApp only, and not of any sub modules.

Note that this isn't an issue in composite, because `DependencyA` is hoisted, so not under `node_modules/MiniAppA` path. That being said, the issue could also appear in Composite for non hoisted dependencies containing a `.babelrc`.

The fix here, is to use a regular expression with negative look ahead for matching the MiniApp path in babelRcRoots, rather than using a fixed string path. 

To reuse previous example, what we now add in `babelRcRoots` will be `node_modules\/MiniAppA(?!.+\/)`)` rather than `/node_modules/MiniAppA`, which will only match `.babelrc` found at the root of the MiniApp but not any `.babelrc` from subdirectories.

### DIFF
--- a/ern-composite-gen/src/generateComposite.ts
+++ b/ern-composite-gen/src/generateComposite.ts
@@ -273,14 +273,16 @@ async function addBabelrcRoots({ outDir }: { outDir: string }) {
   //
   // ... is added to the babelRcRoots array, so that we can properly
   // configure Babel to process the .babelrc of these dependencies.
-  const babelRcRoots: string[] = []
+  const babelRcRootsRe: RegExp[] = []
+  const babelRcRootsPaths: string[] = []
   for (const dependency of dependencies) {
     if (fs.existsSync(path.join(compositeNodeModulesPath, dependency))) {
       const depPackageJson = await readPackageJson(
         path.join(compositeNodeModulesPath, dependency)
       )
       if (depPackageJson.ern && depPackageJson.ern.useBabelRc === true) {
-        babelRcRoots.push(`./node_modules/${dependency}`)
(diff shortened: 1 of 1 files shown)


belemaire:
Merging for imminent `0.36.1` release

PR #1261, opened 2019-07-25
Requalify error log to warning

belemaire:
Requalify error log to warning, as it is actually not causing any problem in most cases and wrongly leading users to thinking something went wrong.

We should add a strict mode to actually throw error if no configuration is found in manifest for some plugins.


gmbharath12:
🎉 👍

PR #1280, opened 2019-08-13
Add initial React Native 0.60 support for Android

belemaire:
Update Composite and Container generators for proper support of React Native 0.60.

**Composite generator update**

- For React Native 0.60, the react-native dependency need to be present inside the `dependency` object of the composite package.json. Otherwise metro bundler will complain.
Adding `react-native` dependency in composite package.json was only done if react-native-code-push native module was detected, as it had a similar requirement. 
However some other things were done in `patchPackageJsonForCodePush` function, for example setting `name` and `version` fields of package.json. This is not needed anymore though because composite generator have moved to use `yarn init` a while back, which already sets these fields. Because the `react-native` dependency needs to be set in composite package.json for code push and now react native (meaning it should always be there), code has just been refactored to get rid of `patchPackageJsonForCodePush` (now useless) and a new function `addReactNativeDependencyToPackageJson` was created to specifically add the react native dependency to the composite package.json.

- For React Native 0.60 (and above) there is no more need to patch metro resolver, as the bug being patched was only present in metro 0.51 (React Native 0.60 is using metro >= 0.53), so we just skip this function call.

**Android Container generator update**

- Starting with RN 0.60, JavaScriptCore engine (`libjsc.so` native library files for different architectures) is not distributed inside the React Native AAR any longer but distributed inside a separate AAR that is itself distributed as a node package. Also, this library now comes in two variants : `android-jsc` and `android-jsc-intl`.
Update to Android Container generator is to retrieve the node package containing the JavaScriptCore engine, extract (unzip) the AAR content, to get access to the `libjsc.so` files, and inject these files in the `jniLibs` of the Container, so that JavaScriptCore engine is shipped inside the Container. 
Also added code to let users choose between `android-jsc` or `android-jsc-intl` variant of JavaScriptCore (default to `android-jsc`, as react-native does), through container generator configuration, as well as to select the version of JavaScriptCore to be used (default to `245459`).

This PR does not add support for alternative JavaScript engine (Hermes), this will be added later on.

[`com.walmartlabs.ern:react-native:0.60.0`](https://repo1.maven.org/maven2/com/walmartlabs/ern/react-native/0.60.0/) AAR artifact has been published to Maven Central for proper testing/validation. We will publish latest version of 0.60 line (0.60.5) to Maven Central soon and this should be the version targeted for upcoming Electrode Native release (0.38.0).

Please note that `ern run-android` will not work with the default runner hull. It will only work with `hullx` introduced in 0.37.0 which has been converted to AndroidX. To get access to this hull, just add a dependency on `ern-navigation` to the MiniApp and regenerate the hull (by trashing the MiniApp `android` directory and running `ern run-android` command again).

### DIFF
--- a/docs/platform-parts/container/container-integration.md
+++ b/docs/platform-parts/container/container-integration.md
@@ -147,6 +147,30 @@ You can configure `androidConfig` in the cauldron as show below.
 }
 ```
 
+##### JavaScript Engine (RN 0.60 and above)
+
+Starting with React Native 0.60, the JavaScript engine is distributed separately from the React Native AAR.
+Also, prior to this version, JavaScriptCore was the only JavaScript engine that could be used on Android for React Native applications. Starting with this new version, it is now possible to use alternative JavaScript engines such as Hermes.
+
+With React Native 0.60.0, JavaScriptCore engine now comes in two variants : `android-jsc` and `android-jsc-intl`. The later is the international variant. It includes ICU i18n library and necessary data allowing to use e.g. Date.toLocaleString and String.localeCompare that give correct results when using with locales other than en-US. This variant is about 6MiB larger per architecture.
+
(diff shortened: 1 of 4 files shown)


friederbluemle:
Wow nice :+1:

deepueg:
👍

PR #1285, opened 2019-08-19
Update Android Container for rn>=0.60.1 support

belemaire:
React Native 0.60.1 introduced Hermes support.

One issue (that looks like a bug in React Native, introduced in 0.60.1 and still present in 0.60.5) is that `SoLoader.loadLibrary` method is called before `SoLoader.init` was called, resulting in a `java.lang.RuntimeException: SoLoader.init() not yet called` exception crashing the app.

This is due to a kinda chicken and egg problem, where `ReactInstanceManager` constructor calls [`initializeSoLoaderIfNecessary`](https://github.com/facebook/react-native/blob/v0.60.1/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManager.java#L220) to actually call `SoLoader.init`, but in `ReactInstanceManagerBuilder`, which creates a new instance of `ReactInstanceManager`, a call to [`getDefaultJSExecutorFactory`](https://github.com/facebook/react-native/blob/v0.60.1/ReactAndroid/src/main/java/com/facebook/react/ReactInstanceManagerBuilder.java#L304) is made, which triggers a call to `SoLoader.loadLibrary` before `SoLoader.init` was called.

This probably went unnoticed because application templates for React Native are actually calling `SoLoader.init` during `onCreate` as can be [seen here](https://github.com/facebook/react-native/blob/v0.60.5/template/android/app/src/main/java/com/helloworld/MainApplication.java#L47)

Solution in our case is just to make a call to `SoLoader.init` in the Android Container `initialize` method, before the `ReactInstanceManager` gets created. If for some reason React Native or app is doing it before or after our call, this won't cause any issues, thanks to the fact that `SoLoader.init` is idempotent.

### DIFF
--- a/ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
+++ b/ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
@@ -32,6 +32,9 @@ import com.facebook.react.modules.network.OkHttpClientFactory;
 {{/RN_VERSION_GTE_54}}
 import com.facebook.react.modules.network.OkHttpClientProvider;
 import com.facebook.react.shell.MainReactPackage;
+{{#RN_VERSION_GTE_60_1}}
+import com.facebook.soloader.SoLoader;
+{{/RN_VERSION_GTE_60_1}}
 {{#plugins}}
 import com.walmartlabs.ern.container.plugins.{{name}};
 {{/plugins}}
@@ -136,6 +139,10 @@ public class ElectrodeReactContainer {
             {{/apiImplementations}}
      ) {
         if (sElectrodeReactNativeHost == null) {
+            
+            {{#RN_VERSION_GTE_60_1}}
+            SoLoader.init(application, /* native exopackage */ false);
+            {{/RN_VERSION_GTE_60_1}}
 
             // ReactNative general config
 
--- a/ern-core/src/injectReactNativeVersionKeysInObject.ts
+++ b/ern-core/src/injectReactNativeVersionKeysInObject.ts
@@ -8,6 +8,7 @@ export function injectReactNativeVersionKeysInObject(
     RN_VERSION_GTE_45: semver.gte(reactNativeVersion, '0.45.0'),
     RN_VERSION_GTE_54: semver.gte(reactNativeVersion, '0.54.0'),
     RN_VERSION_GTE_59: semver.gte(reactNativeVersion, '0.59.0'),
+    RN_VERSION_GTE_60_1: semver.gte(reactNativeVersion, '0.60.1'),
     RN_VERSION_LT_54: semver.lt(reactNativeVersion, '0.54.0'),
     RN_VERSION_LT_58: semver.lt(reactNativeVersion, '0.58.0-rc.2'),
     RN_VERSION_LT_59: semver.lt(reactNativeVersion, '0.59.0'),


ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
```diff
@@ -27,6 +27,7 @@
 import com.facebook.react.ReactPackage;
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.SafeActivityStarter;
+import com.facebook.soloader.SoLoader;
```
friederbluemle: Would it makes sense to also wrap this import in a conditional? Otherwise it would result in an unused import if `RN_VERSION_GTE_60_1` is false.

ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
```diff
@@ -136,6 +137,10 @@ public synchronized static void initialize(@NonNull Application application, @No
             {{/apiImplementations}}
      ) {
         if (sElectrodeReactNativeHost == null) {
+            
+            {{#RN_VERSION_GTE_60_1}}
+            SoLoader.init(application, /* native exopackage */ false);
+            {{/RN_VERSION_GTE_60_1}}
```
friederbluemle: Just double checking: This is _not_ required for React Native 0.60.0?

ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
belemaire: Yes it would definitely make sense :+1: Adding it

ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
belemaire: Yes correct. Change was only introduced starting with RN 0.60.1 (for Hermes support)

ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
```diff
@@ -27,6 +27,9 @@
 import com.facebook.react.ReactPackage;
 import com.facebook.react.bridge.ReactContext;
 import com.facebook.react.bridge.SafeActivityStarter;
+{{#RN_VERSION_GTE_60_1}}
+import com.facebook.soloader.SoLoader;
```
friederbluemle: Any chance you could move this down to keep it in alphabetical order?

Otherwise
`com.facebook.soloader.SoLoader`
would be placed _above_
`com.facebook.react.modules.network.OkHttpClientFactory`

and could result in unnecessary reordering/line changes when optimizing imports.

ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
friederbluemle: Yes, you're right, it won't crash below RN 0.60.1 - I'm just wondering if there could be any other unintentional effects by _not_ calling it? It appears the call to `SoLoader.init` has been there for a long time in the RN templates (at least all the way back to React Native `0.53.0`). Are there any disadvantages to simply always call this method?

ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactContainer.java
belemaire: Done, but there are plenty of other imports in this file that are not in lexical order, so cleaning up the rest of the imports (not related to this PR) should be done through a separate PR.

It might also be very difficult for some of them because of the mustache templating that is grouping some imports based on some conditionals. 

for ex :
```
{{#conditionalA}}
com.a.b.c
com.e.f.g
{{/conditionalA}}
{{#conditionalB}}
com.c.d.e
{{/conditionalB}}
```

If both conditionals resolves to `true`, then the ordering of imports will be wrong, unless we make the mustache templating more complex. Need to find a right balance here, but I would rather favor the readability of the mustache templates rather than proper ordering given that these generated code is not really generally meant to be opened directly in Android Studio (apart from debugging purposes mostly done by our team).

friederbluemle: Yeah, definitely agree, it should be done in a separate PR 👍

Issue #1292, opened 2019-08-27
Generated Swift API: No types in responseCompletionHandler


axelander:
Trying to use the generated Swift API we just created, but can't seem to get the correct types on the responseCompletionHandler.
```swift
MyApi.init().requests.getMovie(id: 1) { (movie, error) in
  // movie is Any? here, expecting it to be Movie
}
```
Digging around in some tests it looks like I might need to cast this myself.
https://github.com/electrode-io/react-native-electrode-bridge/blob/1b554ef1d61d72b1459255c1b6b4387651d54067/ios/ElectrodeReactNativeBridgeTests/EventProcessorTests.swift#L33-L36

Is this really the only way to get the correct return types? The Kotlin API seems to be able to infer the correct types.

axelander:
I created a PR to get the correct types on the responseCompletionHandler https://github.com/electrode-io/electrode-native/pull/1294

friederbluemle:
According to the discussion, this was addressed in #1294.

PR #1294, opened 2019-08-28
Swift: Include type on responseCompletionHandler

axelander:
I'm trying to get the models/types in the generated Swift API client. This seems to do the trick. However, I am no Swift-developer so there might be a better way to do this 😛 . Is it possible to define some generics on the `ElectrodeBridgeResponseCompletionHandler` so we can avoid the casting from Any? -> MyModel ?

Issue: https://github.com/electrode-io/electrode-native/issues/1292

### DIFF
--- a/ern-api-gen/resources/swift/libraries/ern/api.mustache
+++ b/ern-api-gen/resources/swift/libraries/ern/api.mustache
@@ -81,7 +81,7 @@
 
     {{#operation}}
     {{^isEvent}}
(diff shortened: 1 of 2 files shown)


friederbluemle:
@axelander - Thanks for your PR! Please rebase your branch. Do not use the "Update branch" button (all it does is create a merge commit on your branch).

deepueg:
@axelander thanks for this PR. Generics definitely makes the API more readable and usable. 
We need to test this thoroughly for backward compatibility, various data types etc as we have plenty of APIs that are being generated and regenerated frequently.  We will take some time to asses this, hopefully by next week and merge this PR.

axelander:
@deepueg thanks, that sounds great!

deepueg:
@alexrwegener Thanks for your contribution.

PR #1297, opened 2019-08-30
Add support for dependencies assets exported through react-native.config.js

belemaire:
Add initial support of new React Native configuration mechanism introduced in React Native 0.60 through `react-native.config.js` configuration files. 

This initial support is only dealing with `assets` (fonts only) exported by packages through `react-native.config.js` configuration (formerly done through `rnpm` configuration, now deprecated). Fonts exported by packages through this new configuration file, will now be properly added to the Container.

This will be shipped in patch release `0.37.4` to unblock some internal teams.


ern-container-gen/src/copyRnConfigAssets.ts
```diff
@@ -0,0 +1,47 @@
...
+  handleCopyDirective,
+  NativePlatform,
+  findDirectoriesHavingRnConfig,
+  shell,
+} from 'ern-core'
+import { getAssetsPath } from './getAssetsPath'
+import readDir from 'fs-readdir-recursive'
+import path from 'path'
+
+export const supportedAssetsExts = ['.ttf', '.otf']
```
krunalsshah: what is the context behing supporting 2 font extensions?

belemaire: As far as I know, they are the only two font formats supported by iOS/Android

krunalsshah: There is no clear documentation though that's the closest I found 
Android : https://developer.android.com/guide/topics/resources/font-resource?authuser=1 
iOS : https://developer.apple.com/documentation/uikit/text_display_and_fonts/adding_a_custom_font_to_your_app

belemaire: Yes, so for iOS `ttf` and `otf`. For Android doc is referring to resources kept in the `res` directory as resources, not the `assets` directory that is the target directory used by React Native. Googling around I can see that for `ttc` support in assets you need some extra code. Should be good for `ttf` and `otf`, these are the most common font types anyway

belemaire:
Merging for 0.37.4 release

PR #1305, opened 2019-09-10
deprecating `finishedCallBack`, use `finish` to get a json

liulianci:
chgne Hull project because we updated NavigationAPIImpl, and regen fixture

### DIFF
--- a/ern-runner-gen-ios/src/hull/ErnRunner/ViewController.m
+++ b/ern-runner-gen-ios/src/hull/ErnRunner/ViewController.m
@@ -18,7 +18,8 @@
 #import "RunnerConfig.h"
 #import <ElectrodeContainer/ElectrodeContainer.h>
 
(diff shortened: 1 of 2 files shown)


ern-runner-gen-ios/src/hull/ErnRunner/ViewController.m
```diff
@@ -47,20 +49,39 @@ - (void)prepareForSegue:(UIStoryboardSegue *)segue sender:(id)sender {
...
-    
+
 - (NSDictionary * _Nullable)properties {
     return nil;
 }
-    
+
+//`finisedCallback` is deprecated, use `finish`
+//finishedCallback, finish and nativeNavigation are optional properties
 - (MiniAppFinishedCallback _Nullable)finishedCallback {
```
deepueg: Can we remove this from the runner, since this is a code that is generated we don't need to keep this here? Perhaps we can add a deprecated note in the Protocol level.

PR #1307, opened 2019-09-10
Regenerate fixtures and exclude Libraries

friederbluemle:
The update to RN 0.60.5 brought in a **ton** of changes to the generated files.

This PR also removes the `Libraries` folders from fixtures.


friederbluemle:
Good idea @belemaire - Should we try to ignore `system-tests/**/Libraries` as part of this PR (before merging it)? Any other changes that are required when we do that?

friederbluemle:
```
system-tests/fixtures/api-impl-native/ReactNativeErnmovieApiImplNative/ios/ElectrodeApiImpl/Libraries
system-tests/fixtures/ios-container/ElectrodeContainer/Libraries
```
These two folders together contain more than 2000 files and more than 160.000 lines of code - Removing them from source control would be great! @belemaire - How would this impact CI etc.? Would it increase build/test time? Any build scripts that need to be updated to regenerate these folders every time?

belemaire:
@friederbluemle 
Yes sounds good. Could you just `git rm` everything under `system-tests/fixtures/ios-container/ElectrodeContainer/Libraries` and add `system-tests/fixtures/ios-container/ElectrodeContainer/Libraries/` to the `.gitignore` as part of this PR.
There are some changes that are needed for system tests to properly exclude this `Libraries` directory from comparison during system tests as it will still be generated part of the tests. I have opened https://github.com/electrode-io/electrode-native/pull/1308 to handle this.

belemaire:
@friederbluemle 
One sec, will update #1308 to get rid of `system-tests/fixtures/api-impl-native/ReactNativeErnmovieApiImplNative/ios/ElectrodeApiImpl/Libraries` as the change was only made for `system-tests/fixtures/ios-container/ElectrodeContainer/Libraries`

belemaire:
@friederbluemle 
Please also `git rm` path `system-tests/fixtures/api-impl-native/ReactNativeErnmovieApiImplNative/ios/ElectrodeApiImpl/Libraries` as part of this PR and add to `.gitignore` in addition to the `system-tests/fixtures/ios-container/ElectrodeContainer/Libraries`
Will lighten the repo :)

belemaire:
Ok, just updated #1308 to exclude the `Libraries` directory of native implementation fixture.

friederbluemle:
Awesome! :+1: Which PR should be merged first?

belemaire:
@friederbluemle 
Ideally mine should be merged first otherwise if the directories are not there anymore (through your PR) then system tests will still generate the `Libraries` and try to compare with a fixture that does not have this directory (because your PR will remove these directories from the repo) and will fail.

friederbluemle:
Cool! I'll also include a couple of other minor consolidations for `.gitignore` at the same time 👍

PR #1312, opened 2019-09-12
set always_embed_swift_swift_standard_libraries

liulianci:
we don't need to `always_embed_swift_swift_standard_libraries`, this is only needed if it is a pure objective c framework and depending on some swift module


belemaire:
Closing this one in favor of setting this specific config through transformer configuration in Cauldron for now

PR #1316, opened 2019-09-17
Legacy code cleanup and refactor

belemaire:
Since [v0.31.0](https://github.com/electrode-io/electrode-native/releases/tag/v0.31.0) of Electrode Native, we are not any-longer downloading plugins but rather directly picking them inside the composite. Some code was leftover to actually download the plugin if it wasn't found in composite, which can not happen anymore. This PR gets rids of this legacy logic/code, simplifying the logic to only look into the composite for the plugins source. 
It also get rids of some parameters in api implementation generators that were tied to this extra logic, and some cleanup along the way, making the methods a bit leaner.


friederbluemle:
Any chance we can exclude `system-tests/fixtures/api-impl-native/ReactNativeErnmovieApiImplNative/yarn.lock`? - Or is there anything useful we can verify by having this system-tests fixture lock file in source control?

belemaire:
@friederbluemle :+1: yes these yarn.lock are useless (not only the one in apiimpl but all the ones in the  in system-test fixtures) they are anyway ignored when system tests are comparing fixture vs generated code. I'll remove them / git ignore them via a different dedicated PR

friederbluemle:
@belemaire - Sounds good, thanks! I also have a PR coming soon with some changes related to the lock files.

PR #1317, opened 2019-09-17
chore: change the badge to point to azure pipeline

krunalsshah:
chore: change the badge to point to azure pipeline


belemaire:
You picked the whole pipeline badge, I think it picks system tests only on master, not even sure.

Could you rather pick the one for system tests and the one for unit tests and format it in somewhat this way :

| Test Suite | Status |
| --- | ----------- |
| Unit Tests |  [Build Status](https://dev.azure.com/ElectrodeNative/Electrode%20Native/_build/latest?definitionId=1&branchName=master) |
| System Tests | [Build Status](https://dev.azure.com/ElectrodeNative/Electrode%20Native/_build/latest?definitionId=1&branchName=master) |

Will add a daily unit test run on master piror to dailu system tests run, so that we get a non 'unknown' status for unit tests.

PR #1324, opened 2019-09-19
Expose size flexibility of RCTRootView

gmbharath12:
- Instead of exposing RCTRootView, we expose just the size flexibility. 
-  This is an enum to define size flexibility type of the root view. If a dimension is flexible, the view will recalculate that dimension so the content fits. After a recalculation, `rootViewDidChangeIntrinsicSize` method of the RCTRootViewDelegate will be called.

### DIFF
--- a/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
+++ b/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
@@ -112,6 +112,14 @@ NS_ASSUME_NONNULL_BEGIN
                      properties:(NSDictionary *_Nullable)properties;
 
 /**
+ @param name The name of the mini app, that is registered with the AppComponent.
+ @param properties initialprops for a React Native miniapp.
+ @param sizeFlexibilty defines size flexibility type of the root view
+ @return UIView
+ */
+- (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *_Nullable)properties sizeFlexibility:(NSInteger)sizeFlexibilty;
+
+/**
  Call this to update an RCTRootView with new props. Calling this with new props will cause the view to be rerendered.
  Request will be ignored if the returned view is not an RCTRootView instance.
  */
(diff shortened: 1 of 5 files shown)


ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
```diff
@@ -159,6 +159,14 @@ - (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *)prop
     return rootView;
 }
 
+- (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *_Nullable)properties
+            withSizeFlexibility:(NSInteger)sizeFlexibilty {
+    RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:self.bridge moduleName:name initialProperties:properties];
+    rootView.sizeFlexibility = (RCTRootViewSizeFlexibility)sizeFlexibilty;
+    rootView.backgroundColor = [[UIColor alloc] initWithRed:1.0f green:1.0f blue:1.0f alpha:1];
```
deepueg: Do we need to set the background color as well? Or only sizeFlexibility?

deepueg:
Can you please squash these two commits.

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
gmbharath12: Good question! By default I think it returns a black color view (need to test it). So I just maintained the same `RGB` values as other methods do.

liulianci:
lgtm

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
```diff
@@ -111,6 +111,14 @@ NS_ASSUME_NONNULL_BEGIN
 - (UIView *)miniAppViewWithName:(NSString *)name
                      properties:(NSDictionary *_Nullable)properties;
 
+/**
+ @param name The name of the mini app, that is registered with the AppComponent.
+ @param properties initialprops for a React Native miniapp.
+ @param sizeFlexibilty defines size flexibility type of the root view
+ @return UIView
+ */
+- (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *_Nullable)properties withSizeFlexibility:(NSInteger)sizeFlexibilty;
```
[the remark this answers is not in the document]
gmbharath12: Done.

PR #1329, opened 2019-09-24
missed `{` in api.mustache

liulianci:
add it back to fix the compiling issue


belemaire:
Issue was introduced in https://github.com/electrode-io/electrode-native/pull/1294
Not released in any version

PR #1332, opened 2019-09-25
Expose React Native delegate

qw1024:
Adding a new method to expose `RCTRootViewDelegate ` to native app so that it could use it to update the mini app's view frame or height / width constraint layout accordingly. 

`-rootViewDidChangeIntrinsicSize:` will be called by React Native after the mini app's root view's intrinsic content size is changed.

### DIFF
--- a/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
+++ b/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
@@ -50,6 +50,10 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy) NSString *packagerPort;
 @end
 
+@protocol MiniAppViewDelegate <NSObject>
+- (void)rootViewDidChangeIntrinsicSize:(UIView *)rootView;
+@end
+
 ////////////////////////////////////////////////////////////////////////////////
 #pragma mark - ElectrodeReactNative
 /**
@@ -101,23 +105,42 @@ NS_ASSUME_NONNULL_BEGIN
                            properties:(NSDictionary * _Nullable)properties;
 
 /**
- Returns a react native miniapp (from a JSBundle) inside a view controller.
+ Returns a react native miniapp (from a JSBundle).
 
  @param name The name of the mini app, preferably the same name as the jsbundle
  without the extension.
  @param properties Any configuration to set up the mini app with.
- @return A UIViewController containing the view of the miniapp.
+ @return a UIView of the miniapp.
  */
 - (UIView *)miniAppViewWithName:(NSString *)name
                      properties:(NSDictionary *_Nullable)properties;
 
 /**
+ Returns a react native miniapp (from a JSBundle).
+
  @param name The name of the mini app, that is registered with the AppComponent.
  @param properties initialprops for a React Native miniapp.
  @param sizeFlexibilty defines size flexibility type of the root view
- @return UIView
+ @return a UIView of the miniapp.
  */
-- (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *_Nullable)properties sizeFlexibility:(NSInteger)sizeFlexibilty;
+- (UIView *)miniAppViewWithName:(NSString *)name
+                     properties:(NSDictionary *_Nullable)properties
+                sizeFlexibility:(NSInteger)sizeFlexibilty
+        __attribute((deprecated("use -miniAppViewWithName:properties:sizeFlexibility:delegate instead")));
+
+/**
+ Returns a react native miniapp (from a JSBundle).
+
+ @param name The name of the mini app, that is registered with the AppComponent.
+ @param properties initialprops for a React Native miniapp.
+ @param sizeFlexibilty defines size flexibility type of the root view
+ @param delegate
+ @return a UIView of the miniapp.
+ */
+- (UIView *)miniAppViewWithName:(NSString *)name
+                     properties:(NSDictionary *_Nullable)properties
+                sizeFlexibility:(NSInteger)sizeFlexibilty
+                       delegate:(id<MiniAppViewDelegate> _Nullable)delegate;
 
 /**
  Call this to update an RCTRootView with new props. Calling this with new props will cause the view to be rerendered.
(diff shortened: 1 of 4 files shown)


gmbharath12:
👍 LGTM

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
```diff
@@ -101,23 +105,29 @@ NS_ASSUME_NONNULL_BEGIN
...
  @param sizeFlexibilty defines size flexibility type of the root view
- @return UIView
+ @param delegate of the root view that handles intrinsic size updates
+ @return a UIView of the miniapp.
  */
-- (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *_Nullable)properties sizeFlexibility:(NSInteger)sizeFlexibilty;
+- (UIView *)miniAppViewWithName:(NSString *)name
+                     properties:(NSDictionary *_Nullable)properties
+                sizeFlexibility:(NSInteger)sizeFlexibilty
+                       delegate:(id<MiniAppViewDelegate> _Nullable)delegate;
```
liulianci: could you write a new method for this? this might break the existing feature which is using this method

deepueg:
@qw1024 Can you please provide more details to this PR. And this would be a breaking change since we are updating the method signature.

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
```diff
@@ -50,6 +50,10 @@ NS_ASSUME_NONNULL_BEGIN
 @property (nonatomic, copy) NSString *packagerPort;
 @end
 
+@protocol MiniAppViewDelegate <NSObject>
+- (void)rootViewDidChangeIntrinsicSize:(UIView *)rootView;
```
deepueg: Who will trigger this method call? Is it supposed to be invoked after this [line](https://github.com/electrode-io/electrode-native/pull/1332/files#diff-7cae239e661f25961b5b8e45e627f218R167)  ?

liulianci: this is getting called when rootView's content size updated to a new size @deepueg

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
qw1024: done

friederbluemle:
@qw1024 - Please squash the three commits. Also: Are you intentionally using a different author and committer id for the first commit?

```
commit 0af19ed91d9cadc9618b75f818eee97f6a300b5d
Author:     Wei Qiu <newtown2@gmail.com>
AuthorDate: Wed Sep 25 14:05:39 2019 -0400
Commit:     Wei Qiu <vn0jsqy@wal-mart.com>
CommitDate: Wed Sep 25 14:17:28 2019 -0400

    Expose React Native delegte
```

qw1024:
@friederbluemle Squashed the commits. No different author and committer is not my intention but caused by forgetting change repo setting.

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
```diff
@@ -101,23 +105,42 @@ NS_ASSUME_NONNULL_BEGIN
...
+        __attribute((deprecated("use -miniAppViewWithName:properties:sizeFlexibility:delegate instead")));
+
+/**
+ Returns a react native miniapp (from a JSBundle).
+
  @param name The name of the mini app, that is registered with the AppComponent.
  @param properties initialprops for a React Native miniapp.
  @param sizeFlexibilty defines size flexibility type of the root view
- @return UIView
+ @param delegate of the root view that handles intrinsic size updates
```
liulianci: oh BTW, could you remove this comment?

friederbluemle: @qw1024 - While you fix this up, please also remove `PR comments` from the commit message body.

qw1024: done

ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
```diff
@@ -159,14 +159,26 @@ - (UIView *)miniAppViewWithName:(NSString *)name properties:(NSDictionary *)prop
...
 }
 
+- (UIView *)miniAppViewWithName:(NSString *)name
+                     properties:(NSDictionary *_Nullable)properties
+                sizeFlexibility:(NSInteger)sizeFlexibilty
+                       delegate:(id<MiniAppViewDelegate> _Nullable)delegate {
+    RCTRootView *rootView = [[RCTRootView alloc] initWithBridge:self.bridge moduleName:name initialProperties:properties];
+    rootView.sizeFlexibility = (RCTRootViewSizeFlexibility)sizeFlexibilty;
+    rootView.backgroundColor = [[UIColor alloc] initWithRed:1.0f green:1.0f blue:1.0f alpha:1];
+    rootView.delegate = delegate;
```
deepueg: Don't we need to change the order like below to get the `rootViewDidChangeIntrinsicSize` to get triggered when the `sizeFlexibility` is set?
```
rootView.delegate = delegate;
rootView.sizeFlexibility = (RCTRootViewSizeFlexibility)sizeFlexibilty;
    rootView.backgroundColor = [[UIColor alloc] initWithRed:1.0f green:1.0f blue:1.0f alpha:1];
```

qw1024: Good point. In this case we don't - setting `sizeFlexibility` simply sets the property and calls `setNeedsLayout`, which only invalidate current layout but does not force an immediate update, instead waits for the next update cycle. The `rootViewDidChangeIntrinsicSize` delegate method, as a result, won't be called until the next update cycle. Therefore, the order of setting `sizeFlexibility ` and `delegate` does not matter.

deepueg: Thanks for the clarification.

deepueg:
Please rebase.

PR #1345, opened 2019-10-02
Fix android native api implementation config

belemaire:
Fix issue with custom Android permissions and repositories configured in native api implementation package.json, not being injected in the generated Container.

Fixed through a quick small refactoring, centralizing plugin configuration inject in AndroidGenerator but also taking care of deduping repositories/permissions which was not done previously (small cherry on the cake in addition to the fix ;)).


krunalsshah:
@belemaire thanks for the fix. 

Assuming no fixtures require change for `api-impl`?

belemaire:
@krunalsshah No fixture change

PR #1347, opened 2019-10-04
Fix runner backcompatibility

deepueg:
This is to fix an issue when the previously generated miniapps tries to execute `ern run-android` command, the container name was not reflecting the one in `build.gradle` causing a build failure.

### DIFF
--- a/ern-orchestrator/src/runMiniApp.ts
+++ b/ern-orchestrator/src/runMiniApp.ts
@@ -161,14 +161,28 @@ export async function runMiniApp(
     outDir,
   })
 
+  const pathToRunner = path.join(cwd, platform)
+  const oldAndroidRunner = isOldAndroidRunner()
+
   if (platform === 'android') {
+    if (oldAndroidRunner) {
+      log.warn(
+        'Looks like you are running on an older version of the android runner project generated using ern version < 0.38.0. \n We recommend that you delete your current Android directory and regenerate the new runner by executing `ern run-android` command again.'
+      )
+    }
     extra = extra || {}
     extra.androidConfig = {
       ...(extra && extra.androidConfig),
-      artifactId: `runner-ern-container-${entryMiniAppName.toLowerCase()}`,
+      artifactId: oldAndroidRunner
+        ? `runner-ern-container`
+        : `runner-ern-container-${entryMiniAppName.toLowerCase()}`,
       groupId: 'com.walmartlabs.ern',
-      packageFilePath: `com/walmartlabs/ern/${entryMiniAppName.toLowerCase()}`,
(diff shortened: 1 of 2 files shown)


ern-orchestrator/src/runMiniApp.ts
```diff
@@ -161,14 +161,28 @@ export async function runMiniApp(
     outDir,
   })
 
+  const pathToRunner = path.join(cwd, platform)
+  const oldAndroidRunner = isOldAndroidRunner()
+
   if (platform === 'android') {
+    if (oldAndroidRunner) {
+      log.warn(
+        'Looks like you are running on an older version of the adroid runner project generated using ern version < 0.38.0. \n We recommend that you delete your current Android directory and regenerate the new runner by executing `ern run-android` command again.'
```
belemaire: typo `adroid` (should be `android`)

ern-orchestrator/src/runMiniApp.ts
```diff
@@ -238,10 +239,13 @@ export async function runMiniApp(
   function isOldAndroidRunner(): boolean {
     const paths = path.join(
       pathToRunner,
-      'app/src/main/java',
-      extra.androidConfig.packageFilePath
+      'app/src/main/java/com/walmartlabs/ern/RunnerConfig.java'
     )
-    return !fs.existsSync(paths)
+
+    if (extra && extra.forceNewAndroidRunner) {
```
belemaire: Having trouble understanding the purpose of `forceNewAndroidRunner`. Has it only been done to support tests ?

ern-orchestrator/src/runMiniApp.ts
friederbluemle: Also - Related question (more general, not about this PR): When we suggest to "regenerate" the project - What happens if the user has made some modifications to the native files? Maybe we can discuss this separately at some point.. ;)

ern-orchestrator/src/runMiniApp.ts
deepueg: yes, only for tests. Perhaps we can have a way to return true for this method while running tests ?

ern-orchestrator/src/runMiniApp.ts
deepueg: Those changes will have to be reapplied. Perhaps We can have a mention of that here or altogether remove this warning.

ern-orchestrator/src/runMiniApp.ts
belemaire: Yes. It's bad to have code that exists only to facilitate unit tests, usually hides some kind of design flaw or just quick shortcut. Do not bother with this one, will have a look at it, will send you updated content

PR #1354, opened 2019-10-08
introduce addition of Kotlin plugin via config

krunalsshah:
Problem
[react-native-webview](https://github.com/react-native-community/react-native-webview/tree/master/android/src/main/java/com/reactnativecommunity/webview/events) Native Module packages Kotlin files in its source. This leads to errors during container generation or while running local runner where `uploadArchive` job fails with a compilation error.

Subsequent PR to make `compileOptions` configurable to follow.


krunalsshah:
I'd have to include a conditional to include the kotlin gradle plugin
https://developer.android.com/studio/releases/gradle-plugin
Supported `3.0.0+`

deepueg:
Does it mean that we will have these plugins added to the container even if the app is not using this plugin?

krunalsshah:
There is 1 improvement I am making, is to have `kotlin` plugin included only if gradle plugin is greater than or equal to `3.0.0`
Even with the addition of `kotlin`gradle plugin, the build file looks fairly standard. Another way to add more selective check is to read the `build` file of the plugin and check if it includes kotlin plugin and include only than over `3.0.0` comparison?
Thoughts @belemaire @deepueg @friederbluemle

belemaire:
@krunalsshah 
No need to do this improvement for gradle check version, not sure who would forcefully set gradle version used by Container to a version less than 3.0.0. Does not make any sense at all, so pkease don't make things more complex than they're already are ;P

belemaire:
And size of binary won't be impacted, these are gradle plugins used to perform the build, not libraries shipped inside the Container.

belemaire:
That being said, one thing I fea is that kotlin need some dependency on [`org.jetbrains.kotlin:kotlin-stdlib`](https://kotlinlang.org/docs/reference/using-gradle.html#configuring-dependencies). If the native application is also built using Kotlin, this might be an issue as there might be a mismatch in term of version and we might run into some scenarios similar to OkHttp, where we would have to use `forceResolution` ... 
Need to look into this more closely.

friederbluemle:
I just had a quick call with Krunal.. The hull changes here will be useful at the point that the React Native default boilerplate project includes the kotlin dependencies as well. Until then, let's try to keep our hull as close as possible to what's shipped with RN.

Regarding react-native-webview - Since there are very minimal changes required to restore Java-only compilation, we can use our own fork until the changes are merged upstream (Krunal and I will work on submitting a PR).

belemaire:
Ok sounds good, that's the less riskier road indeed. 
That being said, I don't see much problem moving down this road especially if we start seeing / need adding more native modules that are built using Kotlin. Would say that as soon as we have another need for a native module written in Kotlin we'd have to have this ready (if not before) don't want to start forking all native modules, does not scale ;P
Also @krunalsshah  regarding the kotlin stdlib, because it has ideally to match the kotlin version used (as per kotlin guidelines), we should not add it in the manifest plugin configuration, but we should inject it per default in the container, with the version of kotlin being configured for the container. As done [here](https://github.com/electrode-io/electrode-native/blob/7145c0ef25d1a7e5bb7065c10f37eb9788676fea/ern-container-gen-android/src/AndroidGenerator.ts#L287) for example.
So something like : 
```javascript
dependencies.regular.push(`org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlinGradlePlugin}`)
```
You might want to rename your `kotlinGradlePlugin` to `kotlinVersion` to better align with other namings and also because the version used is really the kotlin version which is aligned by kotlin team to the gradle plugin / stdlib etc.
No rush though if we're not to merge this PR yet, but let's keep it open so that we can do some testing to ship it at some point.

krunalsshah:
@deepueg @friederbluemle @belemaire thanks for the review and feedback.

I'll keep the logic like https://github.com/electrode-io/electrode-native/pull/1354#issuecomment-539696507

belemaire:
Closing this one as we are not planning to merge it yet, it does not make sense to keep it lingering here. Can always be reopened and updated once needed.

PR #1370, opened 2019-10-22
Bump yargs to 14.2.0

belemaire:


friederbluemle:
Also update `yarn.lock`?

belemaire:
This version bump didn't update the `yarn.lock` because `yargs@^14.2.0` entry was already present in the lock file (set through https://github.com/electrode-io/electrode-native/pull/1366  (lerna `3.18.1` is depending on `yargs@^14.2.0`)).

friederbluemle:
Ooh okay got it. :+1:

PR #1377, opened 2019-10-24
read default schema from a file

krunalsshah:
read default schema from a file

### DIFF
--- a/ern-api-gen/initialApiSchema.json
+++ b/ern-api-gen/initialApiSchema.json
@@ -1,0 +1,105 @@
+{
+  "swagger": "2.0",
+  "info": {
+    "description": "Walmart Item Module",
+    "title": "WalmartItem",
+    "contact": {
+      "name": "ERN Mobile Platform Team"
+    }
+  },
+  "paths": {
+    "/items": {
+      "get": {
+        "tags": [
+          "WalmartItem"
+        ],
+        "description": "Returns all items from the system that the user has access to",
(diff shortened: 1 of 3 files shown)


ern-api-gen/src/generateProject.ts
```diff
@@ -105,112 +106,7 @@ export async function generateInitialSchema({
...
-        },
-        "desc": {
-          "type": "string"
-        }
-      }
-    }
-  }
-}
-`
+    : fileUtils.readFile(path.join(__dirname, '..', INITIAL_SCHEMA_FILE))
```
friederbluemle: Both paths of the conditional are now `fileUtils.readFile(..)`. Can this be simplified?

ern-api-gen/initialApiSchema.json
```diff
@@ -0,0 +1,105 @@
...
+        "name": {
+          "type": "string"
+        },
+        "desc": {
+          "type": "string"
+        }
+      }
+    }
+  }
+}
```
friederbluemle: Missing newline at EOF.

krunalsshah: I remember the atom editor adds the EOF automatically. Looks will have to set the option to do so for Visual Studio Code.

ern-api-gen/src/generateProject.ts
krunalsshah: 👍 modified the conditional

PR #1379, opened 2019-10-24
Set octokit to fixed version

belemaire:
Set `octokit/rest` to fixed `16.33.1` version to avoid an ongoing issue with `ern github` subcommands failing with such errors :

```
An error occurred: [BRANCH_NAME] is not a valid git ref of [GITHUB_REPO_URL]
```

Issue is impacting some users (and our CI jobs), so will release `0.38.7` with this fix while looking more closely as to why `16.34.0` of `octokit/rest` broke things.

### DIFF
--- a/ern-core/package.json
+++ b/ern-core/package.json
@@ -56,7 +56,7 @@
   },
   "license": "Apache-2.0",
   "dependencies": {
-    "@octokit/rest": "^16.28.6",
+    "@octokit/rest": "16.33.1",
     "@yarnpkg/lockfile": "^1.1.0",
     "archiver": "^2.1.1",
     "bluebird": "^3.5.1",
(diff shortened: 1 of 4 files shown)


belemaire:
Looks like this is where the problem might be coming from
https://github.com/octokit/rest.js/commit/e891d4e31c41e7a6afe0c8257404bf1f51607f5e#diff-0071a9eb9e809c9aacfc547f7514add6L1066

When trying to access existing branch `foo` of `myorg/myrepo` through GitHub API :

```
https://[GITHUB_URL]/api/v3/repos/myorg/myrepo/git/refs/heads/foo
```

works and return proper response body with 200 OK

while

```
https://[GITHUB_URL]/api/v3/repos/myorg/myrepo/git/ref/heads/foo
```

returns error body with 404 Not Found.

Will look more closely to it

belemaire:
Merging for imminent `0.38.7` release

belemaire:
Related to https://github.com/octokit/rest.js/issues/1506

PR #1372, opened 2019-10-23
fix a bug on customPermissions

w000xj:
encountered "Cannot read property 'push' of undefined" when generate a android container

PR #1387, opened 2019-10-30
Join string literal path segments

belemaire:
Quick refactoring to make paths more compact / readable, by joining string literal path segments, where applicable. For joins that end up as a single string, replace `path.join` with `path.normalize`, given that there is nothing left to join.

### DIFF
--- a/ern-container-gen-android/src/AndroidGenerator.ts
+++ b/ern-container-gen-android/src/AndroidGenerator.ts
@@ -92,7 +92,7 @@ export default class AndroidGenerator implements ContainerGenerator {
     let mustacheView: any = {
       customPermissions: [],
       customRepos: [],
-      permissions: []
+      permissions: [],
     }
     injectReactNativeVersionKeysInObject(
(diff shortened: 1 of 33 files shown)


ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -92,7 +92,7 @@ export default class AndroidGenerator implements ContainerGenerator {
     let mustacheView: any = {
       customPermissions: [],
       customRepos: [],
-      permissions: []
+      permissions: [],
```
friederbluemle: Yay for trailing comma 👍 Can we make this a default?

belemaire: Already the case, enforced by prettier on pre-commit hook. Got bypassed by the external contributor that changed this file (https://github.com/electrode-io/electrode-native/pull/1372).. didn't pay too much attention on this one (didn't even squash merged).

PR #1392, opened 2019-11-04
regenerate the fixtures to fix the system test

krunalsshah:
Fix the failing system tests
https://dev.azure.com/ElectrodeNative/Electrode%20Native/_build/results?buildId=1856&view=logs&s=a91c8f70-7e8a-583d-a65c-db3cf7ee37c6


ern-runner-gen-ios/test/regen-fixtures.ts
```diff
@@ -16,6 +16,7 @@ getRunnerGeneratorForPlatform('ios')
     },
     mainMiniAppName: 'dummy',
     outDir: pathToGeneratedFixtures,
+    reactNativeVersion: '0.60.5',
```
friederbluemle: Just wondering: Was this one necessary to fix the fixtures? What does it do/why do we need it?

krunalsshah: To be able to execute the `regen-fixtures` script in `ern-runner-gen-ios` package, we have to supply the mandatory property `reactNativeVersion` of `RunnerGeneratorConfig`.

PR #1395, opened 2019-11-06
Add proxy support to binary store client

belemaire:
Make binary store http client on par with bundle and source map store clients by adding configurable proxy support to it.

### DIFF
--- a/docs/cli/platform/config/del.md
+++ b/docs/cli/platform/config/del.md
@@ -16,6 +16,7 @@
 
 **Whitelisted properties**
 
+- `binaryStoreProxy`
 - `bundleStoreProxy`
 - `codePushAccessKey`
 - `codePushCustomHeaders`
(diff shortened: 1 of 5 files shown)


docs/cli/platform/config/del.md
```diff
@@ -17,6 +17,7 @@
 **Whitelisted properties**
 
 - `bundleStoreProxy`
+- `binaryStoreProxy`
```
friederbluemle: Minor: Want to keep this in alphabetical order too? Looks like the rest of the list is? :)

docs/cli/platform/config/get.md
```diff
@@ -17,6 +17,7 @@
 **Whitelisted properties**
 
 - `bundleStoreProxy`
+- `binaryStoreProxy`
```
friederbluemle: Same here.

docs/cli/platform/config/del.md
belemaire: You are right :) 
Actually that's what I thought I was doing but for some reason at the time I did the change looks like my mind was using a different alphabet where `i` comes after `u` ;P
Good catch, thanks :+1:

PR #1403, opened 2019-11-12
Fix api generation system tests

belemaire:
Follow up to https://github.com/electrode-io/electrode-native/pull/1400

The system tests are unfortunately still failing :( This PR fixes the remaining issues.

System test run locally ✅

### DIFF
--- a/system-tests/tests/misc.js
+++ b/system-tests/tests/misc.js
@@ -41,7 +41,7 @@ run('compat-check', { expectedExitCode: 1 })
 
 // Miniapp commands
 run(`create-miniapp ${f.systemTestMiniAppName} --packageName ${f.systemTestMiniAppPkgName} --language JavaScript --packageManager yarn --skipNpmCheck`)
-const miniAppPath = path.join(process.cwd(), f.systemTestMiniAppPkgName)
+const miniAppPath = path.join(process.cwd(), f.systemTestMiniAppName)
 console.log(info(`Entering ${miniAppPath}`))
 process.chdir(miniAppPath)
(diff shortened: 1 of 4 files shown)


system-tests/tests/misc.js
```diff
@@ -41,7 +41,7 @@ run('compat-check', { expectedExitCode: 1 })
 
 // Miniapp commands
 run(`create-miniapp ${f.systemTestMiniAppName} --packageName ${f.systemTestMiniAppPkgName} --language JavaScript --packageManager yarn --skipNpmCheck`)
-const miniAppPath = path.join(process.cwd(), f.systemTestMiniAppPkgName)
+const miniAppPath = path.join(process.cwd(), f.systemTestMiniAppName)
```
friederbluemle: Shouldn't this remain `systemTestMiniAppPkgName`?

belemaire: It fails the test :(
The command that is run on the line above is 
`ern create-miniapp MiniAppSystemTest --packageName miniapp-system-test --language JavaScript --packageManager yarn --skipNpmCheck`
This command end up creating the miniapp in `MiniAppSystemTest` directory, not `miniapp-system-test` directory, so the path to the miniapp is wrong and test fail to locate directory.
If the command should  be generated in `miniapp-system-test` instead it means that's a bug with the command that should be fixed, but with current code base state on master, this test is failing :(

friederbluemle: Oh wow.. Okay, yeah, then we need other fixes as well. We can merge this first and create a separate PR. I'll investigate. Thanks!

PR #1414, opened 2019-11-22
Fix system tests

belemaire:
Just happened to notice -a tad late- that system tests got broken after the update of react-native version to 0.60.6 in Manifest for dev platform.

This PR just regenerates the fixtures and get rids of two tests that don't add much but are a pain to maintain.

It also can be noticed that package.json generation of api impls seems to have been broken somehow (still functional but not nicely formatted) I'll open an issue for that.

On another note ... still have to figure out why email notification is not sent when ST nightly run is failing :(

### DIFF
--- a/system-tests/fixtures/api-impl-js/ern-movie-api-impl/package.json
+++ b/system-tests/fixtures/api-impl-js/ern-movie-api-impl/package.json
@@ -4,7 +4,7 @@
   "main": "index.js",
   "license": "MIT",
   "devDependencies": {
-    "react-native": "0.60.5"
+    "react-native": "0.60.6"
   },
   "dependencies": {
     "react-native-ernmovie-api": "0.0.11"
--- a/system-tests/fixtures/api/complex-api/package.json
+++ b/system-tests/fixtures/api/complex-api/package.json
@@ -1,5 +1,5 @@
 {
-  "author": "f0b00n7",
+  "author": "blemair",
   "dependencies": {
     "react-native-electrode-bridge": "1.5.x"
   },
--- a/system-tests/fixtures/api-impl-js/ern-movie-api-impl/yarn.lock
+++ b/system-tests/fixtures/api-impl-js/ern-movie-api-impl/yarn.lock
@@ -10,17 +10,17 @@
     "@babel/highlight" "^7.0.0"
 
 "@babel/core@^7.0.0":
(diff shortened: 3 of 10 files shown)


system-tests/fixtures/api-impl-js/ern-movie-api-impl/package.json
```diff
@@ -1,25 +1 @@
...
-        "Movies"
-      ]
-    },
-    "moduleType": "ern-js-api-impl"
-  },
-  "keywords": [
-    "ern-js-api-impl"
-  ]
-}
+{"name":"ern-movie-api-impl","version":"1.0.0","main":"index.js","license":"MIT","devDependencies":{"react-native":"0.60.6"},"dependencies":{"react-native-ernmovie-api":"0.0.11"},"ern":{"containerGen":{"hasConfig":false,"moduleName":"ReactNativeErnmovieApiImplJs","apiNames":["Movies"]},"moduleType":"ern-js-api-impl"},"keywords":["ern-js-api-impl"]}
```
friederbluemle: Wait .. Is this expected?!

belemaire: No ;) As mentioned in my PR comment, package.json generation of api impls seems to have been broken somehow in a previous merge. Will open a separate issue for that.

friederbluemle: Oops sorry, I did read the description but somehow missed that part ;)

system-tests/fixtures/api/complex-api/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "author": "f0b00n7",
+  "author": "blemair",
```
friederbluemle: It's kinda annoying that this file gets touched and this field updated based on the local username of whoever runs the script. The author/committer ID is already tracked in source control, there should be no need to have this _inside_ of `package.json`.

Suggestion:
Update the two functions that call `ern create-api` in `regen-fixtures.js` to also include a `-u "generated"` flag.

belemaire: Agreed, that's a good idea :+1:

friederbluemle: #1418

PR #1418, opened 2019-11-25
Avoid unnecessary modifications to test fixtures

friederbluemle:
No longer take the machine-local user name and modify the `author` field in the test fixtures on regeneration.


friederbluemle:
Aah perfect, makes sense @belemaire - I'll update these file too! :+1:

PR #1419, opened 2019-11-26
Generate hermes bundle if hermes is enabled

belemaire:
Prior this PR, if Hermes engine was enabled (through container generator configuration), Electrode Native was only doing what was needed for the Container library to use Hermes JavaScript engine instead of JavaScriptCore. However, it was not creating a bytecode bundle, but still only generating / using a standard JS minified bundle using metro.

This PR updates the generation logic, for code paths that are generating a bundle, specifically
 - Container generation
- CodePush release

so that if Hermes is enabled, Electrode Native will compile the JS bundle to an Hermes bytecode bundle and use this new bundle rather than the JS one (inside Android Container or for a CodePush release).


ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -34,9 +36,9 @@ import semver from 'semver'
 const PATH_TO_TEMPLATES_DIR = path.join(__dirname, 'templates')
 const PATH_TO_HULL_DIR = path.join(__dirname, 'hull')
 
-const DEFAULT_JSC_VARIANT = 'android-jsc'
-const DEFAULT_JSC_VERSION = '245459'
-const DEFAULT_HERMES_VERSION = '0.2.1'
+export const DEFAULT_JSC_VARIANT = 'android-jsc'
+export const DEFAULT_JSC_VERSION = '245459'
+export const DEFAULT_HERMES_VERSION = '0.2.1'
```
krunalsshah: can we please move these constants to `electrode-native/ern-core/src/android.ts`

belemaire: :+1:

PR #1420, opened 2019-11-26
Fix ern-api-gen template file formatting

friederbluemle:
Part of #1511

This is major cleanup/partial rewrite of all generator template files of `ern-api-gen`.

All mustache control/conditional markers have been moved to column 1: They should be considered _meta symbols_ that dynamically assemble _lines_ of actual source code. Aside from the minor benefit of saving many leading spaces, this makes it easier to see which lines will part of the resulting generated source file (and which are control statements). It also reduces the chance of accidentally introducing whitespace errors (trailing spaces), or blank lines.

The generated code now (mostly) passes various linters and conforms to the default style of IDE formatters. This is a prerequisite for the eventual default integration of linters and testing tools.

The source code has been processed by the following tools and programs (in order from top to bottom):

## Android

- `google-java-format` (AOSP setting, skip sorting imports)
- Android Studio 3.6.2 "Reformat code" (including "Optimize imports", "Rearrange entries", "Cleanup code")

## JavaScript

- ESLint and Prettier (with `@react-native-community/eslint-config`)
- `prettier-eslint`
- WebStorm 2019.3 (default settings, including "Optimize imports", "Rearrange entries", "Cleanup code")

## Swift

- Xcode 11.4 Re-indent (Ctrl+I)
- SwiftFormat
- SwiftLint

(Setting for both SwiftFormat and SwiftLint: Swift 3.0)

This also fixes the **field order** of the generated `package.json` file. Fields are typically _not_ in alphabetical order, but start with `name`, `version`, etc.

Generated projects are **fully whitespace error-free** (they pass the test against the empty tree object, as well as the `trailing-whitespace`/`end-of-file-fixer` pre-commit checks)


belemaire:
Wow ! That's great, thanks 👍 
Given the number of files changed it'll take me a little bit of time to properly review. 
I'll have a look at it later on today, mostly focusing on the diffs of what gets generated (system tests fixtures).

friederbluemle:
Sure, thanks @belemaire 👌 Yes, I'd also suggest to mostly focus on the generated files, the templates are a bit harder to "read". ;) I've also asked Deepu and will check with Lianci to have a look.

belemaire:
Also one more thing just thought of that might be problematic.
From what I know, compared to most other languages, Swift does not really care of introducing breaking changes to the language / maintain forward/backward compatibility :(
This is why we have some such directives ```#if swift(>=4.0)``` in the generated files. I am not sure what exactly the formatter/linter are doing (but can see some `self` removed, `return` removed, some Array<any> converted to [any] ... so it's def altering the code itself), but because they target a specific version of Swift, I hope this won't introduce issues. For example I can see you setup Swift target to 3.0. What about all of our generated code in ```#if swift(>=4.0)```. 
It's Swift 4 compatible code. Because we set target to Swift 3, I fear that the formatter/linter might detect something is not valid (in Swift3 that is) and turn it back to Swift3 constructs that could break with Swift4 .
Not sure ... need a deeper dive into what these guys are doing, but I think there might be a risk here. If there is, only solution would be to have separate directories for different Swift versions for our generators so that we can apply the tools at different swift versions on these directories, rather than keeping the swift language versions switches inline in the same source file, because I don't think these tools are "smart" enough to detect macro directives for language version and not alter the code within blocks of a different swift version than target.
Doesn't impact my approval, but @liulianci please double check generated code diff in `system-tests/fixtures/*` for swift files, see if anything looks wrong.

friederbluemle:
Thanks again for the super detailed review and comments @belemaire :+1:
What you said definitely makes sense. I also thought about the Swift thing, but it's good that you point it out again here.
When I ran the linter, **even when set to Swift 1.0**, `swiftformat` would perform the same changes (`self`, `return` removed etc.). Only when set to Swift 5.1 it would do a few additional cleanups (which of course are _not_ included here, as I assume they're not compatible).

I'm hoping at some point we can remove things like `#if swift(>=4.0)` since there are very few differences anyway, but I don't have enough information/background of why they're there in the first place, so I didn't touch them. Also I'm not sure if/where we "set target to Swift 3".

I think as a precaution, I may re-add the removed `self` qualifiers and `return` statements.

Leaving this PR open for now to give @liulianci a chance to take a look as well.

A couple of questions for @liulianci:
- Which Swift version(s) do we "officially" support?
- Is there a way to test compile the source with different versions (in Xcode/on command line)?

belemaire:
Following up on this one. 

I think we should get rid of Swift 3 support. 
I doubt any internal project is still on Swift 3, nor external. We can list that as a breaking change for API generation in our release notes (i.e, api generation now only support Swift >= 4.0 ... if you need to regenerate some APIs specifically for Swift 3, you can use ern <= 0.39).
This will allow us to greatly cleanup the swift generated code by getting rid of big chucks of code only there for Swift 3 (for ex [this one](https://github.com/electrode-io/electrode-native/blob/master/system-tests/fixtures/api/complex-api/IOS/SystemTestsAPI.swift#L47-L94) ... but same in all generated swift files, half of the code is to support swift 3.

Thoughts on this ?

friederbluemle:
Oh I would love to get rid of those 95% duplicated code chunks. :+1: Since I don't have enough experience/knowledge with Swift versions I left the code as is, but also thought about it. Thanks for bringing it up!

@liulianci @jw2175 - Are there any statistics regarding Swift version usage for React Native (or more specifically ERN) projects?

Maybe somewhat related: Yesterday I noticed a commit was pushed to the react-native master branch that sets the Swift version in the default iOS template to **5.0** (and now also _always_ bundles the Swift standard libraries!). Not 100% sure how/if this will impact us, and how intentional the change was in the first place (it was related to the Flipper integration).

belemaire:
Our default iOS template right now for Container is setup to use Swift 4.0
https://github.com/electrode-io/electrode-native/blob/2141867bf5ae10dc5ef09b8387e14ef63b116ae7/ern-container-gen-ios/src/hull/Config/ElectrodeContainer-Debug.xcconfig#L29

This can always be overridden by clients generating containers, through configuration.
As we do for Android versions, we should ideally, in every ern release, update the default target to latest version available if our generated code is compatible with it. As said before, configuration can override swift target version for container, so if our default target is swift 5.1 but some client need a build of a container against swift 4.0 it can be configured for that (as long as our generated API and container code is still compatible with 4.0 that is).

In any case, Swift 4.0 was released in September 2017, so more than 2 years ago. We should really get rid of the Swift 3 code generation in our templates, this will make code leaner / more readable and also we won't have to maintain both Swift 3 / Swift >= 4 sections. 

Now regarding React Native itself, this will be a separate discussion. RN61 full move to CocoaPods already makes it impossible to generate containers with our current generator (apart from what you just mentioned). We will need some major redesign of iOS Container generator to support this new version and also see how this will impact container client apps.

friederbluemle:
Great, thanks for the additional insights! 👌 

A quick Git log check additionally reveals that `ElectrodeContainer-Debug.xcconfig` **never** had anything other than `Swift 4.0` set after it was first added in July 2018. So as mentioned, it is very unlikely that any current project is using something lower than Swift 4.0.

For the removal, would you prefer a separate commit for the removal or the Swift < 4.0 blocks, or squash it into the existing `Fix ern-api-gen template file formatting` commit? Keeping it separate would make it possible to selectively revert, in the extremely unlikely case it becomes necessary.

ern-api-gen/resources/swift/libraries/ern/api.mustache
```diff
@@ -1,184 +1,171 @@
...
-    public func remove{{camelizedNickName}}EventListener(uuid: UUID) -> ElectrodeBridgeEventListener? {
+    public func remove{{camelizedNickName}}EventListener(uuid _: UUID) -> ElectrodeBridgeEventListener? {
         assertionFailure("should override")
         return nil
     }
-    {{/isEvent}}
-    {{/operation}}
 
-    {{#operation}}
-    {{#isEvent}}
```
friederbluemle: @belemaire - It's a bit hard to see, but this file had three duplicated sections that were consolidated:

```
{{#operation}}
{{#isEvent}}
item 1
{{/isEvent}}
{{/operation}}

{{#operation}}
{{#isEvent}}
item 2
{{/isEvent}}
{{/operation}}

{{#operation}}
{{#isEvent}}
item 3
{{/isEvent}}
{{/operation}}
```

replaced with:

```
{{#operation}}
{{#isEvent}}
item 1

item 2

item 3
{{/isEvent}}
{{/operation}}
```

Not sure what the initial reason was, but does the change look OK to you?

belemaire: :( Yes this looks much better and will indeed produce same output

belemaire:
@friederbluemle OK let's remove swift<4 code blocks in a follow up PR, you'll have to also re-run SwiftFormat / SwiftLint for 4.0 Swift target on this upcoming PR after doing the changes. 
I think we can merge this current one as is and make a more in depth review on the next one which will have a leaner Swift code surface making it easier to review / try out.

friederbluemle:
Sounds good. I already have the commit ready locally, but will push it to a separate branch/separate PR.

PR #1425, opened 2019-12-03
Remove JS APIs impls special handling

belemaire:
Remove JS APIs implementations special handling in Cauldron/Container.

Keeping JS API implementation packages separate from MiniApps proved to be of no real use and introduced additional complexity in the codebase as well as to the commands and overall platform understanding. 

Even though the concept of keeping JS API implementations orphaned (not being depended upon by any MiniApp) made sense at the time, in practice it's not been used, as MiniApps are directly depending on the JS API implementations. Especially since we have removed support for orphaned standalone native API implementations, it does not make a lot of sense to keep this feature for JS API implementations.

This is a breaking change :boom:

It removes the following commands:

- `cauldron add jsapiimpls`
- `cauldron del jsapiimpls`
- `cauldron update jsapiimpls`

It also removes the `--jsApiImpl` option from the following commands:

- `code-push release`
- `create-composite`
- `start`

And removes the `--miniAppsOnly` / `--jsApiImplsOnly` flags from the following commands:

- `github align-dependencies`
- `github create-ref`
- `github delete-ref`

Projects using orphaned JS API implementations, should now move the JS API implementation(s) as a direct dependency of one or more MiniApps that are using these API implementations, or move them to a base composite project or an 'umbrella' MiniApp that only contains JS API implementations / native API implementations dependencies.


deepueg:
@belemaire This will now force the dependency to be added to anyone of the MiniApps to make it to the container. If someone wants to write a JS implementation for a utility/business logic for a native app/module to consume, then this has to be added to one of the MinIApps even if the MiniApp does not need it.

belemaire:
Closing this one for now. 
May reopen later on

PR #1429, opened 2019-12-04
Update containsVersionsMismatch function to support prerelease versions

belemaire:
Update `containsVersionsMismatch` function to support pre-release packages versions. 
Refactor the function to make it more readable/understandable.

Before this change, the function was throwing an error in case the versions array was containing different pre-release versions.

The function will treat different pre-release versions as mismatching, whichever mismatch level is passed to the function (similar to `patch` mismatch level which will return true if patch/minor or major is different). There is no widely adopted convention for pre-release package versioning, and a different pre-release id might contain a breaking change.

i.e:

```
[ '1.0.0-foo.1', '1.0.0-foo.2' ] => true
[ '1.0.0', '1.0.0-foo.1' ]       => true
['1.0.0-foo.1', ''1.0.0-foo.1' ] => false
```

Unit tests for this function have been updated accordingly, and refactored by the same occasion to reduce verbosity and enhance readability / maintainability.

### DIFF
--- a/ern-core/src/resolveNativeDependenciesVersions.ts
+++ b/ern-core/src/resolveNativeDependenciesVersions.ts
@@ -7,17 +7,18 @@ export function containsVersionMismatch(
   versions: string[],
   mismatchLevel: 'major' | 'minor' | 'patch'
 ): boolean {
-  const minVersion = semver.minSatisfying(versions, '*')
-  const maxVersion = semver.maxSatisfying(versions, '*')
-  const majorMismatch = semver.major(maxVersion) !== semver.major(minVersion)
-  const minorMismatch = semver.minor(maxVersion) !== semver.minor(minVersion)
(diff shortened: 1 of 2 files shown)


friederbluemle:
Great, thanks @belemaire - With this, the custom versions of react-native-webview will be properly supported as well I assume?

belemaire:
pre-release versions are already supported. The problem here, addressed by this PR is only in case different miniapps are using different pre-release versions of the same native module (for ex foo@1.0.0-alpha.1 and foo@1.0.1-alpha.2  ... or even foo@1.0.0-alpha.1 and foo@1.0.0-alpha.2), which should fail anyway but with a clearer error message (thrown later on in the code path, but not reached because this function was not handling pre-release versions properly, leading to an obscure exception being thrown from inside the semver lib).

PR #1430, opened 2019-12-04
Align jetpack libraries to the latest stable version

deepueg:
This updates the jetpack Libraries version to the latest stable release. 

https://developer.android.com/jetpack/androidx/versions


friederbluemle:
Do we need to regenerate any fixtures?

deepueg:
Don't think so.!

belemaire:
@deepueg Looks like @friederbluemle was right and there was some system test fixture to regen  ;)
https://github.com/electrode-io/electrode-native/blob/master/system-tests/fixtures/api-impl-native/ern-movie-api-impl/android/lib/build.gradle#L31-L32
Could you please regen `native api impl` fixtures and open a PR ?

PR #1431, opened 2019-12-05
Regenerate native api impl system test fixture

belemaire:
Missing in https://github.com/electrode-io/electrode-native/pull/1430

### DIFF
--- a/system-tests/fixtures/api-impl-native/ern-movie-api-impl/android/lib/build.gradle
+++ b/system-tests/fixtures/api-impl-native/ern-movie-api-impl/android/lib/build.gradle
@@ -28,8 +28,8 @@ android {
 dependencies {
     implementation fileTree(dir: 'libs', include: ['*.jar'])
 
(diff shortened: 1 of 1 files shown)


friederbluemle:
@deepueg - Can you confirm as well please? You mentioned this might need to be reverted, but not sure if still valid?

belemaire:
Oh OK, sorry didn't know it was planned to be potentially reverted :(
Is revert intended because moving to these new versions not needed right now or because it introduces some issue(s) ? If it's the former, we should not have to revert, will check with @deepueg.
Merging now anyway because nightly STs run will be triggered at 8PM PST, just want to make sure they don't fail just because of this (or for anyone running STs locally from master as I just did).
In case we need to revert, we will just have to revert two PRs instead of one, not a big deal ;)
At least  UTs and STs are working on master through this one.

friederbluemle:
No problem. In case it needs to be reverted, it's pretty simple and contained anyway. 👍

deepueg:
No, it does not need to be reverted. We are good.

PR #1432, opened 2019-12-05
Bump typescript

belemaire:
Bump TypeScript to `3.7.3` version (latest), along with bumping all TypeScript associated dev dependencies to their latest, for proper [TypeScript 3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html) support.

Please note that the only dev dependency to get a major bump is `pretty-quick`, going from `1.x` to `2.x`. [Only breaking change in 2.x](https://github.com/azz/pretty-quick/releases/tag/v2.0.0) is a strong requirement on Node 8, which is already our minimal Node version requirement, so no issue there.

This PR also contains an update to the vscode workspace settings to use the version of TypeScript installed in the workspace, rather than the one shipped in vscode. This is achieved through [typescript.tsdk](https://code.visualstudio.com/docs/typescript/typescript-compiling#_using-the-workspace-version-of-typescript) setting. This is needed because current stable release of vscode (October 2019) is still shipping with TypeScript 3.6 (next vscode release will ship with 3.7).

Finally, this PR contains some code refactoring, to make some statements less verbose, using new language constructs introduced in TypeScript 3.7, respectively [Optional Chaining](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#optional-chaining) and [Nullish Coalescing](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#nullish-coalescing). Did a quick pass through the code base, probably missed a lot of LOC where these constructs could be used, but will convert any other we notice, over time.

### DIFF
--- a/ern-container-gen/src/copyRnpmAssets.ts
+++ b/ern-container-gen/src/copyRnpmAssets.ts
@@ -24,7 +24,7 @@ export function copyRnpmAssets(
 
   for (const miniAppPath of miniAppPaths) {
     const packageJson = readPackageJsonSync(miniAppPath)
-    if (packageJson.rnpm && packageJson.rnpm.assets) {
+    if (packageJson.rnpm?.assets) {
       for (const assetDirectoryName of packageJson.rnpm.assets) {
         const source = path.join(assetDirectoryName, '*')
         const dest = getAssetsPath(platform, assetDirectoryName)
--- a/ern-runner-gen-android/src/AndroidRunnerGenerator.ts
+++ b/ern-runner-gen-android/src/AndroidRunnerGenerator.ts
@@ -75,13 +75,7 @@ function getPackageFilePath(config: RunnerGeneratorConfig): string {
 }
 
 function isOldRunner(config: RunnerGeneratorConfig): boolean {
-  const oldRunner =
-    config.extra &&
-    config.extra.androidConfig &&
-    config.extra.androidConfig.isOldRunner
-      ? true
-      : false
-  return oldRunner
+  return !!config.extra?.androidConfig?.isOldRunner
 }
 
 // Given a string returns the same string with its first letter capitalized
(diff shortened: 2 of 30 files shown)


ern-container-gen/src/copyRnpmAssets.ts
```diff
@@ -24,7 +24,7 @@ export function copyRnpmAssets(
 
   for (const miniAppPath of miniAppPaths) {
     const packageJson = readPackageJsonSync(miniAppPath)
-    if (packageJson.rnpm && packageJson.rnpm.assets) {
+    if (packageJson?.rnpm?.assets) {
```
friederbluemle: So `packageJson` can be null? Previously it was a potential bug?

ern-runner-gen-android/src/AndroidRunnerGenerator.ts
```diff
@@ -75,13 +75,7 @@ function getPackageFilePath(config: RunnerGeneratorConfig): string {
...
 
 function isOldRunner(config: RunnerGeneratorConfig): boolean {
-  const oldRunner =
-    config.extra &&
-    config.extra.androidConfig &&
-    config.extra.androidConfig.isOldRunner
-      ? true
-      : false
-  return oldRunner
+  return !!config.extra?.androidConfig?.isOldRunner
```
friederbluemle: Sorry JavaScript noob here, still getting used to this syntax.. Can you quickly explain what this does? ;) `!!` turns null into boolean?

belemaire: Yes you are right :) 
I agree it looks weird, but it's indeed to reduce an expression to a boolean in a less verbose way than 
```javascript
config.extra.androidConfig.isOldRunner	
      ? true	
      : false
```
for example if `config.extra.androidConfig.isOldRunner` evaluates to `null` or `undefined` or `0` or `''`, JS consider these values as "falsy", but they are not of boolean type (and the function signature says it'll return a boolean).
`!!` is just a double negation to make sure that we evaluate the expression to a boolean in any case (it also works if the expression is already evaluating to a boolean).
`!!false` => `!(!false)` => `!(true)` => `false`
`!!0` => `!(!0)` => `!(true)` => `false`

ern-container-gen/src/copyRnpmAssets.ts
belemaire: No you are right, it shouldn't be `packageJson?`, will update it.

ern-runner-gen-android/src/AndroidRunnerGenerator.ts
friederbluemle: Got it. Thanks for explaining!

PR #1452, opened 2020-01-10
add ERNDelegate to listen for react native initialization

jw2175:


ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
```diff
@@ -88,7 +92,7 @@ NS_ASSUME_NONNULL_BEGIN
  multiple plugins.
  */
 
-+ (void)startWithConfigurations:(id<ElectrodePluginConfig>)reactContainerConfig
++ (void)startWithConfigurations:(id<ElectrodePluginConfig>)reactContainerConfig ernDelegate: (id<ERNDelegate>) ernDelegate
```
deepueg: This would result in a breaking change. Can we deprecate the old method and add a new one instead ?

jw2175: Yes, thank you.

krunalsshah:
@jw2175 thanks for your contribution.
- Sign the CLA
- System test will fail, we'd have to regen the fixtures for iOS
- We can apply the changes to the runner hull

deepueg:
Can you please squash these commits?

friederbluemle:
Fixture regeneration should be a _separate commit_. Let me know if you prefer I separate PR. I can quickly take care of that.

friederbluemle:
There are **tons** of formatting changes completely unrelated to `add ERNDelegate to listen for react native initialization` in that commit.

system-tests/fixtures/ios-container/ElectrodeContainer/ElectrodeReactNative.m
```diff
@@ -128,13 +129,24 @@ @implementation ElectrodeReactNative
 #pragma mark - Public Methods
 
 + (void)startWithConfigurations:(id<ElectrodePluginConfig>)reactContainerConfig
+{
+    id sharedInstance = [ElectrodeReactNative sharedInstance];
+    static dispatch_once_t startOnceToken;
+    semaphore = dispatch_semaphore_create(0);
+    dispatch_once(&startOnceToken, ^{
+        [sharedInstance startContainerWithConfiguration:reactContainerConfig ernDelegate:nil
+];
```
friederbluemle: Not sure about the formatting on this line, can you double check and confirm @jw2175

jw2175: Yeah, it's not pretty but works. I believe the previous version has similar formatting when there are no configs set. Not too familiar with the code generation portion to make the brackets on the same line.

PR #1457, opened 2020-01-13
Regenerate fixtures

friederbluemle:
Regen fixtures on `master` with a newer version of `react-native-electrode-bridge` (which included updated templates).


friederbluemle:
Should be merged before #1452

PR #1463, opened 2020-01-21
update the yarn.lock to fix security warnings

krunalsshah:
`handlebars`
Versions of handlebars prior to 4.3.0 are vulnerable to Prototype Pollution leading to Remote Code Execution.
Transitive dependency enforced by `istanbul@^0.4.5`

`mem`
In nodejs-mem before version 4.0.0 there is a memory leak due to old results not being removed from the cache despite reaching maxAge. The exploitation of this can lead to exhaustion of memory and subsequent denial of service.
Transitive dependency enforced by `yargs@^11.0.0`

There is a

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,101 +2,95 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.5.5":
-  version "7.5.5"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.5.5.tgz#bc0782f6d69f7b7d49531219699b988f669a8f9d"
-  integrity sha512-27d4lZoomVyo51VegxI20xZPuSHusqbQag/ztrBC7wegWoQ1nLREPVSKSW8byhTlzTKyNE4ifaTA6lCp7JjpFw==
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.8.3":
+  version "7.8.3"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.8.3.tgz#33e25903d7481181534e12ec0a25f16b6fcf419e"
+  integrity sha512-a9gxpmdXtZEInkCSHUJDLHZVBgb1QS0jhss4cPP93EW7s+uC5bikET2twEF3KV+7rDblJcmNvTR7VJejqd2C2g==
   dependencies:
-    "@babel/highlight" "^7.0.0"
+    "@babel/highlight" "^7.8.3"
 
-"@babel/generator@^7.4.0", "@babel/generator@^7.6.0":
-  version "7.6.0"
-  resolved "https://registry.yarnpkg.com/@babel/generator/-/generator-7.6.0.tgz#e2c21efbfd3293ad819a2359b448f002bfdfda56"
-  integrity sha512-Ms8Mo7YBdMMn1BYuNtKuP/z0TgEIhbcyB8HVR6PPNYp4P61lMsABiS4A3VG1qznjXVCf3r+fVHhm4efTYVsySA==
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -4428,21 +4091,21 @@ gradle-to-js@^2.0.0:
...
 growl@1.10.5:
   version "1.10.5"
   resolved "https://registry.yarnpkg.com/growl/-/growl-1.10.5.tgz#f2735dc2283674fa67478b10181059355c369e5e"
   integrity sha512-qBr4OuELkhPenW6goKVXiv47US3clb3/IbuWF9KNKEijAy9oeHxU9IgzjvJhHkUzhaj7rOUD7+YGWqUjLp5oSA==
 
-handlebars@^4.0.1, handlebars@^4.1.2:
-  version "4.2.0"
-  resolved "https://registry.yarnpkg.com/handlebars/-/handlebars-4.2.0.tgz#57ce8d2175b9bbb3d8b3cf3e4217b1aec8ddcb2e"
-  integrity sha512-Kb4xn5Qh1cxAKvQnzNWZ512DhABzyFNmsaJf3OAkWNa4NkaqWcNI8Tao8Tasi0/F4JD9oyG0YxuFyvyR57d+Gw==
+handlebars@^4.0.1, handlebars@^4.4.0:
```
krunalsshah: recommended version

yarn.lock
```diff
@@ -5987,12 +5751,14 @@ meant@~1.0.1:
   resolved "https://registry.yarnpkg.com/meant/-/meant-1.0.1.tgz#66044fea2f23230ec806fb515efea29c44d2115d"
   integrity sha512-UakVLFjKkbbUwNWJ2frVLnnAtbb7D7DsloxRd3s/gDpI8rdv8W5Hp3NaDb+POBI1fQdeussER6NB8vpcRURvlg==
 
-mem@^1.1.0:
-  version "1.1.0"
-  resolved "https://registry.yarnpkg.com/mem/-/mem-1.1.0.tgz#5edd52b485ca1d900fe64895505399a0dfa45f76"
-  integrity sha1-Xt1StIXKHZAP5kiVUFOZoN+kX3Y=
+mem@^4.0.0:
```
krunalsshah: recommended version

friederbluemle:
Thanks @krunalsshah - As mentioned in the standup last week, I'm already working on _removing_ the dependency on the deprecated `istanbul` library (which was last updated 3 years ago!). Do you think you can rebase your changes once that's migrated to `nyc`?

krunalsshah:
@friederbluemle 

If there is already a plan to migrate the project to the newer test coverage library, I can see the remediated version is taken care https://github.com/istanbuljs/nyc/blob/v15.0.0/package-lock.json#L2716

In which case it's okay to close the PR. The motivation is to remediate the message `We found potential security vulnerabilities in your dependencies.`

It's been up for quite some time :)

friederbluemle:
Yeah I know it's ben up for quite some time :/
Just leave this PR open for now (maybe marked as `do-not-merge`), and we'll see if any of the other updated (`mem`?) will remain after `istanbul` removal.

I've had some issues running the coverage locally, will check with you directly

belemaire:
Sorry for late answer, but sadly I think you can close this PR anyway :(
`yarn.lock` should never be updated manually. 
It should only get updated by yarn itself as a side effect based on the dependencies versions that are resolved at the time of yarn install, following a change of dependencies in our modules package.json (adding/removing/upgrading dependencies).
If the problem is due to istanbul, this should be fixed by replacing istanbul with nyc. Unfortunately I gave it a shot a while back, spent some time on it, and this is not as straightforward as it should be, and requires much more digging. 
I know an issue is opened for this replacement, will try to dig into it and get done with it when I have more time.

krunalsshah:
@belemaire I will close the PR. The `yarn.lock` wasn't updated manually, the outcome is result of `yarn install` after rebasing the master of the project.

PR #1441, opened 2020-01-03
fix(ern-api-gen): patch bad object ref, improve error messaging, patch $ref bugs

cdaringe:
# problem

on generating an api, i encountered this:

```
$ ern regen-api
[v0.38.7] [Cauldron: -NONE-]
✔ Syncing https://github.com/electrode-io/electrode-native-manifest.git Manifest 1s
? Would you like to bump the plugin version from [react-native-walmartsmartlist-api@1.8
✔ Syncing https://github.com/electrode-io/electrode-native-manifest.git Manifest 1s
✖ An error occurred: Could not resolve type undefined undefined
```

as you can see, the error message is invalid and not actionable.

additionally,

- regen-api was attempting to set an undefined value if license was omitted
- regen-api was _not_ setting `additionalProperties` appropriately
- regen-api was not handling $refs in additionalProperties

# solution

- fix the invalid `parent` reference passed into the function stack that induces the reported error
- enhance the error messaging to direct the user more closely to his/her failing code
- test it.
- set `additionalProperties`
- do not set an undefined license value
- permit $ref processing in additionalProperties

# discussion

- now the user gets meaningful, actionable information on error:

image

- the regen-fixtures-check may need some re-wiring:
```
$ node regen-fixtures-check
fatal: ambiguous argument 'ORIG_HEAD': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
child_process.js:642
    throw err;
    ^

Error: Command failed: git diff-tree -r --name-only --no-commit-id ORIG_HEAD HEAD
fatal: ambiguous argument 'ORIG_HEAD': unknown revision or path not in the working tree.
```

PR #1471, opened 2020-02-01
Improvements and fixes to yarn info method

belemaire:
Various fixes / improvements to `YarnCli:info` method :

- Remove `json` option from the method given that all clients calls are currently setting this option to `true`. It simplifies call sites and method implementation. I don't see an internal use for a call without `json` option anyway.

- Improve the method by returning the requested data object rather than the top level yarn response object, simplifying response handling by caller.

- Improve/Fix the method by letting any retries go through and only returning a success when we got a proper response from `yarn info`. This avoid the return of an improper response on first retry, leading to deserialization errors such as `Unexpected token { in JSON at position 95`. With this fix retries will go through properly rather than throwing this obscure error on first retry.

- Fix method implementation for local package, which was not properly considering the `field` option if it was provided to the method.

*Will squash all commits upon approval, prior to merge.*


ern-api-impl-gen/src/generators/ApiImplGen.ts
```diff
@@ -63,9 +63,7 @@ export default class ApiImplGen {
               apiPackagePath,
               PackagePath.fromString(`react-native@${reactNativeVersion}`),
               PackagePath.fromString(
-                `react-native-electrode-bridge@${
-                  reactNativeElectrodeBridge.version
-                }`
+                `react-native-electrode-bridge@${reactNativeElectrodeBridge.version}`
```
friederbluemle: Looks like a formatting-only change.. Just friendly request to use a separate commit next time for easier review ;)

belemaire: Totally agreed and would make the same remark to anyone ;)

Not my intention though, it slipped through it seems ... looks like `prettier` was applied on this file as I have enabled `format on save` in vs-code. Even if I had not enabled it, it would have updated the file via a pre-commit hook (which is running `prettier` on all staged files, in case contributors are not using vs-code or format on save).

That being said, I had already noticed this behavior but never looked into it more closely ... just took a closer look and looks like we should run prettier on all source files whenever we are bumping the version of it, has it comes with some formatting rules updates. I think this is what happened here and in the past when I noticed this behavior. Because `prettier` is only run on any saved files (with proper vs-code config) or any git staged files as pre-commit hook ... if we bump prettier version and it comes with new formatting rules, then this new formatting rules will be only applied when the file will be touched (either as save or git staged). 

So rule of thumb, we should run prettier on all source files whenever we bump the version of it (should be done in same PR as the one containing bump to prettier so that we can see all files for which formatting was changed by new version of it).

For now, I'll just run `prettier` on all files and open a PR with changed files. This way at least this kind of issue won't happen on any future PR, at least until next version bump of prettier (then we have to keep in mind that we should re-run prettier post version bump of it... and can think in // of a way to automate this so that we don't have to keep it in mind at all ;P)

friederbluemle: Sounds good, and I agree - Updates to the formatter that come with change to how files are formatted need to go in the _same commit_. Otherwise, there is no other way to avoid inconsistent state. I have a feeling that a different version of Prettier also caused the unintended formatting changes in #1441 a few weeks ago. Overall, this project is already in a pretty good state (compared to many others), but we can make it even better. Let's chat next week 👍

PR #1472, opened 2020-02-01
Prettier changes

belemaire:
Address what's been discussed in https://github.com/electrode-io/electrode-native/pull/1471#discussion_r373753927

- `prettier` task has been renamed to `prettier:staged`. This task is only of use for the pre-commit hook defined in `.huskyrc`.
- New `prettier:check` task that can be run to check if all non prettier ignored source files of codebase are properly formatted according to current prettier rules.
- New `prettier:write` task that can be run to format in place, all non prettier ignored source files of codebase. This task should be manually run whenever we bump prettier version and any files updated through running this task should be included along with the PR that is bumping prettier version.
- All source files modified in this PR are the result of running `prettier:write` task manually. This way all current source files are properly matching prettier rules, so that we don't run into similar issue as the one noticed in https://github.com/electrode-io/electrode-native/pull/1471#discussion_r373753927


package.json
```diff
@@ -79,4 +81,4 @@
     "url": "https://github.com/electrode-io/electrode-native.git"
   },
   "homepage": "http://www.electrode.io"
-}
+}
```
friederbluemle: Did a tool remove the final newline here or what happened?

belemaire: No, my bad it would seem. fixed.
prettier is currently configured to ignore json files, otherwise it would have not happened.
don't remember the details as to why we added json files to the exclusion list, but should look into it and open another PR to address this (i.e remove json files from exclusion). Looks like it was done as a quick workaround to some kind of issue (maybe some tests doing exact file content comparison and produced json content is not using prettier rules making test fails if fixture is prettified) but if that's the case we should rather ignore specific json files or directories rather than excluding them all (as first step, and then maybe, if that's the case, apply same prettier rules to generated json files). Will open issue for this based on this comment.

PR #1479, opened 2020-02-07
Fix hermes source map generation

belemaire:
Missed a crucial step in bundle source map generation when using Hermes engine.
The non-hermes JS bundle source map ('packager' one) needs to be combined with the Hermes one ('compiler' one) to obtain the final usable bundle source map, by running a react-native script.
This PR fixes this issue, so that we get proper source maps when using Hermes engine.


krunalsshah:
very nice 👍

PR #1480, opened 2020-02-07
Revert link/start improvements

belemaire:
This PR reverts some changes done on master, but not released yet in any version.
In an nutshell, it reverts these two PRs:
- https://github.com/electrode-io/electrode-native/pull/1369
- https://github.com/electrode-io/electrode-native/pull/1388

It also reverts a PR for a fix of `ern start` command, but reapplies it manually based on v0.38 codebase which has the fix for the 'old' (reverted) start command.

**Why ?**

The changes introduced by these two PRs have been made a while back and not released in any version yet. They contain major changes to link commands (breaking changes) as well as to start command. We are now aware of new ways to handle link and start flows, which will greatly improve performance of ern start command. That being said, the changes needed will probably invalidate most of the work done in these PRs, if not also probably change the way we expose link commands. Based on this, it does not make sense to release this breaking change to link/start, to potentially reintroduce a new breaking change soon after. We will work on improved link/start flow, from scratch based on our new ideas. It does not mean that this work was done in vain. We may manually get back some of the code/ideas from these PRs at some point.

This revert is mostly done so that we can release 0.39 version asap, which is long overdue (now causing cherry-picking complexities on 0.38 for patch releases), so that we can get back to a regular bi weekly release cycle of main line versions.


belemaire:
Yeah :(
afaik, first time ever (and hopefully the last) we had to use `git revert` on this repo. This is an historical day ;P

PR #1493, opened 2020-02-18
Speedup yarn info for local packages

belemaire:
Electrode Native is making use of `yarn info` for different reasons, one of them being to identify a module type, by looking at the `ern` property in package metadata. 

When generating containers that are including a lot of native dependencies packages, the overall cost of running a remote `yarn info` call for each package, can take minutes. In the case of container generation, considering that we first generate a composite, prior to injecting native dependencies in the container, all the native dependencies packages are going to be there locally (in `node_modules` of the composite). In that case, why incur the cost of a remote `yarn info` for each of these packages when we can just directly parse their `package.json` from disk ?

This is basically what this PR does. To achieve that, it was needed to expose the `name` of the package more directly in the case of file based packages (to avoid having to read it from the package.json every time). That's why there is quite a bit of refactoring, mostly adding a new `name` property to `PackagePath` class, having ripple effect throughout the code base (most of the time simplifying things). One problem with the `name` property though is that it'll be undefined for git based packages. Unfortunately it's difficult with git to get the name of package dynamically, unless we do a git clone (heavy) or a github raw url fetch or github api call (not that heavy but relies on having some token set in config and dependent on github provider).

### DIFF
--- a/ern-core/src/GitManifest.ts
+++ b/ern-core/src/GitManifest.ts
@@ -232,12 +232,12 @@ export default class GitManifest {
       let pluginScope
       let pluginName
       let basePluginPath
-      if (scopeNameRe.test(plugin.basePath)) {
-        pluginScope = scopeNameRe.exec(plugin.basePath)![1]
-        pluginName = scopeNameRe.exec(plugin.basePath)![2]
+      if (scopeNameRe.test(plugin.name!)) {
+        pluginScope = scopeNameRe.exec(plugin.name!)![1]
+        pluginName = scopeNameRe.exec(plugin.name!)![2]
         basePluginPath = path.join(pluginsConfigurationDirectory, pluginScope)
       } else {
-        pluginName = plugin.basePath
+        pluginName = plugin.name!
         basePluginPath = pluginsConfigurationDirectory
       }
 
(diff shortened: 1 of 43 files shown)


ern-core/src/GitManifest.ts
```diff
@@ -232,12 +232,12 @@ export default class GitManifest {
...
-      if (scopeNameRe.test(plugin.basePath)) {
-        pluginScope = scopeNameRe.exec(plugin.basePath)![1]
-        pluginName = scopeNameRe.exec(plugin.basePath)![2]
+      if (scopeNameRe.test(plugin.name!)) {
+        pluginScope = scopeNameRe.exec(plugin.name!)![1]
+        pluginName = scopeNameRe.exec(plugin.name!)![2]
         basePluginPath = path.join(pluginsConfigurationDirectory, pluginScope)
       } else {
-        pluginName = plugin.basePath
+        pluginName = plugin.name!
```
friederbluemle: What exact does it mean that these properties now have a `!` at the end (compared to the previous behavior)?

friederbluemle: Does it mean the property can't be null?

belemaire: Yes, because `name` can be undefined (in the case of packages with a git path), TypeScript will complain. This is just to tell TypeScript "I know this property can be undefined under certain circumstances, but in the context of this code, it will never be undefined" (in this specific case, we do not support plugins -native modules- coming from git yet, only direct registry package paths, so `name` will always be defined).
It also answers your question in the comment you left, regarding how git `PackagePath` will be handled given that `name` will be undefined. TypeScript will complain every time we use `name` at locations where it cannot be `undefined`, so at least we are made aware that `name` can be undefined in certain cases and react accordingly.

belemaire:
@friederbluemle 
Will release this one in 0.39, as soon as possible this week (just waiting to finalize bugsnag integration that will be shipped in 0.39 as well).

PR #1494, opened 2020-02-18
Add new features directive for android plugin config

belemaire:


docs/platform-parts/manifest/native-modules.md
```diff
@@ -153,6 +153,24 @@ compile 'com.google.android.gms:play-services-maps:10.0.1'
 compile('com.crashlytics.sdk.android:crashlytics:2.9.2@aar') { transitive = true }
 ```
 
+- `features`
+
+An array of one or more [Android hardware or software features](https://developer.android.com/guide/topics/manifest/uses-feature-element) used by the native module, to add to the container manifest when injecting this plugin.
+
+**For example**
```
krunalsshah: Will create an issue to support `android:required=["true" | "false"]` &   `android:glEsVersion="integer"` directives

PR #1495, opened 2020-02-18
Fix crash when calling optional delegate method

jw2175:
Method is marked as optional, should check if delegate responds to selector before calling.


belemaire:
:+1: Great, thanks !

PR #1496, opened 2020-02-19
Enable basic GitHub Actions CI

friederbluemle:
This adds a basic GitHub Actions CI config, running Lint checks and unit tests (using Node.js v10 and v12).

We can just run this in parallel to the existing CI for a while to see how it works out.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,17 @@
+name: ci
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [8.x, 10.x, 12.x]
+    steps:
+      - uses: actions/checkout@v1
+      - name: Use Node.js ${{ matrix.node-version }}
+        uses: actions/setup-node@v1
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,17 @@
+name: ci
+on: [push, pull_request]
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.x, 12.x]
```
belemaire: Actually can we also add `8.x` here ? That's the minimal version we support.

friederbluemle: Yep sure I'll add it. But just FYI, React Native is in the process of dropping 8.x (the version is also no longer officially supported, maintenance ended at the end of 2019). For example one commit that got pushed to `master` today: https://github.com/facebook/react-native/commit/b53aeab10cecfa9b08879b846f8e33d41b28c898
But I'm OK adding it here for now, and we can consider removing it once we move up to the latest version of RN.

friederbluemle: https://nodejs.org/en/about/releases/

Issue #1473, opened 2020-02-01
Prettify json files


belemaire:
No, my bad it would seem. fixed.
prettier is currently configured to ignore json files, otherwise it would have not happened.
don't remember the details as to why we added json files to the exclusion list, but should look into it and open another PR to address this (i.e remove json files from exclusion). Looks like it was done as a quick workaround to some kind of issue (maybe some tests doing exact file content comparison and produced json content is not using prettier rules making test fails if fixture is prettified) but if that's the case we should rather ignore specific json files or directories rather than excluding them all (as first step, and then maybe, if that's the case, apply same prettier rules to generated json files). Will open issue for this based on this comment.

_Originally posted by @belemaire in https://github.com/electrode-io/electrode-native/pull/1472/review_comment/create_

PR #1502, opened 2020-02-23
Use TypeScript noImplicitAny

belemaire:
Improve type safety by enabling TypeScript compiler option [`noImplicitAny`](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html#no-implicit-any) across all modules, but `ern-api-gen` (too complex to port fully to TypeScript at this point).

- **Add typings dependencies *`@type`* for packages that are exposing one.**
There are two ways for TypeScript types to be distributed for a package. Either they are shipped within the package itself, or separately. In the latter case, the types are distributed as `@type` packages. A community maintained monorepo of package typings, [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped), store all these type packages.
Because we are using VSCode for Electrode Native development, these type packages were automatically retrieved through [automatic type acquisition](https://code.visualstudio.com/docs/nodejs/working-with-javascript#_typings-and-automatic-type-acquisition) feature of VSCode. That's why even though we were not explicitly setting the packages in `devDependencies`, they were still downloaded and we had access to types of these packages.
But this feature while looking super convenient, is a double edged sword ... Type safety is only enforced locally on our development machines (even during build checks prior to a git push), but is not enforced for developers not using VSCode, nor is it enforced during build on CI (as it is not using VSCode). Meaning that an external contributor, not using VSCode, could make a bad commit, that still would pass CI build / PR checks, but which would fail our build locally *-sad face-*
Having *`@type`* dependencies explicitly listed, ensure that whatever IDE is used, the build will fail locally if there is some type issue, but will also properly fail on CI. Enabling `noImplicityAny` does not leave us a choice, as at compile time the compiler will reject any package that does not have types, and forces us to explicitly declare type packages *-happy face-*.

- **Fix build errors, following usage of noImplicitAny option**
Once `noImplicitAny` got enabled, a few build errors started popping. The good thing is that because of our already -relatively- proper use of TypeScript, there were not that many. Basically every places were `any` was implicitly inferred, it raised an error.
Update of code in that case is pretty straightforward. Just specify a proper type. If the type is complex, just use `any`. Yes, we can still use `any`, it just has to be explicit ;)
Ideally we shouldn't have any `any` in the code. Through further refactoring we might get rid of them all. But at least with this change, we are properly noticing all of them, and **any time we explicitly use `any` we should really think hard to introduce a proper type instead**.
Unfortunately some packages do not include TypeScript types, nor do they have a `@type` package available. This is the case for some less popular packages. In that case, we just a `declare module [package-name]` statement to explicitly define an empty typing module for this package. Ideally at some point, we should contribute back to the typings monorepo for these packages.

- **Bump inquirer to v7**
This might seems unrelated to this PR, and I don't remember the exact context, having started this work a few months back and put it on hiatus until now. But I think this is because there was no typings exposed for the older version of this package. Anyway, this is a good thing to bump to this latest version of the library. But yes, updating the version is something that should ideally have been done through a different PR. That's ok, no big deal ;)

- **Fix a couple of bugs**
A couple of bugs were revealed in the `GitHubApi` class, in methods to read/update a file in a remote repository. These methods were buggy, but we never noticed the bug as we are not using these methods internally and it seems they were not relied on externally as well. They got fixed in this PR.


package.json
```diff
@@ -35,9 +35,34 @@
...
+    "@types/fast-levenshtein": "^0.0.1",
+    "@types/fs-readdir-recursive": "^1.0.0",
+    "@types/fs-extra": "^8.1.0",
+    "@types/get-folder-size":"^2.0.0",
+    "@types/got":"^9.6.9",
+    "@types/glob":"^7.1.1",
+    "@types/lodash": "^4.14.149",
     "@types/mocha": "^5.2.7",
+    "@types/mustache": "^0.8.32", 
+    "@types/ncp": "^2.0.3",
```
friederbluemle: Some added trailing whitespace on _several_ new lines in this file. Could you remove it before merge please?
Also: Please ensure _one_ space character after `:` (will be autoformatted by Yarn/npm otherwise).

package.json
```diff
@@ -35,9 +35,34 @@
...
+    "@types/lodash": "^4.14.149",
     "@types/mocha": "^5.2.7",
+    "@types/mustache": "^0.8.32", 
+    "@types/ncp": "^2.0.3",   
+    "@types/node-fetch": "^2.5.4", 
+    "@types/semver": "^6.2.0",  
+    "@types/shelljs":"^0.8.6",
+    "@types/treeify": "^1.0.0",
+    "@types/tunnel":"^0.0.1",
+    "@types/tmp":"^0.1.0",
```
friederbluemle: Example for no space after `:`.

belemaire:
@friederbluemle Thanks
Looking forward to get this one cleared https://github.com/electrode-io/electrode-native/issues/1473 to avoid this kind of issues in the future.

PR #1503, opened 2020-02-24
Enforce yarn version

belemaire:
Use [Yarn policies](https://classic.yarnpkg.com/en/docs/cli/policies/) mechanism to enforce yarn version to be used by the project.

This ensure that the version of [yarn](https://classic.yarnpkg.com/en/) used when running yarn commands inside Electrode Native workspace is the same across all developers and build systems working on this project.

Initially enforce yarn version to current latest `1.22.0`.


belemaire:
Holding off this one for now following some internal discussion.

> Just want to make sure that the yarn version used for the project is similar for all developers / builds, so that we don't run into such "oh this doesn't work on my machine, but it works on yours" issues.
> I'm not that much concerned of some additional MB in repo if it guarantees env consistency. Putting version in the README can be informal, but we cannot guarantee devs to follow it.
> That being said, we never ran into any issue with different versions of yarn afaik, so it was more of a preventive PR. Recent versions are pretty much compatible and didn't change the structure of the yarn.lock.
> I'll close this PR for now, but if we ever strongly require a specific yarn version in the future (i.e we use a specific feature only available from this version onward) or if we get some report of some issues between different versions when with electrode native project, we will make this move.

PR #1513, opened 2020-03-07
Set node engine restriction to >=8.3

friederbluemle:
A quick change to align all `engines` declaration in our 18 package.json files to now require `node >=8.3`.

Previously it was set to `>=6` for most packages (`ern-util-dev` was still on `>=4.5`). A few packages also had an (now?) unnecessary `npm` requirement, which I removed.

`react-native` was updated to require `>=8.3` on Jul 17 2018 (https://github.com/facebook/react-native/commit/e64e13fce394332ce609f0def35fa573f30138e9) and `>=8` for a significant time before that.

While I didn't check, it's unlikely that Node.js 6 would even work for us (we're also not running any CI with if as far as I know).


belemaire:
Thanks @friederbluemle :+1:
Our current minimum Node version for Electrode Native is Node 8, even though it had not been enforced through `engines` of `package.json` (looks like forgot to do that).
I think we mention the minimum version of Node in the README in prerequisites I guess. Could you have a look, quick sweep through .md files. Should prob be updated as well part of this PR

friederbluemle:
Good cache, the docs needed some updates too. I amended the commit.

PR #1517, opened 2020-03-10
Local config improvements

belemaire:
- Look up in parent directories chain to resolve `.ernrc` configuration file
- Add support for some initial placeholders that can be used in `.ernrc` configuration files

With this update, one will be able to get rid of some `ern platform config set` commands especially in CI envs, which was the only way until now to set some configuration keys with a value coming from an environment variable (in our case, this will speed up our CI runs a bit because we won't have to run these commands). 

These changes are also necessary to support upcoming Electrode Native workspaces feature.

### DIFF
--- a/docs/cli/platform/config/set.md
+++ b/docs/cli/platform/config/set.md
@@ -68,7 +68,7 @@ HTTP/HTTPS proxy to use to connect to the bundle store server.
 Should be the full url to the proxy, including the port. For example `http://10.0.0.0:9089`.  
 **default** : no proxy
 
-- `sourceMapStoreProxy` [string]  
+- `sourceMapStoreProxy` [string]
 HTTP/HTTPS proxy to use to connect to the source map store server.  
 Should be the full url to the proxy, including the port. For example `http://10.0.0.0:9089`.  
 **default** : no proxy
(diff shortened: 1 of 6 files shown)


docs/cli/platform/config/set.md
```diff
@@ -82,4 +82,18 @@ Should be the full url to the proxy, including the port. For example `http://10.
  
 * In case a value already exists in the configuration for a given key, this command will not fail and will overwrite the existing value.
 
+#### Placeholders
+
+Electrode Native support the following placeholders and will replace them accordingly when loading the configuration :
+
+- `${env.[ENV_VAR_KEY]}`  
+Will be replaced with the value of `[ENV_VAR_KEY]` environment variable.
```
friederbluemle: Maybe just use `ENV_VAR_KEY` instead of `[ENV_VAR_KEY]` here? Otherwise it might look like the square brackets are required/valid syntax.

Also: There are two trailing spaces on three of the 14 added lines in this file (not sure if they were added explicitly or somehow by the editor) ;). Please consider just using a backslash (`\`) if a simple line break without creating a new paragraph is necessary. The hope is that a some point in the future (the final step of #1511), we will be able to run the check against the empty tree object on CI to ensure _all_ files in the repo remain whitespace error free.

In this specific case:

```
- `${env.ENV_VAR_KEY}`\
Will be replaced with the value of `ENV_VAR_KEY` environment variable.
```

instead of

```
- `${env.[ENV_VAR_KEY]}`<SPACE><SPACE>
Will be replaced with the value of `[ENV_VAR_KEY]` environment variable.<SPACE><SPACE>
```

Since `[...] environment variable.` is followed by two LF characters, neither `<SPACE><SPACE>` nor `\` is necessary there to create the paragraph (it will be rendered exactly the same way).

belemaire: Thanks for review :+1: Will address this

belemaire: Yes I was using two spaces for line break, didn't know about the `\` option. Thanks for letting me know :)

PR #1523, opened 2020-03-11
Introduce workspace composite

belemaire:
Some refactoring around composite / container, introducing a new composite type.

Up until now, the composite project was fully generated by Electrode Native. 

This PR makes it possible to instead use an existing local composite project, rather than generating it. The `Composite` class is now becoming a top level interface, and two new implementers classes have been created : `GeneratedComposite` (our current fully generated composite) and `WorkspaceComposite` (a local non generated workspace composite). 

This new composite type is only being declared/introduced through this PR. It's not used yet part of this PR.  It is some necessary prep work for upcoming Electrode Native workspaces feature.

Given the extent of this refactoring, system tests have been run locally and are all green.


ern-composite-gen/src/patchCompositeBabelRcRoots.ts
```diff
@@ -0,0 +1,84 @@
...
+    metroVersion: compositeMetroVersion,
+    rnVersion: compositeReactNativeVersion,
+  })
+
+  // If React Native version is less than 0.56 it is still using Babel 6.
+  // In that case .babelrc files will be processed in any node_modules
+  // which is not a desired behavior.
+  // Therefore we just remove all .babelrc from all node_modules.
+  // We only keep .babelrc of node_modules that are "whitelisted",
+  // as we want them to be processed.
```
friederbluemle: That's wild.. Did you run these tests locally? I assume it's such a specific scenario we don't have CI tests for it?

ern-container-gen/src/addContainerMetadata.ts
```diff
@@ -6,8 +6,10 @@ import fs from 'fs-extra'
 export async function addContainerMetadata(conf: ContainerGeneratorConfig) {
   const metadata: ContainerMetadata = {
     ernVersion: Platform.currentVersion,
-    jsApiImpls: conf.composite.getJsApiImpls().map(j => j.toString()),
-    miniApps: conf.composite.getMiniApps().map(m => m.packagePath.toString()),
+    jsApiImpls: (await conf.composite.getJsApiImpls()).map(j => j.toString()),
+    miniApps: (await conf.composite.getMiniApps()).map(m =>
```
friederbluemle: Just wondering: Is this general cleanup or related to the other changes?

belemaire: Related to other changes. The two functions that are called where synchronous prior to this PR, and were changed to asynchronous ones, as part of a change in this PR, thus the need to add the `await` keyword when calling these functions.

ern-composite-gen/src/patchCompositeBabelRcRoots.ts
belemaire: Yes that is wild. This function was actually not introduced nor modified in any way part of this PR, it was just renamed. I don't think we have tests for this one, but it's been manually tested prior to initial release (more than a year back), and has been battle proofed since then (most of the MiniApps we have are having local .babelrc and thus all composites going from more than a year have been going through this code)

friederbluemle: Ooh okay got it. I saw it was just moved to a different file. :+1:

PR #1525, opened 2020-03-14
Add resetCache option to start command

belemaire:
Left over from https://github.com/electrode-io/electrode-native/pull/1520

Add `resetCache` option to `start` command.

### DIFF
--- a/docs/cli/start.md
+++ b/docs/cli/start.md
@@ -51,6 +51,10 @@ If you do not pass an argument to this command, you are prompted to select a nat
 `--port`
 * Port on which the local packager should listen on *(default: 8081)*
 
+`--resetCache`\
+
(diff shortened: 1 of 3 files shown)


docs/cli/start.md
```diff
@@ -51,6 +51,10 @@ If you do not pass an argument to this command, you are prompted to select a nat
 `--port`
 * Port on which the local packager should listen on *(default: 8081)*
 
+`--resetCache`\
```
friederbluemle: Minor suggestion (no need to change here, just for next time): When using _two_ `LF`, the `\` is _not_ needed. It's only necessary when the source markdown has _one_ `LF` (to create a single line break in the rendered document).

PR #1535, opened 2020-04-06
Inline documentation fixes

simonbilskyrollins:
This PR fixes three issues, all related to inline documentation:

1. Some links to the documentation website were broken and would redirect to the docs home page.
2. iOS docstrings with HTML `<a></a>` tags inside them were causing Xcode compiler warnings. I replaced these tags with Markdown-formatted links as per [Apple's guidelines](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/Links.html) and reformatted the docstrings into modern Swift style.

Xcode rendered documentation (before):

Screen Shot 2020-04-06 at 10 23 35 AM

Xcode rendered documentation (after):
Screen Shot 2020-04-06 at 10 30 21 AM

3. iOS docstrings with empty `@param` declarations were causing Xcode compiler warnings. I filled them out with a short description of the corresponding parameter.


simonbilskyrollins:
That's a good question. Both styles are equally valid/functional as far as I know, but I default to the triple slash because that's what Xcode generates for you when you select a function definition and choose `Editor > Structure > Add Documentation (option-command-/)`.

simonbilskyrollins:
And although I can't find anything about it in Apple's documentation, [Google's Swift style guide](https://google.github.io/swift/#documentation-comments) prefers the triple slash as well.

friederbluemle:
Great thanks for the additional information!

PR #1540, opened 2020-04-09
Refactor manifest plugin config

belemaire:
Refactor manifest plugin configuration, getting rid of some verbosity and improve type safety. 
One of the main change is that `getPluginConfig` method now takes the `platform` for which to get configuration of. Therefore, instead of returning the whole `ios` and `android` configuration objects, it only returns the configuration of the requested platform.

This PR is in the context of RN61+ support, as this refactoring is needed to more easily support some needed changes in the manifest plugin configuration in this context. It's been open separately not to pollute RN61+ support PR with such a consequent refactoring.

### DIFF
--- a/ern-container-gen-android/src/AndroidGenerator.ts
+++ b/ern-container-gen-android/src/AndroidGenerator.ts
@@ -17,6 +17,7 @@ import {
   BundlingResult,
   HermesCli,
   gitApply,
+  PluginConfig,
 } from 'ern-core'
 import {
   ContainerGenerator,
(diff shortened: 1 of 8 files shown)


ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -184,10 +179,11 @@ export default class AndroidGenerator implements ContainerGenerator {
             config.outDir,
             'lib/src/main/java/com'
           )
+
```
friederbluemle, line 182: Minor: Remove added blank line (or alternatively, add one before line 192 as well).

PR #1544, opened 2020-04-13
Add baseComposite option to code-push release command

belemaire:
Add missing `--baseComposite` option to `code-push release` command.
The logic for this option was already in place in `performCodePushOtaUpdate` method, so it's clear that missed adding this option to this command when this feature was introduced.
Also apply fix from https://github.com/electrode-io/electrode-native/pull/1532 for properly coercing base composite config from cauldron.

### DIFF
--- a/docs/cli/code-push/release.md
+++ b/docs/cli/code-push/release.md
@@ -18,6 +18,10 @@ ern code-push release
 
 ### Options
 
+`--baseComposite <compositePath>`
+
+* Git or File System path, to the custom Composite repository (refer to the [custom Composite documentation][4] for more information).
+
 `--miniapps`
 
 * One or more MiniApps (separated by spaces) version(s) to CodePush.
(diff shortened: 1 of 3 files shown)


krunalsshah:
Does `baseComposite` resolve to `<updateContentsPath>` ? of the code-push command?

belemaire:
@krunalsshah it's unrelated

docs/cli/code-push/release.md
```diff
@@ -18,6 +18,10 @@ ern code-push release
 
 ### Options
 
+`--baseComposite <compositePath>`
```
krunalsshah, line 21: We can mention support for this command in section https://native.electrode.io/reference/index#through-commands

belemaire: Good point :+1: post merge though ... 
Will open another PR for this

PR #1545, opened 2020-04-13
Update composite documentation

belemaire:
Add additional command to the list of commands supporting `--baseComposite` option.


belemaire:
Missed in https://github.com/electrode-io/electrode-native/pull/1544

PR #1546, opened 2020-04-14
Remove overlay permission upon app startup in Android

deepueg:
Since React Native 52, the dev screen is displayed using the `currentActivity` context instead of using the overlay permission. This PR is to remove the dependency electrode native has added inside the container code. 

Reference: 
https://github.com/facebook/react-native/commit/d19afc73f5048f81656d0b4424232ce6d69a6368

### DIFF
--- a/ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactActivityDelegate.java
+++ b/ern-container-gen-android/src/hull/lib/src/main/java/com/walmartlabs/ern/container/ElectrodeReactActivityDelegate.java
@@ -8,7 +8,6 @@ import android.os.Bundle;
 import android.provider.Settings;
 import android.support.annotation.NonNull;
 import android.view.View;
-import android.widget.Toast;
 
 import com.facebook.react.ReactActivityDelegate;
 import com.facebook.react.ReactNativeHost;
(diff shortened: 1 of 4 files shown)


belemaire:
Thanks :+1: 
LGTM but one thing though. If we want to keep support for RN<52 in versions of ERN>0.41.0 (when this will be shipped) the code shouldn't be removed but instead wrapped in mustache blocks based on RN version (only added for RN>=52). Backward compatibility with older RN versions is something that ERN did up until now.
That being said, I would be O.K to start saying that we drop support for versions of RN<52 in ERN 0.41.0. Because this version will target. 0.62, we can decide starting from now to only support at most 10 versions back of RN in each ERN release targeting a new RN version. This would allow us to get rid of some conditional RN versions switches in the codebase over time. But we would have to come up with a matrix in our README of versions of RN supported by ERN versions.
Thoughts ?

friederbluemle:
Nice :+1:
Just checking: Does the performance monitor still work/show up?

Regarding the ERN<->RN compatibility, I also had several questions.. Maybe @deepueg @belemaire @krunalsshah we can have a separate discussion later today or tomorrow?

deepueg:
> Nice 👍
> Just checking: Does the performance monitor still work/show up?
> 
> Regarding the ERN<->RN compatibility, I also had several questions.. Maybe @deepueg @belemaire @krunalsshah we can have a separate discussion later today or tomorrow?

@frieder: The performance monitor will ask to turn on the overlay permission when clicked.

deepueg:
> Thanks 👍
> LGTM but one thing though. If we want to keep support for RN<52 in versions of ERN>0.41.0 (when this will be shipped) the code shouldn't be removed but instead wrapped in mustache blocks based on RN version (only added for RN>=52). Backward compatibility with older RN versions is something that ERN did up until now.
> That being said, I would be O.K to start saying that we drop support for versions of RN<52 in ERN 0.41.0. Because this version will target. 0.62, we can decide starting from now to only support at most 10 versions back of RN in each ERN release targeting a new RN version. This would allow us to get rid of some conditional RN versions switches in the codebase over time. But we would have to come up with a matrix in our README of versions of RN supported by ERN versions.
> Thoughts ?

I would agree with the supporting 10 versions or so considering the number of changes RN has undergone and gong as deep as 41 might be an overkill.

PR #1551, opened 2020-04-15
Include iPad in iOS simulator deployment targets

### LINKED ISSUE #1549: ern run-ios not able to identify ipad simulators
basically we are using same miniapp which we built it for android to generate the code for IOS and install it on IPAD. our target machine to run app in IPAD devices so when i run ern run-ios it compiles and generate the code required for IOS but ern is listing only mobile iphones and not able to identify IPad simulators. could you please share is there any fix we need to do it.

workaround: once ern-ios compiles the code we are terminating the process and run this command ->npx react-native run-ios --simulator="iPad (7th generation)" to install it on the ipad simulators

friederbluemle:
Previously `ern run-ios` would only list `iPhone`-type simulators. Now, `iPad` is included as well.

Also fixes a minor wording inconsistency (asking to choose "device" when selecting a simulator).

Fixes #1549

### DIFF
--- a/ern-core/src/ios.ts
+++ b/ern-core/src/ios.ts
@@ -18,7 +18,7 @@ export interface IosDevice {
 export async function getiPhoneSimulators(): Promise<any> {
   const iosSims = await simctl.getDevices()
   return _.filter(_.flattenDeep(_.map(iosSims, (val, key) => val)), device =>
-    device.name.match(/^iPhone/)
+    device.name.match(/^iPhone|iPad/)
   )
 }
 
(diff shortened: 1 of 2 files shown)


belemaire:
Due to renaming from `selectedDevice` to `selectedSimulator`
Need to update prompt object returned by some tests :
https://github.com/electrode-io/electrode-native/blob/master/ern-core/test/ios-test.js#L124

friederbluemle:
Sure, I'll check the tests. Had a quick look at them while changing the code, but did not notice, sorry

PR #1554, opened 2020-04-20
Speedup composite generation and start command when using local miniapps

belemaire:
This PR reduces the time it takes to generate a composite that includes local miniapps (not retrieved from git or npm, but sitting in a local directory). It also contains an improvement to `ern start` command, leveraging this new composite generation.

From a high level perspective, instead of running `yarn add` to add a local miniapp to the composite, as done before, each local miniapp is now "linked" with the composite, by building an adequate metro configuration. This means that local miniapps are not taking any extra time to be added to a composite during its generation.

On the down side, react-native still needs to be hoisted to the top level composite directory, so when generating a composite with only local miniapps, the composite generator will still have to install react-native in the top level composite. This is the only extra install needed, meaning that composite generation time will be constant independently of the number of local miniapps part of it.

This change will speedup all commands that are, under the hood, generating a composite with local miniapps.

Regarding `ern start` command changes, in addition to the command speedup itself being a by product of faster composite generation, this PR also gets rid of the file system watcher that was used by the command to detect any changes to linked miniapps directories and copy the changes to the miniapp directory in composite. It is not needed anymore because new composite generator stopped installing local miniapps into the composite node modules. This means that fewer machine resources are now needed for the command. Also, in addition of getting rid of some extra code, we are also getting rid of two dependencies, that are not needed anymore, respectively `chokidar` (was the fs watcher, not needed anymore) and `ncp` (was used for recursively copying the linked miniapp to the composite node modules).


deepueg:
Nice :tada:

ern-core/src/resolveNativeDependenciesVersions.ts
```diff
@@ -66,7 +66,10 @@ export function resolvePackageVersionsGivenMismatchLevel(
     const pluginVersions = _.map(entry, 'version')
     if (pluginVersions.length > 1) {
       // If there are multiple versions of the dependency
-      if (containsVersionMismatch(<string[]>pluginVersions, mismatchLevel)) {
+      if (
+        name !== 'react-native-electrode-bridge' &&
+        containsVersionMismatch(<string[]>pluginVersions, mismatchLevel)
```
friederbluemle, line 71: Are there any other options that we could use to avoid this kind of special-casing?

belemaire: Yes tested a few uses cases with only local miniapps and a mix of git / local / npm miniapps, all good :) Also we had some internal users giving it a shot using my branch with their own use cases.
And yes for this specific line, it's one of the only thing in this PR that I was planning to do another pass on today because it's not proper, you have a good eye ;) Expect some change later on.

belemaire: OK, updated. Actually not much change here, just narrowing the exception for bridge to only return mismatched versions in case there are different major versions of bridge used across miniapps (instead of different patch/minor versions as its the case for other native modules). 
When generating composite/container it's fine if there are multiple bridge versions as long as they don't major differ, as we will retain the highest version used, and we know that for the bridge we guarantee that minor/patch updates are backward compatible (we don't know that for sure for other native modules thus the higher restrictions).
We need this in this PR, because most miniapps are depending on the same semver range for bridge (actually the APIs used by the miniapps), so when we were running `yarn add` for each local miniapp, it would install  the latest version of the bridge matching the semver and would be aligned for all miniapps. With new logic, we don't `yarn add` the miniapps into the composite but pick them directly from their local directory, where they have a `yarn.lock` or `package-lock.json` that has captured a specific version of the bridge, which might not be aligned with other miniapps.

friederbluemle: Okay cool, thanks for the additional details! 👍 Definitely better to only lock it down on the major version range.

@deepueg - Can you please also take a quick look at this and add your approval? Because I remember several months ago, we had to apply a **very** similar patch locally to get something with navigation (?) working. Was it related to the same code?

I must also admit that my understanding of the exact design process/thinking behind the version matching/enforcement logic is not deep enough. I guess the main point is that hard-coding and special casing a package name like that suggests that a more generic and scalable solution must be possible (not saying this has to be done here, since I can imagine the impact would be much greater) So yeah, definitely not blocking this PR :)

deepueg: @friederbluemle That was to omit a navigation-impl plugin getting added into the container. That is still a requirement for us to properly generate containers for api-impl development.

PR #1556, opened 2020-04-22
Deduplicate yarn.lock

friederbluemle:
A quick run of [`yarn-deduplicate`](https://www.npmjs.com/package/yarn-deduplicate), saving several hundred lines in `yarn.lock`. The resulting `node_modules/` folder is around 30 MB smaller, and _should_ result in the same resolved dependency tree. All tests are passing locally, and if CI is green too, I think we should be good with the merge - @belemaire?

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -52,12 +52,7 @@
     esutils "^2.0.2"
     js-tokens "^4.0.0"
 
-"@babel/parser@^7.4.3":
-  version "7.5.5"
-  resolved "https://registry.yarnpkg.com/@babel/parser/-/parser-7.5.5.tgz#02f077ac8817d3df4a832ef59de67565e71cca4b"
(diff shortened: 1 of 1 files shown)


friederbluemle:
Mmmmh..

```
error mkdirp@1.0.4: The engine "node" is incompatible with this module. Expected version ">=10". Got "8.17.0"
```

```
=> Found "js-beautify#mkdirp@1.0.4"
info This module exists because "_project_#ern-composite-gen#js-beautify" depends on it.
info Disk size without dependencies: "64KB"
info Disk size with unique dependencies: "64KB"
info Disk size with transitive dependencies: "64KB"
```

friederbluemle:
Most likely an issue with `mkdirp`.. Looks like it recently had the maintainer/org changed. Opened an issue here: https://github.com/isaacs/node-mkdirp/issues/24

PR #1560, opened 2020-04-22
Exit process if Node version does not satisfy requirement

belemaire:
Ensure that the Node version used to run `ern` is compatible with Node engine requirements of current `ern` version, otherwise exit the process.

This will avoid some bad surprise when not using a compatible node version, which might cause some `ern` command to behave incorrectly or crash.

This is needed because Iinstallation of an `ern` version and actual run of `ern` commands are decoupled in time, meaning that one could use Node 12 to install ern 0.41.0 for example, satisfying the requirement at the time of installation, but at the time of actually running an `ern` command, the node version might have been changed by the user, to an incompatible one.

### DIFF
--- a/ern-local-cli/src/index.ts
+++ b/ern-local-cli/src/index.ts
@@ -1,6 +1,6 @@
 import chalk from 'chalk'
 import { execSync } from 'child_process'
-import fs from 'fs'
+import fs from 'fs-extra'
 import { getActiveCauldron } from 'ern-cauldron-api'
 import {
   config,
@@ -20,6 +20,7 @@ import {
 } from 'kax'
 import path from 'path'
 import yargs from 'yargs'
+import semver from 'semver'
 
 // ==============================================================================
 // Geeky eye candy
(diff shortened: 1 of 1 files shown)


ern-local-cli/src/index.ts
```diff
@@ -189,6 +201,15 @@ const logLevelStringToEnum = (level: string) => {
       ? new KaxNullRenderer()
       : new KaxAdvancedRenderer(kaxRendererConfig)
 
+  const nodeVersion = process.version
+  if (!isNodeVersionCompatible(nodeVersion)) {
+    const requiredVersion = getNodeEngineVersionRequirement()
+    log.error(`This version of ern requires Node ${requiredVersion}
+Node version currently in use is ${nodeVersion}
+Please switch to a version of Node satisfying the version requirement`)
```
friederbluemle, line 209: Minor: If you still need to rebase, you can update the text to the full name `Node.js` instead of `Node`.

Issue #1443, opened 2020-01-05
An error occurred: ChildProcess exited with code null when creating a MiniApp


DylanStandridge:
I also have android studio using 7.0 Nougat as the tutorial directs. Looking over creating a mini app i also see no options for additional debugging info. 
ElectrodeNativeTutorial dstand1$ npm -v
6.12.1
ElectrodeNativeTutorial dstand1$ node -v
v12.13.1
ElectrodeNativeTutorial dstand1$ ern -v
[v0.38.10] [Cauldron: -NONE-]

ern create-miniapp MovieListMiniApp
 ___ _        _               _       _  _      _   _         
| __| |___ __| |_ _ _ ___  __| |___  | \| |__ _| |_(_)_ _____ 
| _|| / -_) _|  _| '_/ _ \/ _` / -_) | .` / _` |  _| \ V / -_)
|___|_\___\__|\__|_| \___/\__,_\___| |_|\_\__,_|\__|_|\_/\___|
[v0.38.10] [Cauldron: -NONE-]

✔ Ensuring that Electrode Native module name is valid 0s
? Input a package name. Press Enter to use the default. movie-list-miniapp
? Choose the language to use for this MiniApp TypeScript
✔ Ensuring that Electrode Native module name is valid 0s
✔ Ensuring that NPM package name is valid 0s
✖ Creating MiniApp 49s
  ✔ Retrieved react-native version from Manifest [0.60.5] 2s
    ✔ Syncing https://github.com/electrode-io/electrode-native-manifest.git Mani
fest 2s
  ✖ Creating MovieListMiniApp project using react-native v0.60.5 47s
✖ An error occurred: ChildProcess exited with code null

friederbluemle:
Hi @DylanStandridge - Please let us know if you're still seeing the same issue.

You can get additional output for debugging by setting the `ERN_LOG_LEVEL` environment variable, or the `logLevel` option in your global `.ernrc` file.

```sh
ERN_LOG_LEVEL=trace ern create-miniapp MovieListMiniApp
```

```
// ~/.ern/.ernrc

{
  // ...
  "logLevel": "trace"
}

Issue #1254, opened 2019-07-17
Bundler error in Typescript miniapps


romanlv:
When attempting to `ern run-ios` miniapps created with Typescript support I'm getting 

```
✖ Generating Container locally 11s
  ✖ Generating Container 11s
    ✔ Preparing Native Dependencies Injection 0s
    ✔ Preparing API Implementations Injection 0s
    ✔ Injecting Native Dependencies 3s
    ✔ Adding Native Dependencies Hooks 0s
    ✖ Bundling MiniApps 4s
      ✖ Running Metro Bundler 4s
  ✖ runLocalContainerGen failed: Error: Command failed: /Users/vn0epsg/.ern/versions/0.34.0/node_modules/ern-core/nod
e_modules/.bin/react-native bundle --entry-file=index.js --dev=false --platform=ios --bundle-output=/Users/vn0epsg/.e
rn/containergen/out/ios/ElectrodeContainer/Libraries/MiniApp/MiniApp.jsbundle --assets-dest=/Users/vn0epsg/.ern/conta
inergen/out/ios/ElectrodeContainer/Libraries/MiniApp 
  ✖ error Unable to resolve module `react-native/Libraries/NewAppScreen` from `/private/var/folders/w1/fr2ppby15bjb5t
6mb0hrx059h5v0_t/T/tmp-64231I2QN7RvX41xe/node_modules/ts-miniapp/App.tsx`: Module `react-native/Libraries/NewAppScree
n` does not exist in the Haste module map

```
It seems to be related to this [react-native-template-typescript issue](https://github.com/react-native-community/react-native-template-typescript/issues/42)

ℹ ern-local-cli : 0.34.0
ℹ electrode-native : 1.0.17

Js based apps start fine

belemaire:
Hi @romanlv 
Thanks for reporting this one and doing some initial research.
As described in the issue you linked, the TypeScript template that we are using internally when generating a TypeScript MiniApp (the exact one you linked) is always targeting latest version of React Native. That is going to cause problems as Electrode Native users are using different versions of React Native and not always latest. 
We'll look into that ASAP and probably release a patch version of Electrode Native to include a fix. Given that react-native CLI does not currently support templates for different versions of RN, we'll have to come up with our way. 
Meanwhile, if your'e blocked, have your tried the workaround suggested in this comment ? 
https://github.com/react-native-community/react-native-template-typescript/issues/42#issuecomment-510827188

romanlv:
[...]

PR #1563, opened 2020-04-23
Fix create-miniapp template processing (TypeScript)

friederbluemle:
A few closely related fixes and updates here. High level summary:

### React Native CLI

With this PR merged, we will be using `npx --ignore-existing react-native` to invoke the `init` command of the React Native CLI when RN >=0.60. Previously, we would just run `react-native`, which either resolves in a globally installed old CLI, or the `react-native-cli` package dependency (old as well). It is not recommended to use the old CLI with RN 0.60+. One problem that it puts non-reproducible and potentially incompatible devDependencies in the resulting `package.json`. The other, more visible problem is related to broken template processing (see below).

### TypeScript template

`create-miniapp`'s template processing that has been broken for quite a while, and as a side effect of the updated RN CLI, this is now fixed. The problem was the old cli's ability to correctly process a template with RN 0.60+ was completely broken. Now that we're using the new way to call RN using `npx` (with a couple of tweaks), template processing works again!

### `--language` deprecation

The `--language` option was never really compatible with the `--template` option in the first place (an unlikely scenario tbh), but now it has been cleaned up and streamlined. The functionality has been fully replaced by the more precise and flexible `--template` option.
We now follow the behavior of React Native (i.e. _no_ additional prompt, and _no_ custom template by default). If needed, a template _can_ be supplied on command line. For now, existing options should continue to work as before, and a deprecation notice will be printed to the console. We can remove them in a future release.

I tested this with RN 0.60 and RN 0.62 (pre-merge, @belemaire's current `cocoapods` branch), both work with the default template and [react-native-template-typescript](https://github.com/react-native-community/react-native-template-typescript).

For example, if the TypeScript template should be used, this is the command with the current version of ERN (i.e. RN 0.60):

```sh
ern create-miniapp --template react-native-template-typescript@6.2.0 app01
```

For RN 0.62, **no** version number after the template is needed. The version compatibility is listed on their website.

Related: #1254 #1443


belemaire:
Look like CI checks are failing because this PR uses `deprecateOption` of `yargs` but it seems like it doesn't exist in the version used. My guess is that this PR is dependent on this one https://github.com/electrode-io/electrode-native/pull/1562 which bump `yargs` version.
If that's the case could you merge https://github.com/electrode-io/electrode-native/pull/1562 asap and rebase this one ? I'll review it more in depth later on.

friederbluemle:
Yes, it initially did fail with `deprecateOption` (which is why I temporarily included the commit to update yargs here). The CI failure was then again because of the unit test timeout. Anyway, now that #1562 is merged, I can rebase now.

friederbluemle:
Ooh.. Hmmm.. So it does fail with a different error on Azure and GH Actions. Unit test timeout on GH Actions and this error on Azure:

```
$ ern-typescript
running tsc with [ '--project',
  '/Users/runner/runners/2.166.3/work/1/s/ern-local-cli/tsconfig.release.json' ]
src/commands/create-miniapp.ts(31,6): error TS2339: Property 'deprecateOption' does not exist on type 'Argv<{ language: string | undefined; }>'.
```

I know the `deprecateOption` is pretty new and not yet part of the type definitions. I am using it locally, and while the IDE does put a squiggly line under it, the code works fine. Does it mean we cannot use it yet @belemaire? Or need a manual type definition?

belemaire:
Yeah the code will work, but type checking will fail :(

Indeed it's not yet available in [yargs type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/yargs/index.d.ts)
Should be added soon I guess, it's a popular package so types should be well maintained (I could have a take at it and open a PR but it's a complex one, and it'll take time to work on and get merged anyway ...).

Meanwhile two solutions I see.
Either get rid of `deprecateOption` and just log that in the description of the command directly 

```
Language to use for this MiniApp [Deprecated. Use --template]
```

Or do a quick TypeScript module augmentation to add the missing type definition on `yargs` module for this function, by just adding this in the `create-miniapp.ts` source file :

```
declare module 'yargs' {
  interface Argv<T = {}> {
    deprecateOption<K extends string>(key: K, message: string): Argv<T>
  }
}
```

And we can then remove it later whenever official type definition is updated to add this function.

belemaire:
@friederbluemle 
Best way for now is to use the `deprecated` property directly rather than chaining function:

```js
.option('language', {
  choices: ['JavaScript', 'TypeScript', undefined],
  deprecated: 'use --template directly',
  describe: 'Language to use for this MiniApp',
  type: 'string',
})
```

This won't complain / won't fail the build and achieve the same.

friederbluemle:
Awesome, thanks a lot for the additional information!!
I went with just putting it in as a separate field (rather than method call). :+1:

belemaire:
Is `npx` required here to make the typescript template properly work or are they unrelated ? From your comment it would sound like its needed.

There are unfortunately some catches with `npx react-native`.

-  It's only supported [starting from react native 0.60.0](https://github.com/react-native-community/cli#using-npx-recommended). 
For example `npx react-native init foo --version 0.59.0` will fail, while it will succeed with any version >= 0.60.0. 
So the problem here is that if we do it this way, we have to mention that ern 0.41.0 will only support react native >= 0.60.0. Because when we run `ern create-miniapp` it checks in the team override manifest for version of react native to use (on which all miniapps are aligned) and some teams might still be on versions of react native < 0.60.0, which will then fail the command (as it runs `react-native init` under the hood). 
I think this could be addressed by only using `npx react-native init` in our `init` function, if the react native versions is >= 0.60.0, otherwise fallback on global `react-native`. The problem should only be limited to `init` because for other commands `npx react-native` will just use the locally installed `react-native` from `node_modules/.bin` (see my next point).
But if not using `npx` (for react native < 0.60.0), I guess typescript template will still fail. But I think we should still keep the `--template` option as it might be used for other templates. The quirks with this option will be the same as if directly using the react native cli and not ern.

- `npx` will only download react-native and use the latest CLI if `react-native` is not in global path. That's the behavior of `npx`, it will first look it a `react-native` binary is on global path, if so it will use it, otherwise it will look in current local path `node_modules/.bin` and if so it will use it, otherwise it will install it on the fly.
What this means is that even if we run `npx react-native init` for version of RN > 0.60.0, it will still use the old CLI that was installed globally :( And will result in same problems with typescript templates I guess. And almost all of current RN developers have it installed globally.
So we will have the kind of 'it's working on my machine but not yours' issues :(
One way to bypass that would be to use the `--ignore-existing` option of `npx` :
```
--ignore-existing      Ignores existing binaries in $PATH, or in the local
                         project. This forces npx to do a temporary install and
                         use the latest version.
```
But the problem with that, is that it will be costly to do that when `react-native` is available locally (for example run-android, start ... all these commands but init will use the local `node_modules/.bin/react-native` when being run with `npx`, so won't need another install).
But I think we could just use this option in `init` and in all other commands just use `npx react-native` without this flag as the binary will be present locally.

So yeah, to put it in a nutshell, maybe a solution here addressing both points is just to :
- Use `npx` in `init` iff react native version >= 0.60.0, and use it with `--ignore-existing` flag.
- Use `npx` in all other functions, without version check, but use it without `--ignore-existing` flag.

You can have a closer look at it to validate :)

friederbluemle:
Thanks @belemaire for the detailed review! 👌 🙇

What you said makes sense, especially with `--ignore-existing` - I actually ran into this same exact problem a while ago, and should have remembered!

I guess if we absolutely have to keep a new "major" version of ERN compatible with old versions of RN (something that even RN's _own_ CLI is _not_ doing!), then we can try the approach of selectively using npx and a locally installed older cli depending on the use case. Btw, personally I think it's great how ERN is attempting to preserve backwards compatibility as much as possible. Though as we all know, looking around the whole ecosystem, breaking compatibility is something that's done in many projects every day, without strong reasons :/ And no, being on a fake "`0.x`" major version range is not a valid excuse (looking at react-native here).

One thing we do need in order to keep this project relevant is to find the right balance of proper engineering and blending in+keeping **up-to-date** with the 100s/1000s of dependencies we have to deal with.

belemaire:
@friederbluemle Ran a few manual tests with this one, all good :+1:
Just one minor thing, could you remove the default value for `language` that is set [here](https://github.com/electrode-io/electrode-native/blob/a494ad49c7186b04816ca90c6894144f6b5f577f/ern-core/src/MiniApp.ts#L102)
Otherwise we run into warning message `⚠ Both language and template are set. Ignoring language.` when running `ern create-miniapp` even without providing the `language` option to the command ;) A bit bothering.
Once addressed, please merge this PR asap :)
Thanks !

friederbluemle:
Found another scenarios that failed (already updated the code): Creating a miniapp using RN < 0.60 and `--language TypeScript` - The template needs a different name depending on the version 🤦

belemaire:
That's OK, starting with 0.41.0 the option is now deprecated so it's assumed it's not maintained anymore and can break in some context 🤣
Anyway, I think we could safely remove this option in 0.42.0 now that its been deprecated. Because this is not a command that is used on a frequent basis, I really doubt some users have scripted / automated it, so a disappearing option won't break a lot -if any- usages.
Please merge whenever you can, waiting on this one for 0.41.0 release 😄

PR #1571, opened 2020-04-27
Update types devDependencies

friederbluemle:
This updates all `@types` dependencies (within semver range).


belemaire:
Could you instead update them in top level `package.json` instead and regenerate `yarn.lock` by running `yarn install` or was there a reason to edit `yarn.lock` directly ?

friederbluemle:
> Could you instead update them in top level `package.json` and regenerate `yarn.lock` by running `yarn install` or was there a reason to edit `yarn.lock` directly ?

I didn't update `yarn.lock` manually. As mentioned, all updates are within semver range (it's just the result of the `upgrade` command).
But sure, I agree it's easier to visually check the "minimum" versions used if `packages.json` is updated as well. I'll align the version ranges to match :+1:

PR #1573, opened 2020-04-28
Remove npm dependency

friederbluemle:
It doesn't look like this dependency was used. Also, Node.js already ships with npm.

This decreases the size of `node_module` by almost 40 MB! 🎉


friederbluemle:
Yup, I also traced back the Git logs, and it turned out it was added by Justin Spears on Apr 4 2017. 😄

PR #1574, opened 2020-04-28
skip native dependencies check during code push promote

krunalsshah:
Moving forward we'd skip the native dependencies alignment checks during the code-push promotion 
- if the source and target app version is the same during the code-push promotion of the bundle. As the native dependencies alignment check is already done during the code-push release.
-  if the source and target app version is different and one needs to skip native dependencies alignment check, pass the `skipNativeDependenciesVersionAlignedCheck` as `true`.

### DIFF
--- a/docs/cli/code-push/promote.md
+++ b/docs/cli/code-push/promote.md
@@ -14,61 +14,16 @@ ern code-push promote
 
 #### Options
 
-`--reuseReleaseBinaryVersion`
+`--description/--des`
 
-* Indicates whether to reuse the target binary version that was used for the initial release
-* If omitted, and `targetBinaryVersion` is not set, the promotion will target the exact version of the descriptor
-* This option is mutually exclusive with `targetBinaryVersion`
+* Description of the changes made to the app with this release. If omitted, the description from the release being promoted will be used.
 
-`--sourceDescriptor <descriptor>`
+* **Default** Empty string
 
-* Specify the native application version from which to promote a release in the form of a *complete native application descriptor*.
-* The release to be promoted will be the latest non disabled release of this native application version.
-* **Default** The command will list all released native applications versions stored in the Cauldron and will prompt to select a target native application version from the list.
+`--disableDuplicateReleaseError`
 
-`--targetDescriptors <descriptors..>`
+* When this flag is set, promoting a package that is identical to the latest release on the target deployment will produce a warning instead of an error
 
-* Specify one or more target native application version to promote the release to, in the form of a *complete native application descriptor* list (separated by spaces).
-* The target descriptor can be the same as the source descriptor if the promotion is only changing the deployment name (for example promoting a release from Staging to Production for the same native application version).
-
-`--targetSemVerDescriptor <descriptor>`
-
-* A native descriptor using a semantic version string for its version. The promotion will target all native application versions matching the semver.
-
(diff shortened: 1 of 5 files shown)


ern-orchestrator/src/codepush.ts
```diff
@@ -99,6 +100,7 @@ export async function performCodePushPromote(
     rollout?: number
     reuseReleaseBinaryVersion?: boolean
     targetBinaryVersion?: string
+    skipNativeDependenciesVersionAlignedCheck?: boolean
```
friederbluemle, line 103: Minor: Looks like the other entries are (roughly) in alphanumeric order. Want to move this line up above `targetBinaryVersion`? Same on line 93.

ern-orchestrator/src/codepush.ts
```diff
@@ -140,19 +142,30 @@ export async function performCodePushPromote(
...
-        throw new Error(
-          'Native dependencies versions of MiniApps are not aligned. Relaunch the operation with the force flag if you wish to ignore.'
+      // If sourceNapDescriptor.version === targetNapDescriptor.version skip the compatibility check
+      // As the compatibility check would have already happened as part of the code-push release
+      // If skipNativeDependenciesVersionAlignedCheck is set to true
+      // and sourceNapDescriptor and targetNapDescriptor versions are different, explicity skip the check
+      // would be required for patch version promotion
+      if (
+        sourceNapDescriptor &&
+        sourceNapDescriptor.version !== targetNapDescriptor.version &&
```
friederbluemle, line 152: Just confirming: It _is_ possible for `sourceNapDescriptor` to be null, but _not_ for `targetNapDescriptor`?

docs/cli/code-push/promote.md
```diff
@@ -71,27 +57,46 @@ If no `targetDescriptors` nor a `targetSemVerDescriptor` is specified, the comma
...
+`--targetDescriptors <descriptors..>`
+
+* Specify one or more target native application version to promote the release to, in the form of a *complete native application descriptor* list (separated by spaces).
+* The target descriptor can be the same as the source descriptor if the promotion is only changing the deployment name (for example promoting a release from Staging to Production for the same native application version).
+
+`--targetSemVerDescriptor <descriptor>`
+
+* A native descriptor using a semantic version string for its version. The promotion will target all native application versions matching the semver.
+
+* If no `targetDescriptors` nor a `targetSemVerDescriptor` is specified, the command will list all released native application versions stored in the Cauldron and will display a prompt to select one or more target native application version(s) for the promotion.
```
friederbluemle, line 99: Just a minor request: It looks like a rather big _reordering_ of existing content **and** the _addition_ of new content happened in the **same** commit. To make it easier to review, next time please use two separate commits. No need to change here :)

krunalsshah: Yeah, different commits would have been relevant here.

ern-orchestrator/src/codepush.ts
krunalsshah: @friederbluemle https://github.com/electrode-io/electrode-native/pull/1574/files#diff-fa17892fa392626ed2f1932956acba0cR123
The check is already here. It would be okay to skip `sourceNapDescriptor` check, though like being defensive :)

ern-orchestrator/src/codepush.ts
krunalsshah: Yeah, most are aligned, not all. I'll add a commit after re arranging.

PR #1575, opened 2020-04-28
Fix system tests and improve error messaging

friederbluemle:
Now that CocoaPods is required for iOS container generation, macOS is a requirement. It'll fail on Linux and Windows:

```
spawnp => command: pod args: ["install"] options: {}
[ Running pod install (Failed) ]
popd
/tmp/tmp-5319hpxSXui6YFVV /tmp/tmp-5319hpxSXui6YFVV
popd
/tmp/tmp-5319hpxSXui6YFVV
[ Generating Container (Failed) ]
[ Generating Container locally (Failed) ]
✖ An error occurred: spawn pod ENOENT
Error: spawn pod ENOENT
    at Process.ChildProcess._handle.onexit (internal/child_process.js:240:19)
    at onErrorNT (internal/child_process.js:415:16)
    at process._tickCallback (internal/process/next_tick.js:63:19)
```

With this PR we're checking the platform and error earlier with an error message, instead of failing later on.

Note: While CocoaPods itself (the `pod` command) does appear to have some Linux compatibility (I was able to install and run it on my Linux test machine `gem install cocoapods`), it's mostly useful for some server side Swift stuff (?).

With the `pod` command available, it will just delay the eventual error by a little bit:

```
checking whether the C compiler works... no
/usr/bin/bash: line 24: xcrun: command not found
/usr/bin/bash: line 24: xcrun: command not found
~/Library/Caches/CocoaPods/Pods/Release/Flipper-Glog/0.3.6-1dfd6/missing: Unknown `--is-lightweight' option
```

### DIFF
--- a/ern-container-gen-ios/src/IosGenerator.ts
+++ b/ern-container-gen-ios/src/IosGenerator.ts
@@ -82,6 +82,10 @@ export default class IosGenerator implements ContainerGenerator {
   public async fillContainerHull(
     config: ContainerGeneratorConfig
   ): Promise<void> {
+    if (process.platform !== 'darwin') {
+      throw new Error('iOS Container generation is only supported on macOS.')
+    }
+
     const pathSpec = {
       outputDir: config.outDir,
       projectHullDir: path.join(PATH_TO_HULL_DIR, '{.*,*}'),
--- a/system-tests/src/tests/create-container-ios-fixture.js
+++ b/system-tests/src/tests/create-container-ios-fixture.js
@@ -18,12 +18,18 @@ const excludeFilter = [
   .map(s => `**/${s}`)
   .join(',')
 
-run(
-  `create-container --miniapps ${miniapps.join(
-    ' '
-  )} -p ios --out ${process.cwd()}`
-)
-assert(
-  sameDirContent(f.pathToIosContainerFixture, process.cwd(), { excludeFilter }),
-  'Generated IOS Container differ from reference fixture !'
-)
+const command = `create-container --miniapps ${miniapps.join(
+  ' '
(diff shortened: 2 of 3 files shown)


friederbluemle:
Thanks for your review! And yes, I wasn't sure about the location. Spent a few minutes looking around the source code, and then noticed some other checks in the file I put the checks in. But your suggestion makes more sense. I'll move it.

I'm also not a fan at all of running "pod install" automatically.. And if we can, should not create unnecessary coupling. I'm not sure how we can do it in this scenario, but open to other options!

friederbluemle:
Please re-review @belemaire

ern-container-gen-ios/src/IosGenerator.ts
```diff
@@ -82,6 +82,10 @@ export default class IosGenerator implements ContainerGenerator {
   public async fillContainerHull(
     config: ContainerGeneratorConfig
   ): Promise<void> {
+    if (config.targetPlatform === 'ios' && process.platform !== 'darwin') {
```
belemaire, line 85: No need for `config.targetPlatform` check here. This is the iOS container generator implementation so it's explicit that the target platform is iOS ;)

friederbluemle: Haha yeah of course, makes sense :+1:

friederbluemle:
Currently the PR is still failing (it only passes with the fixes from #1585, i.e.  4e18cb1b reverted). Once the underlying issue is addressed (@belemaire is looking into it), I'll rebase.

friederbluemle:
Will rebase after #1592 is merged.

PR #1579, opened 2020-04-28
Fix directory name casing

belemaire:
[Directory name](https://github.com/electrode-io/electrode-native/tree/master/ern-container-gen-ios/src/hull/Config) is `Config` instead of `config`


krunalsshah:
@belemaire the change in the PR for `fillProjectHull` got recently introduced? (checked the blame)
just making sure doesn't impact previous releases.

friederbluemle:
@krunalsshah - It's a recent change that was merged as part of #1542 a few days ago. No releases yet.

PR #1589, opened 2020-04-29
ERNDelegate should be nullable, optional in swift

jw2175:
erndelegate shouldn't be required in swift, this will also surpress null warning on objective c


system-tests/fixtures/ios-container/Podfile.lock
```diff
@@ -15,16 +15,16 @@ PODS:
     - Flipper-Folly (~> 2.1)
     - Flipper-RSocket (~> 1.0)
   - Flipper-DoubleConversion (1.1.7)
-  - Flipper-Folly (2.1.1):
+  - Flipper-Folly (2.2.0):
```
friederbluemle, line 18: Please exclude this file from your PR - I've already put it on the system-test ignore list in #1575 (merge pending).

jw2175: sounds good, thanks

jw2175: i should exclude the proj file as well right?

PR #1600, opened 2020-05-01
Update hulls of api-impl-gen

friederbluemle:
The changes in this PR are limited to `ern-api-impl-gen`.

I've removed _all_ version updates from this PR (initially it was formatting and version updates/androidx migration in one PR). This will make it easier to review and merge. I'll open follow-up PRs with the remaining changes. All generated sources in `ern-api-impl-gen` have consistent formatting now, similar to #1420

#1511


friederbluemle:
Looking at the number of line changes, I think it's better if we separate the version updates/AndroidX migration from the hull/resource updates and improvements.

I'll push an updated branch shortly that will _only_ be concerned about the api-impl-gen hulls and resources. For that we need one preliminary PR that adds a new capability in mustacheUtils (ern-core).

friederbluemle:
I've removed all version updates from the PR. The first draft of if had formatting and version updates/androidx migration in one PR.

Now ready for review and merge! :+1:

PR #1603, opened 2020-05-06
Improve resources handling in android container

belemaire:
Improve resources handing in android container, by getting rid of hardcoded directories, embracing a more dynamic structure, enabling injection of native modules containing android resources. 

**BEFORE**

*Hardcoded resources directories in container hull*

```
sourceSets.main {
  res.srcDirs = [ "src/main/res", "src/main/res/devassist" ]
}
```

*Non flexible, dirty structure with the resources directories of the react native bundle kept in top level directory, with one other directory containing devassist resources*

```
res
├── drawable-mdpi
│   └── node_modules_reactnative_libraries_newappscreen_components_logo.png
└── devassist
    ├── values
    │   └── strings.xml
    └── xml
        └── ern_dev_preferences.xml
```

**AFTER**

*Mustache templated resources directories (dynamic)*

```
sourceSets.main {
  res.srcDirs = [ {{{resSrcDirs}}} ]
}
```

*Automated discovery of all directories contained in res/ top level directory to generate resources directories list, resulting in the following generated code*

```
sourceSets.main {
  res.srcDirs = [ 'src/main/res/bundle', 'src/main/res/devassist', 'src/main/res/react-native-foo' ]
}
```

*A more proper resources structure. Note that the assets specific to the react native bundle have been also moved to a dedicated bundle directory*

```
res
├── bundle
│   └── drawable-mdpi
│       └── node_modules_reactnative_libraries_newappscreen_components_logo.png
├── react-native-foo
│   ├── layout
│   │   └── surface_view.xml
│   ├── layout-v14
│   │   └── texture_view.xml
│   └── values
│       ├── attrs.xml
│       ├── public.xml
│       └── styles.xml
└── devassist
    ├── values
    │   ├── attrs.xml
    │   └── strings.xml
    └── xml
        └── ern_dev_preferences.xml
```


docs/platform-parts/manifest/native-modules.md
```diff
@@ -191,6 +191,31 @@ An array of one or more [Android hardware or software features](https://develope
...
+  }
+]
+```
+
+Also, because resources are accessed via the `R` generated class, you'll have to make sure to update all imports of this class in the source files importing this class, to use the container namespace rather than the native module one. You can do so using `replaceInFile` directive, as illustrated:
+
+```json
+"replaceInFile": [
+  { 
+    "path": "lib/src/main/java/com/org/foo/SourceFile.java",
```
friederbluemle, line 212: Minor: Instead of `com/org/foo/SourceFile` can you use `com/example/SourceFile`? `org` looks like a tld similar to `com`.. The `com.example` namespace/[example.com](https://www.example.com) domain exist specifically for the purpose of examples. ;)

belemaire: Sure, thanks :+1:
Indeed, `org` doesn't have anything to do here ... started from a copy/paste that was a namespace starting with `org` instead of `com`, so it's just a nasty left over. Thanks for catching this one ;)

Issue #1611, opened 2020-05-08
Platform installation using 'ern platform install X' is broken 


LcTwisk:
Hey!

It's currently not possible to install a different version than latest available. 
Running, for example, `ern platform install v0.39.0` does create a folder in `~/.ern/versions` but it does not properly populate the node_modules folder.  Which results in the whole `~/.ern` directory to be discarded on the next run ([here](https://github.com/electrode-io/electrode-native/blob/889074ee62b6899690066ba3c2c9450591970e81/global-cli/src/index.js#L221)) because the `ern-local-cli` module is missing.

PR #1612, opened 2020-05-08
Fix platform installation

LcTwisk:
This fixes: https://github.com/electrode-io/electrode-native/issues/1611

**Root cause**
The `yarn add` command on [this](https://github.com/electrode-io/electrode-native/blob/889074ee62b6899690066ba3c2c9450591970e81/ern-core/src/Platform.ts#L118) line fails silently (but the `ern platform install` operation is marked as successful) because there's no `package.json` in the current folder.  

**Solution**
The initial platform version (the latest) is fetched correctly because it's an entirely different code path ([here](https://github.com/electrode-io/electrode-native/blob/889074ee62b6899690066ba3c2c9450591970e81/global-cli/src/index.js#L85)). For now I've just copied the relevant line but it might be even nicer to unify the platform installation code paths. However, I couldn't figure out what the best place would be. 
I thought of changing the implementation in `Platform.ts`, and import it in `global-cli/src/index.js` but `ern-core` isn't a dependency of `global-cli` at the moment.  Would it be OK to add this dependency?

Apart from this change, there are more differences between the platform-install implementations in `ern-core` and `global-cli`. The one in `global-cli` has some extra logic to support Windows machines, which the implementation in `Platform.ts` doesn't have. This probably means that the `ern platform install` command will still be broken for Windows users after this change is merged...

### DIFF
--- a/ern-core/src/Platform.ts
+++ b/ern-core/src/Platform.ts
@@ -110,10 +110,11 @@ export default class Platform {
     const pathToVersion = this.getRootPlatformVersionPath(version)
 
     try {
-      shell.mkdir('-p', path.join(pathToVersion, 'node_modules'))
+      shell.mkdir(pathToVersion)
       shell.pushd(pathToVersion)
       if (this.isYarnInstalled()) {
         // Favor yarn if it is installed as it will greatly speed up install
+        execSync(`yarn init --yes`, { cwd: pathToVersion })
         execSync(
           `yarn add ${ERN_LOCAL_CLI_PACKAGE}@${version} --exact --ignore-engines`,
           {
@@ -121,6 +122,7 @@ export default class Platform {
           }
         )
       } else {
+        execSync(`npm init --yes`, { cwd: pathToVersion })
         execSync(`npm install ${ERN_LOCAL_CLI_PACKAGE}@${version} --exact`, {
           cwd: pathToVersion,
         })
--- a/ern-core/test/Platform-test.ts
+++ b/ern-core/test/Platform-test.ts
@@ -218,11 +218,19 @@ describe('Platform', () => {
       expect(() => Platform.installPlatform('6.0.0')).to.throw
     })
 
-    it('should create complete directory hierarchy', () => {
+    it('should use yarn init to create a package.json', () => {
       Platform.installPlatform('3.0.0')
-      expect(
-        fs.existsSync(path.join(ernHomePath, 'versions/3.0.0/node_modules'))
-      ).true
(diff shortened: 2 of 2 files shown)


belemaire:
Thanks @LcTwisk :+1:

There is indeed a problem here in some context, that's a nice catch. If there is a `package.json` or `node_modules` somewhere in the parent tree, then it will run `yarn add` or `npm install` at this level. That's how `yarn add` / `npm install` work, if they don't find a package.json / node_modules in working directory they will crawl up the parent directories until they find one. If they find one they will add the package in this directory rather than the one from which the command was run. If they don't find one they will create the package.json/node_modules in current working directory.

There is a low probability of this happening when using default ERN_HOME (`~/.ern`) as it means that for it to fail this way, a package.json need to be present either in `~/` which is not a good thing. But if using a different ERN_HOME the probability is higher to have a package.json up the parent directory tree.

There is a more explicit way to make it work than manually creating a package.json / node_modules.
You could get rid of these two lines
 https://github.com/electrode-io/electrode-native/blob/057f69c21f955719e43db37c738ee6bd9626dbbb/ern-core/src/Platform.ts#L113-L114

And instead replace them by running  `yarn init --yes` / `npm init --yes` in the respective blocks before running `yarn add` / `npm install`. These two commands will create a default package.json automatically.

Would you be OK to update the code this way ?

Will go through your other comments later on today.

LcTwisk:
Sounds good to me! Would the same apply for: https://github.com/electrode-io/electrode-native/blob/057f69c21f955719e43db37c738ee6bd9626dbbb/global-cli/src/index.js#L84-L85 ?

belemaire:
Yes, would be good to do it there as well :)

LcTwisk:
Will keep the last one out of scope. The `global-cli` code doesn't seem to be called when running in dev mode, which makes it harder to verify if it works properly. Furthermore, I don't have access to a Windows machine so it would again be hard to verify.

belemaire:
Sure, that works :+1:
I do have access to Windows machine, so will do it as a follow up PR. Will also validate this PR on Windows.

friederbluemle:
Nice, thanks for your contributions @LcTwisk 🎉 

Can you quickly rebase your branch please?

Btw, I tried this on my Windows test PC, and it works! :+1:

LcTwisk:
Done!

friederbluemle:
The only thing I noticed is that `isYarnInstalled()` appears to have issue on Windows (returning false even if Yarn is installed). This results in the installer using npm. Not a huge issue and not in scope for this PR.

PR #1613, opened 2020-05-08
Fix baseComposite option for start command

LcTwisk:
After https://github.com/electrode-io/electrode-native/pull/1532 and https://github.com/electrode-io/electrode-native/pull/1544 I found one more place where the `baseComposite` wasn't wrapped into a `PackagePath`


belemaire:
@LcTwisk 
FYI we're preparing a patch release `0.41.2` that will be shipped later on today. 
This fix will be included part of it.

PR #1619, opened 2020-05-12
fix memory leak on registering UIFont

liulianci:
fix memory leak on registering UIFont


ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.m
```diff
@@ -285,6 +285,7 @@ - (void)loadCustomFonts {
                 CFStringRef errorDescription = CFErrorCopyDescription(error);
                 NSLog(@"Failed to load font: %@", errorDescription);
                 CFRelease(errorDescription);
+                CFRelease(error);
```
friederbluemle, line 288: Just checking: If `CTFontManagerRegisterGraphicsFont` returns true, then this line will never be reached. Should it be moved outside (below) the `if` block?

liulianci: if `CTFontManagerRegisterGraphicsFont` returns true, error would be nil we don't need to release it

PR #1620, opened 2020-05-13
Add build step and more platforms to CI

friederbluemle:
Extend GitHub Actions CI matrix with additional platforms, and also run `yarn build`.

Instead of just `ubuntu-latest`, with this we're now also running build+unit tests on `macos-latest` and `windows-latest`. Node.js matrix [10, 12, 14] we'll keep on Linux only for now, macOS and Windows are using Node.js 12.

`yarn check:lint` only runs on Linux/Node.js 12.

The new config will yield 5 different jobs, all running in parallel.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,16 +2,29 @@ name: ci
 on: [push, pull_request]
 jobs:
   build:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        node-version: [10, 12, 14]
+        node: [12]
+        os: [macos-latest, ubuntu-latest, windows-latest]
+        include:
+          - node: 10
+            os: ubuntu-latest
+          - node: 14
+            os: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - name: Use Node.js ${{ matrix.node-version }}
+      - name: Use Node.js ${{ matrix.node }}
         uses: actions/setup-node@v1
         with:
-          node-version: ${{ matrix.node-version }}
+          node-version: ${{ matrix.node }}
+      - run: yarn
+      - run: yarn build
+      - run: yarn test:unit
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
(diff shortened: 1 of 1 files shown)


friederbluemle:
All good questions :+1: And I thought about each one of them already, was just too lazy to put it in the PR description. 😄 

> Not sure why you chose Node 12 as the base version for Windows / MacOS. Sure it is the current LTS one, but given that current minimal supported node version for ern is 10, I think it would make more sense to keep this one as the base one for WIndows/MacOS (and keep 12/14 additional on Linux for now).

OK, I don't have strong objections to changing to 10 as the base... What I thought is that most people would probably run LTS at this point (12), if they're not already on the latest (14). Having one job to ensure backwards compatibility may be enough, while at the same time making use of other improvements to Node.js for the rest?

> If we don't go full matrix yet (10/12/14 on all platforms), wouldn't it make more sense to run with 10/12/14 on MacOS rather than Linux, given that it is the predominant OS used to run ern ?

Yeah, I briefly thought about "full matrix", but would prefer to avoid the explosion of jobs and unnecessary resource consumption doing the same exact thing without tangible benefits.
Regarding the "main" platform, I think we should stay on Linux as the main OS for **unit** tests. First, there should not be any difference in the first place, and also Linux is stricter when it comes to path names and capitalization. Unfortunately, as you know, macOS by default uses a non case-sensitive file system, which means some spelling issues might fall through the cracks and then cause issues on case-sensitive file systems. Running the tests on Linux will avoid that at PR level.
Now, for **e2e** tests, of course they will be run on macOS as the primary OS (since we require xcrun, xcodebuild, CocoaPods etc etc), and we can only run a subset on other platforms.

> Not a huge fan of mixing lint checks in just one of the jobs. While I agree that it would be dumb to run it in all jobs, it would be better to have it as a dedicated separate job only running lint check, for clarity. Unsure though if it would be straightforward to achieve or not, if not we can keep it for a follow up PR.

Agree, I also thought it looks kind of hacky... What I wanted to avoid it an additional job (along with a bunch of duplication). After making the change, it looks a bit cleaner I have to admit.. Check the updated diff and let me know what you think.

And yes, we should sunset the Azure unit tests soon after merging this one. I would give this new set up at least a few days to proof itself as stable.

PR #1622, opened 2020-05-14
Update Mustache and add support for partials

friederbluemle:
### Commit 1: `Update mustache to 4.0.1`

Despite being a "major" version upgrade, [the changelog](https://github.com/janl/mustache.js/blob/master/CHANGELOG.md) mainly lists internal improvements. There aren't any breaking api changes, at least the ones we're using. They also mention

> Majority of using projects don't have to worry by this being a new major version.

Overall, I felt a slight (maybe subjective) performance improvements locally. There are no changes to any of the generated fixtures (as expected).

The second one is the more interesting:

### Commit 2: `Add support for mustache partials`

[Partials](https://github.com/janl/mustache.js/#partials) are now supported in the two exported methods of `mustacheUtils`.

This also lifts some unnecessary restrictions that were previously enforced in `ern-api-impl-gen`: It was processing the resource directory under the assumption that _all_ `*.mustache` files have to have a mapping to a file in the output directory. That is obviously not applicable for partials. I removed throwing an Error, and instead only render the mustache file into the output directory if there is a mapping. Changed the `warn` log to `info` if there is no mapping.

We will need this new capability for the hull improvements in #1600

### DIFF
--- a/ern-api-gen/package.json
+++ b/ern-api-gen/package.json
@@ -44,7 +44,7 @@
     "lodash": "^4.17.14",
     "minimatch": "^3.0.4",
     "mkdirp": "^0.5.1",
-    "mustache": "^2.3.1",
+    "mustache": "^4.0.1",
     "semver": "^5.5.0",
     "shelljs": "^0.8.2",
     "swagger-methods": "^1.0.4",
--- a/ern-api-gen/src/java/MustacheWriter.ts
+++ b/ern-api-gen/src/java/MustacheWriter.ts
@@ -2,6 +2,7 @@ import Mustache from 'mustache'
 import { isIterable } from './isIterable'
 
 export class MustacheWriter extends Mustache.Writer {
+  // @ts-ignore
   public renderSection(token, context, partials, originalTemplate) {
     let buffer = ''
     let value = context.lookup(token[1])
--- a/ern-api-impl-gen/package.json
+++ b/ern-api-impl-gen/package.json
@@ -44,7 +44,7 @@
     "ern-core": "1000.0.0",
     "fs-extra": "^8.1.0",
     "lodash": "^4.17.14",
-    "mustache": "^2.3.1",
+    "mustache": "^4.0.1",
     "semver": "5.5.0",
     "xcode": "^1.0.0",
     "xcode-ern": "^1.0.12"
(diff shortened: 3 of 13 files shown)


friederbluemle:
Yikes.. It looks like ern-api-gen is having a build issue:

```
src/java/MustacheWriter.ts(5,10): error TS2416: Property 'renderSection' in type 'MustacheWriter' is not assignable to the same property in base type 'MustacheWriter'.
  Type '(token: any, context: any, partials: any, originalTemplate: any) => string | undefined' is not assignable to type '(token: string[], context: MustacheContext, partials?: Record<string, string> | PartialLookupFn | undefined, originalTemplate?: string | undefined) => string'.
    Type 'string | undefined' is not assignable to type 'string'.
      Type 'undefined' is not assignable to type 'string'.
```

friederbluemle:
So weird.. there was a duplicated reimplementation of `MustacheWriter` in `ern-api-gen`. It was causing the TypeScript error. I removed in together with the one method that we were not even using. 🤦 😄

belemaire:
Yes unfortunately `ern-api-gen` module is not easy to work with / maintain. 
It is only using TypeScript on the surface but is really old school Javascripty under the hood, with things that are dynamically typed / used, so getting rid of something might look ok but in reality it's used somewhere even if not referred to directly, making it quite difficult to trace/follow the code by just looking at it. Will have a closer look at this PR later on today or tomorrow morning.

friederbluemle:
OK thanks for the added info.. I will add the file back. After some more investigation, it turns out that the file is a result of the transpiled Java code from the jmustache implementation that adds support for the special [`-first and -last`](https://github.com/samskivert/jmustache#-first-and--last) variables (something that is not available in vanilla Mustache). We're using these variables in many places, so we definitely need them.

friederbluemle:
OK the easiest solution that only involves a single line `// @ts-ignore` is very much preferred here I think..

There is a *lot* of external code in ern-api-gen that has been transpiled/rewritten multiple times, making it difficult to trace where is originated from. The customized `renderSection` method in `MustacheWriter` is a partial copy from https://github.com/janl/mustache.js ([mustache.js#L581](https://github.com/janl/mustache.js/blob/master/mustache.js#L581))

friederbluemle:
Yes, already did that, and also mentioned it in the PR description ;)

> There are no changes to any of the generated fixtures (as expected).

PR #1623, opened 2020-05-14
Fix an issue with local miniapps on RN<60

belemaire:
Some users reported failures with ERN 0.41 when trying to generate containers with one or more local miniapp(s) running a version of React Native < 0.60, leading to bundling errors such as `'Unable to resolve module [...]`.

This is due to the new mechanism introduced in ERN 0.41 to speedup composite generation when dealing with local MiniApp(s), only installing the dependencies that are required to be found at the root of the composite, react wasn't one of them (at least for RN >= 0.60), so ern is not installing it but instead pointing to its location in a local miniapp directory instead (to reduce generation time). 
For some reason, (probably haste related), with versions of React Native < 0.60.0, react dependency need to be present in the root bundling directory and cannot be located in a different location.

The fix here is simply to add react dependency to the composite in case we're dealing with one or more local miniapp(s) using a version of RN < 0.60.0.

### DIFF
--- a/ern-composite-gen/src/generateComposite.ts
+++ b/ern-composite-gen/src/generateComposite.ts
@@ -174,11 +174,21 @@ async function generateFullComposite(
       // To know the version to install, we will just have a peak to one of
       // the miniapps, given that react native version is aligned across all.
       const pJson = await readPackageJson(localMiniAppsPaths[0])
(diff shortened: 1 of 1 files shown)


friederbluemle:
Is there any to not _always_ add the required `react` peer dependency?

friederbluemle:
I've seen some warning in the past about missing peer dependency

Issue #1489, opened 2020-02-17
Hangs on "Waiting for device to start" on `ern run-android`


AndrewSouthpaw:
Repro

1. Without an Android simulator open, run `ern run-android` on MiniApp
1. Eventually it'll hang, prompting "Waiting for device to start"

For beginners, this line is not obvious: it wants the user to take action, going to Android Studio and turning on the Android simulator through the AVD. If it's not possible to programmatically launch an Android simulator, it would be nice to at least update the text.

I searched for the string across GitHub and didn't find where the logic is living, but if you want to point me to where the CLI prompts this I'd be happy to offer some copy in a PR.

friederbluemle:
I _think_ this issue is about the case where _no_ avds (Android Virtual Devices, i.e. emulators) are _configured_ on the host machine.

I'm not able to repro this as stated in the issue description, at least with the latest version of ERN, and I don't think much has changed for several versions.

- **with** emulators configured (i.e. non-empty `~/.android/avd`)

  ```
  $ ern run-android
  ...

  ? Choose Android device image
  ❯ Nexus_5X_API_29_x86
    Pixel_2_API_21
    Pixel_3_API_29
    Polestar_2_API_28
  ```

- **without** emulators configured (i.e. empty `~/.android/avd`)

  ```
  $ ern run-android
  ...

  ? Choose Android device image (Use arrow keys)
  ❯
  ```

In both cases it is waiting for input. A problem occurs in the second case (empty list), _after_ selecting the empty entry. And yes, that's definitely something ERN should handle better. Instead of displaying an empty list and then waiting indefinitely, how about a simple error message that makes it clear that it's not possible to run an Android app:

```
Unable to launch app: No devices connected and no emulators configured.
```

An emulator is set up _by default_ during the installation of Android Studio/the Android SDK (which is already a requirement of ERN).

friederbluemle:
I just checked how React Native handles the situation, and it's similar to what I suggested above:

```
$ yarn android
...
info Launching emulator...
error Failed to launch emulator. Reason: No emulators found as an output of `emulator -list-avds`.
warn Please launch an emulator manually or connect a device. Otherwise app may fail to launch.

[building app]

error Failed to install the app. Make sure you have an Android emulator running or a device connected. Run CLI with --verbose flag for more details.
```

krunalsshah:
resolved https://github.com/electrode-io/electrode-native/pull/1628

PR #1628, opened 2020-05-20
throw error if emulator fails to start

krunalsshah:
Fixes#https://github.com/electrode-io/electrode-native/issues/1489

### DIFF
--- a/ern-core/src/android.ts
+++ b/ern-core/src/android.ts
@@ -24,6 +24,9 @@ export const DEFAULT_SUPPORT_LIBRARY_VERSION = '28.0.0'
 export const DEFAULT_TARGET_SDK_VERSION = '28'
 export const DEFAULT_SOURCE_COMPATIBILITY = 'VERSION_1_8'
 export const DEFAULT_TARGET_COMPATIBILITY = 'VERSION_1_8'
+const ANDROID_DEVICE_INFO = `
+https://developer.android.com/studio/run/emulator-commandline
+https://developer.android.com/studio/run/emulator`
 
 export interface AndroidResolvedVersions {
   androidGradlePlugin: string
@@ -71,12 +74,24 @@ export function resolveAndroidVersions({
 // Misc utilities
 // ==============================================================================
 
-//
 // Returns a promise that will get resolved after a given delay (in ms)
 async function delay(ms: number) {
   return new Promise((resolve, reject) => {
     setTimeout(() => {
       resolve()
+    }, ms)
+  })
+}
+
+// Returns a promise that will get rejected after a given timeout (in ms)
+async function failAfterTimeOut(ms: number) {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      reject(
+        new Error(
+          `Failed to collect the device/emulator state.${ANDROID_DEVICE_INFO}`
+        )
+      )
     }, ms)
   })
 }
@@ -188,10 +203,9 @@ export async function runAndroid({
 export async function askUserToSelectAvdEmulator(): Promise<string> {
   const avdImageNames = await getAndroidAvds()
   if (Array.isArray(avdImageNames) && avdImageNames.length === 0) {
-    throw new Error(`No Emulator or device found.
-    Install the Android Emulator or connect the device to proceed.
-    https://developer.android.com/studio/run/emulator-commandline
-    https://developer.android.com/studio/run/emulator`)
+    throw new Error(
+      `No Emulator or device found. Launch an emulator manually or connect a device.${ANDROID_DEVICE_INFO}`
+    )
   }
   const deviceConfig = ernConfig.get(deviceConfigUtil.ANDROID_DEVICE_CONFIG)
   // Check if user has set the usePreviousEmulator flag to true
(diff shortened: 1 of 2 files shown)


ern-core/src/android.ts
```diff
@@ -81,6 +83,20 @@ async function delay(ms: number) {
...
 }
 
+// Returns a promise that will get rejected after a given timeout (in ms)
+async function failAfterTimeOut(ms: number) {
+  return new Promise((resolve, reject) => {
+    setTimeout(() => {
+      reject(
+        new Error(
+          `Failed to collect the device/emulator state
+        ${ANDROID_DEVICE_INFO}`
```
friederbluemle, line 93: Because the new `ANDROID_DEVICE_INFO` constant is a template string that _begins_ with `LF`, this error will now have 8 trailing spaces on the second line (and nothing else). Similar issue on line 209 below.
Can you either remove the remove the linefeed from the template string, or (maybe better) remove the indentation if this variable here?

ern-core/src/android.ts
```diff
@@ -248,8 +263,9 @@ export async function runAndroidUsingAvdImage({
 }) {
   // https://issuetracker.google.com/issues/37137213
   spawnp(androidEmulatorPath(), ['-avd', avdImageName], { detached: true })
-
-  await kax.task('Waiting for device to start').run(waitForAndroidDevice())
+  await kax
+    .task('Waiting for device to start')
+    .run(Promise.race([waitForAndroidDevice(), failAfterTimeOut(60000)]))
```
friederbluemle, line 268: Not 100% sure how it works, but what if for instance on a very slow computer, the emulator takes >60s to start up?

ern-core/src/android.ts
krunalsshah: @friederbluemle here's the error 
```
✖ An error occurred: Failed to collect the device/emulator state
✖
✖ https://developer.android.com/studio/run/emulator-commandline
✖ https://developer.android.com/studio/run/emulator
```

ern-core/src/android.ts
krunalsshah: The goal is to not wait forever, Couldn't find the default value set for the timeout in the Android source. except one can override value https://google.github.io/android-gradle-dsl/current/com.android.build.gradle.internal.dsl.AdbOptions.html

Is there any other value you'd recommend?

krunalsshah:
@friederbluemle 
Output
```
✖ An error occurred: No Emulator or device found.Launch an emulator manually or connect a device.
✖ https://developer.android.com/studio/run/emulator-commandline
✖ https://developer.android.com/studio/run/emulator
```
```
✖ An error occurred: Failed to collect the device/emulator state.
✖ https://developer.android.com/studio/run/emulator-commandline
✖ https://developer.android.com/studio/run/emulator
```

ern-core/src/android.ts
friederbluemle: I don't really have a value I would recommend.. So I guess I'm OK with 60s ;)
Can you also please check the comments I left on #1489, since they're somewhat related. If you agree, see if you can repurpose the PR. Or otherwise, we can merge it and we'll address it as a follow-up.

On a more general level, the question is, do we really need code in ERN that waits for an emulator to start? In my opinion, we should ideally **remove this kind of complexity** that is not a concern of ERN. React Native also does not care about it, and I have not seen anyone complain about it. Is there a strong reason we need to?

krunalsshah: I read comments #1489 I believe the PR is in line with the problem reported.
At the moment we surely have the logic to wait for the AVD indefinite time if `init.svc.bootanim` doesn't reach the `stopped` state.

Secondly, if there is no emulator configured we should fail with the proper message, the logic was missing and the PR fixes it. 

I tried, 
- No configured AVD (fails with the proper message)
- The emulator doesn't reach the state `stopped` (fails with the proper error message)
- The happy path where the mini-app runner

friederbluemle: It does now, yes. `a0d0cd5` didn't have the change in `getAndroidAvds` yet. ;) :+1:

ern-core/src/android.ts
```diff
@@ -290,7 +308,7 @@ export async function installAndLaunchApp({
 
 // Utility method that basically completes whenever the android device is ready
 // It check device readiness every 2 sec (poll way)
-export async function waitForAndroidDevice() {
+export async function waitForAndroidDevice(): Promise<void> {
```
friederbluemle, line 311: I'm a JS/TS noobie, but doesn't a function declared as `async` automatically wrap the return inside of a Promise? Is there a reason you explicitly added the return type here?

belemaire: Yeah, we don't have strong coding style guidelines for this one, and nothing enforced by our linter regarding function return types declaration. Given that, if not enforced through clear coding style guide (and potentially linter), this is subject to individual appreciation. You will probably find some of these in the code base, whereas in other locations we don't explicitly state the return type.

One could say that we must explicitly declare all return types so that it makes it clear to developers as to what the function is returning if they are browsing the code from an IDE that don't have TS integration or just from a text editor. One could say that we need to explicitly declare all return types as long as they are not `void` / `Promise<void>`. One could say that we don't explicitly declare return types no matter what to reduce verbosity at the cost of clarity.

So yes, I'll leave this one to the appreciation of @krunalsshah for now until we eventually decide of stronger coding standards that we ideally enforce automatically through linter and apply it throughout the code base.

friederbluemle: Agreed. My main point and question is: Why add it to this particular function? Nothing else in the function signature or body changed. I'm aware the linter does not currently check this, but we should try to be consistent at least on a file level. If this is added here (I don't have a strong preference for or against it at this point), then maybe it should also be added to all other functions in the file? Otherwise there's no reason to modify this line.

krunalsshah: reverted the TS changes.

friederbluemle: Thanks @krunalsshah :+1:

PR #1629, opened 2020-05-21
Update Android container hull

friederbluemle:
The changes in this PR are limited to `ern-container-gen-android`.

While there are many line changes, _most_ of them are formatting related. However, there are also a couple of important fixes and minor functional improvements.

`AndroidGenerator` now supports **mustache partials** as well (commit 1). This is used when updating the hull in the following commit. At the same time, removed some chatty logs (avoiding many dozen `Skipping mustaching of...` printed every time a container is generated). We will now also **no longer** see the following slightly misleading and inapplicable warning printed to console:

```
⚠ Skipping react-native as it does not have an Android configuration
```

We're directly setting the **Android Gradle plugin** version to `3.5.2` for now. We can consider going back to a dynamic injection once the rest of the system has been updated.

The biggest, and probably most important fix in this PR is related to the **Gradle Wrapper** we were using in the container:
According to the wrapper jar's sha256 sum, it was file from **Gradle 1.8**!! (2013) 😱 The updated jar now matches the Gradle Wrapper shipped with `5.3` through (currently) `5.6.4`. Our currently used Gradle Wrapper version `5.4.1` falls within that range. We're using `5.4.1` because that's the version the Android Gradle plugin `3.5.2` uses by default. We use `3.5.2` because that's the version React Native `0.62.2` uses.
Since the jar was pretty old, it's likely the wrapper scripts were from a similar generation as well. Amongst other things, this means the `gradlew` script might have not been properly forwarding environment variable to the Gradle process. The scripts now also match Gradle 5.3+

I will squash commit 2 and 4 before the merge.

#1511


friederbluemle:
> Is that because bumping to `3.5.2` for other hulls (api-impl-gen, runner) doesn't work with these hulls in their current state ?
> You are mentioning considering going back to dynamic injection once the rest of the system has been updated. But once this PR is merged, the [documentation here](https://github.com/electrode-io/electrode-native/blob/c6be4730926296bbcab7b0dce58f618c5c57e03b/docs/platform-parts/container/container-integration.md#configure-android-build-configuration-versions) regarding dynamic support for android gradle plugin version won't be valid anymore.
> 
> Maybe you could just remove this from the documentation part of this PR. We can always reintroduce it in the documentation once we go back to dynamic (if we do). The problem with keeping it in the doc would be if we forget about this and release a new ern version with this PR, the documentation will be invalid. Even without releasing a new version, just merging this PR will render the documentation invalid on master (the documentation should be up to date with the state of master if someone is using ern from master).

Yes, that's precisely the reason. I initially considered updating `DEFAULT_ANDROID_GRADLE_PLUGIN_VERSION`, but it will affect other parts of the system in ways that are totally out of scope of this PR.
The only other solution would be to update _everything_ in one huge commit/PR. I think the risk would be too high and it would not help with the code review at all. Also in case there are any issues, it would make it more difficult to figure out what the exact problem is. That's why I suggest we take these fixes at smaller, more manageable chunks.

As explained, once we have _everything_ in a consistent state, _only then_ we may consider going back to providing more cross-project dynamic configuration options. The primary objective now has to be to gradually fix many of the inconsistencies that have accumulated over the years that are causing more and more problems.
Overall, we need to seriously rethink the effort/reward tradeoffs of some of the system features such as these "dynamic" configuration options. Yes, there are areas where it makes sense. But there are others where it doesn't really make sense, and it simply not possible to "update" a version like that. In some areas it's even dangerous and will lead to problems this very PR is going to fix.

Now regarding the documentation, you are correct: It would temporarily cause the part about "support for dynamic Gradle plugin" to be out-of-sync with master. I'm not saying this is good, and I will guarantee you that I will _not_ "forget" about it when migrating/updating the docs. However, if you think this will be a problem for someone trying to make use of this feature (I am not aware of a single real world example that is using it), then I'm happy to remove that part of the documentation. :) I don't have a strong preference, and I know, the removed docs will still be part of Git history, but the reason I'm slightly leaning towards keeping the docs as is for now is that when doing a big doc update/rewrite, it's most helpful to have the big picture (what was there before/what do we have now) right there in front of you. Let me know how you want me to proceed!

belemaire:
It's OK, let's just keep it in the documentation on master.
Just need to make sure to keep this in mind before releasing 0.42.0 (i.e making sure that we're back to dynamic and don't need to do any change in the doc, or we're not and remove this part of the doc prior to release).
I just faced a similar case with Bugsnag, have to make sure of something prior to releasing v0.42.
I'm sure we will run into more of these things where we don't have time to tackle things immediately or need to have a second eye prior before release.
Maybe you could just use a specific label, for things not to forget about prior to release, that we put on issues / pr that need to be double checked prior to release and we clear labels once addressed ?

friederbluemle:
Nevermind, I've updated the docs.. I don't want this to be a blocker for these important fixes..

PR #1634, opened 2020-05-22
Process all resource files with mustache

friederbluemle:
No longer special case "common classes" from mustache templates. Process all files in the resource directory with mustache. This results in less code and complexity, and allows for even more simplifications in the upcoming #1600

Note that there are zero changes in the test fixtures (as expected, this change does not affect the generated files).


krunalsshah:
@friederbluemle we need to regen the system test fixtures, before merging the change?

PR #1642, opened 2020-05-27
Fix composite issue impacting react hooks

belemaire:
Keep react in composite project root no matter what.
Electrode Native was only installing react in project root for versions of react native <= 0.60.0 as it is needed to have react in the root for these older versions of react native.
But we recently noticed an issue that is causing MiniApps using [react hooks](https://reactjs.org/docs/hooks-intro.html) to fail. After investigating the issue, it appears that react has to be kept in the project root for these guys to properly work.
This PR is also making sure to add the react dependencies kept inside the local MiniApps node modules, to the blacklist of metro config, to avoid running into haste duplication issues.

### DIFF
--- a/ern-composite-gen/src/generateComposite.ts
+++ b/ern-composite-gen/src/generateComposite.ts
@@ -179,25 +179,23 @@ async function generateFullComposite(
       extraJsDependencies.push(
         PackagePath.fromString(`react-native@${miniAppRnVersion}`)
       )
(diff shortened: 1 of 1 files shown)


adiwon9555:
@belemaire I am using ern version 0.48.3, the latest one.. and I am still getting `Invalid hook call` error, when used `ern start`.
But `npm start` is working fine in mini apps.
Note: I have already done `ern link`

Any idea?

PR #1643, opened 2020-05-27
Improve initial installation process

friederbluemle:
This simplifies the initial installation process with two small but significant improvements:

### Support for `ERN_INITIAL_VERSION` 

Now it is possible to override the initial version of ERN (i.e. `ern-local-cli`) that is downloaded on a new system (no previous `.ern` folder) using the environment variable `ERN_INITIAL_VERSION`. This is especially useful on CI systems and/or in combination with specifying a custom `ERN_HOME`.

### Remove the requirement to run `ern` once

Previously, we had the rather unusual requirement to run `ern` (without arguments) at least once after globally installing `electrode-native`:

```
npm install -g electrode-native && ern
```

This is no longer necessary. Any arguments passed to the first invocation of `ern` are now automatically passed to `ern-local-cli` _after_ the initial install. This reduces first time use friction and standardizes usage/installation with many other Node.js command line tools.

Also simplified and enhanced the `setup-dev.js` script. It now automatically creates initial directories/.ernrc if necessary, as well as activates `1000.0.0` (unless a different version has already been set).

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -22,7 +22,7 @@ If you want to contribute code to the Electrode Native platform, you'll first ne
 1. If you have not already installed the platform, install by running the following command:
 
     ```bash
-    npm install -g electrode-native && ern
+    npm install -g electrode-native
     ```
 
 1. Fork and clone the `electrode-native` repository.
--- a/README.md
+++ b/README.md
@@ -21,16 +21,16 @@
 
 ### Prerequisites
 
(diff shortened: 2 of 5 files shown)


belemaire:
wow that's great 🎉 
will have a closer look at it today 😃

friederbluemle:
Tested this (including the changes in #1645) on Linux, macOS and Windows (PowerShell and WSL). Everything looks good, and performance seems slightly faster.

PR #1644, opened 2020-05-27
Update .gitattributes

belemaire:
Update `.gitattributes` file to exclude [ern-api-gen/resources](https://github.com/electrode-io/electrode-native/tree/master/ern-api-gen/resources) directory from GitHub repository language statistics. This directory mostly contains mustache templates showing up as HTML in our repo language statistic.

Screen Shot 2020-05-27 at 10 52 07 AM

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,4 +1,2 @@
-system-tests/fixtures/* linguist-vendored=true
-ern-api-gen/test/fixtures/* linguist-vendored=true
-ern-runner-gen-android/test/fixtures/* linguist-vendored=true
-ern-runner-gen-ios/test/fixtures/* linguist-vendored=true
+**/fixtures/* linguist-vendored
+ern-api-gen/resources/* linguist-vendored


friederbluemle:
Since you're updating the file, do you also want to change `linguist-vendored=true` to just `linguist-vendored`? I think we don't need the `=true` unless we're using `=false` to exclude a broader scope.

https://github.com/github/linguist#vendored-code

Also - How about excluding all _fixtures_, which would reduce the `.gitattributes` file to two lines:

```
**/fixtures/* linguist-vendored
ern-api-gen/resources/* linguist-vendored
```

Currently, we have several fixtures still contributing to the language stats, e.g.

```
ern-cauldron-api/test/fixtures
ern-composite-gen/test/fixtures
ern-container-gen/test/fixtures
```

https://stackoverflow.com/questions/39060111/does-github-linguist-support-prefix-wildcards-for-linguist-vendored

belemaire:
That's a good call, let me update the PR accordingly, we will see how it goes ;)

friederbluemle:
Only a small change, but still better :+1:

*Before*

```
44.49%  TypeScript
27.16%  Java
8.71%   Objective-C
7.68%   HTML
6.01%   JavaScript
5.01%   Swift
0.75%   Shell
0.19%   Ruby
```

*After*

```
44.57%  TypeScript
27.21%  Java
8.73%   Objective-C
7.69%   HTML
5.84%   JavaScript
5.02%   Swift
0.75%   Shell
0.19%   Ruby
```

friederbluemle:
Wait.. I saw your screenshot above. Where/when was that taken?

friederbluemle:
image

Yeah, much better now!

belemaire:
Yes it's taking a little while for GitHub to update the stats after pushing a new `.gitattributes`.
Looks good now :)

PR #1656, opened 2020-06-03
Fix azure system tests

friederbluemle:
Forgot to take out these steps 🤦 

With 1.1.0 of global-cli it is no longer necessary to explicitly install the local cli. `ern` without arguments will now _always_ return 1 (including the very first time it is called).

### DIFF
--- a/azure/templates/system-tests.yml
+++ b/azure/templates/system-tests.yml
@@ -8,17 +8,11 @@ steps:
       git config --global user.name "Electrode Native"
     displayName: 'Set git user'
   - script: |
-      npm install -g electrode-native;
+      npm install -g electrode-native
     displayName: 'Install ern global CLI'
-  - script: |
-      ern
-    displayName: 'Install ern local CLI'
   - script: |
       node setup-dev
     displayName: 'Setup dev version of ern'
-  - script: |
-      ern platform use 1000.0.0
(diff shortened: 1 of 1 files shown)


azure/templates/system-tests.yml
```diff
@@ -8,17 +8,11 @@ steps:
...
+      npm install -g electrode-native
     displayName: 'Install ern global CLI'
-  - script: |
-      ern
-    displayName: 'Install ern local CLI'
   - script: |
       node setup-dev
     displayName: 'Setup dev version of ern'
-  - script: |
-      ern platform use 1000.0.0
```
belemaire, line 20: Sorry, might have missed something, but this is not needed anymore to switch to local dev ern version ?

friederbluemle: Good question - No, this is no longer necessary, because `setup-dev` will take care of it (as briefly described in #1643).

friederbluemle: See this section: https://github.com/electrode-io/electrode-native/pull/1643/files#diff-e0435c93383b7750e22c48846188c7b0R22

belemaire: :+1: Yes remember seeing something like it but couldn't find it back. Thanks :)

PR #1659, opened 2020-06-08
Add support for http_proxy env variables

friederbluemle:
With this small change, the `createProxyAgentFromErnConfig` function now supports the `http_proxy` and `https_proxy` environment variables in addition to the existing support for config keys from the ern config (currently `bundleStoreProxy`, `bugsnagProxy`, or `sourceMapStoreProxy`).

The config keys still take precedence, so if one is set, this PR won't affect the existing behavior.

### DIFF
--- a/ern-core/src/createProxyAgent.ts
+++ b/ern-core/src/createProxyAgent.ts
@@ -40,5 +40,9 @@ export function createProxyAgentFromErnConfig(
   const proxyUrl = config.get(configKey)
   if (proxyUrl) {
     return createProxyAgentFromUrl(new url.URL(proxyUrl), { https })
+  } else if (process.env.http_proxy) {
+    return createProxyAgentFromUrl(new url.URL(process.env.http_proxy), { https })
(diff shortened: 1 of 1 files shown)


ern-core/src/createProxyAgent.ts
```diff
@@ -40,5 +40,9 @@ export function createProxyAgentFromErnConfig(
   const proxyUrl = config.get(configKey)
   if (proxyUrl) {
     return createProxyAgentFromUrl(new url.URL(proxyUrl), { https })
+  } else if (process.env.http_proxy) {
+    return createProxyAgentFromUrl(new url.URL(process.env.http_proxy), { https })
```
krunalsshah, line 44: http?

friederbluemle: Yeah, I agree it looks odd (it threw me off at first too), but if you check the code further up, `https` is actually a boolean that is passed in to this function. It later on determines which `tunnel` methods to call in `createProxyAgentFromUrl`.

PR #1660, opened 2020-06-09
Allow "cauldron repo add" to overwrite alias

friederbluemle:
This adds a new `--force`/`-f` option to the `ern cauldron repo add` command, allowing the user to overwrite an existing alias with a new value, without the need to remove it first.

### Example

```
// ('test-cauldron' alias already exists)

$ ern cauldron repo add test-cauldron /tmp/cauldron
✖ An error occurred: A Cauldron repository already exists with test-cauldron alias

$ ern cauldron repo add -f test-cauldron /tmp/cauldron
ℹ Added Cauldron repository /tmp/cauldron with alias test-cauldron
```

### DIFF
--- a/docs/cli/cauldron/repo/add.md
+++ b/docs/cli/cauldron/repo/add.md
@@ -24,10 +24,14 @@
 
 **Options**
 
-`--current true|false`
+- `--current true|false`
 
-- Set the repository as the active repository after adding it to the collection of repositories.
-- If this option is not provided, you are prompted to choose if you want to set the repository as the active repository.
+  - Set the repository as the active repository after adding it to the collection of repositories.
+  - If this option is not provided, you are prompted to choose if you want to set the repository as the active repository.
+
+- `--force/-f true|false`
+
+  Overwrite an existing alias with the same name
 
 **Example**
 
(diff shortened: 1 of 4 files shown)


docs/cli/cauldron/repo/add.md
```diff
@@ -24,10 +24,13 @@
...
 
-`--current true|false`
+- `--current true|false`
+  - Set the repository as the active repository after adding it to the collection of repositories.
+  - If this option is not provided, you are prompted to choose if you want to set the repository as the active repository.
 
-- Set the repository as the active repository after adding it to the collection of repositories.
-- If this option is not provided, you are prompted to choose if you want to set the repository as the active repository.
+- `--force/-f true|false`
+
```
krunalsshah, line 32: extra line

friederbluemle: For consistency, I added a line below `--current true|false` as well. The blank line is required for markdown to render the indented content (if not a list itself) as part of a new paragraph _below_ the list item. I agree it looks a bit odd in the non-rendered view, and I'm open to suggestions how we should potentially restructure this type of content in the future (there are many other similar cli doc markdown files).

PR #1661, opened 2020-06-09
Regenerate system test fixtures

belemaire:
Following `react-native-electrode-bridge` version update to [v1.5.25](https://github.com/electrode-io/react-native-electrode-bridge/releases/tag/v1.5.25)

### DIFF
--- a/system-tests/fixtures/ios-container/package.json
+++ b/system-tests/fixtures/ios-container/package.json
@@ -1,8 +1,7 @@
 {
   "dependencies": {
-    "@react-native-community/cli-platform-ios": "4.8.0",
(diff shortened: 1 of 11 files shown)


system-tests/fixtures/ios-container/package.json
```diff
@@ -1,8 +1,7 @@
 {
   "dependencies": {
-    "@react-native-community/cli-platform-ios": "4.8.0",
```
friederbluemle, line 3: Ah, so this is no longer required? Good, since I was getting some errors locally when they updated their package to `4.9.0`.

PR #1662, opened 2020-06-09
Bare iOS container

belemaire:
Add support for bare iOS container generation when using React Native 0.61+.

When we added initial support for React Native 0.61+ for iOS, we tightly coupled iOS container generation with MacOS, due to the need to run `pod install` during generation.

This PR removes that coupling, through configuration. It is now possible through a configuration option passed on the command line of `create-container` or through container generation in Cauldron, to have the iOS container generator, generate a more lightweight container (bare v.s full) which won't contain the node modules nor the Pods (`yarn install` and `pod install` not run). We're still keeping full container generation as the default, as being able to natively build an Android/iOS generated container post generation without having to install any extra JS tooling is one of the tenet of electrode native. That being said, this bare mode removes the tight coupling with MacOS for container generation (which we never had pre RN61), making it possible again to generate iOS container from Windows or Linux (knowing that `yarn install` / `pod install` will still need to be run prior to enable container build).

Update system tests to make use of this option. Since RN61 support, iOS containers ST were not really run on CI as we are using Linux to run system tests. This PR reenables container generation part of ST on CI.


belemaire:
Thanks for the review :+1: 

Will update the few things you noticed.

I'm surprised you haven't heard of `bare` before or maybe you just meant in ern / mobile development. Because `git` itself does use this term _(to create a `bare` repo)_ 😄 
I think that if we do explain in documentation and command line option (if making it a first class option as well as you mention) what it is then it's fine. The problem I have with `noInstall` or `skipInstall` is how to denote such a container in comparison with a "full" one.
For example in the documentation update, there is a new section :

```
Generating a bare iOS container 
```

Where we explain what a bare container is, and the config (or option) to generate one, is well ... `bare`, which makes sense ;)

Not sure how I would word that if we use `skipInstall` or `noInstall`.

```
Generating an uninstalled iOS container
``` 

That really doesn't sound right at all ... could maybe use `lightweight container` terminology instead  ?

I think you understand what I mean. I'm OK for `skipInstall`. I agree that it's very descriptive. I'm just trying to figure out how to 'name' such a container (when referring to a container generated this way, in documentation and command line option)

I'm in no rush to have this PR merged, so we can have a few more exchanges ;)

friederbluemle:
Sounds good :+1:

Yes, as you mentioned, my mind associates the term "bare" with Git (maybe too much? 😄) Maybe that was my problem.. It's not really possible to make a 1:1 comparison - while a bare repository is a clone with no checked out working directory, it's still a _complete_ clone. Any repo can be "toggled" between bare/not bare with `git config --local --bool core.bare true/false`.

For the sake of consistency with React Native (since they have the `--skip-install` flag for something _very_ similar), we should probably reuse the same terminology. Only difference would be out camel case naming of cli options `--skipInstall`). Now regarding the docs, I agree, it's easier to talk about this type of container using a short adjective. Maybe we don't need to talk about it extensively after all? ;)

Suggestion:
Use L5 header to hierarchically place the new docs below the L4 iOS header, and replace L243-247 with something like that:

```
##### Skip installing dependencies (RN >= 0.61)

When using React Native >= 0.61, you can configure the iOS container generator
to skip the automatic installation of dependencies. This can be useful to
generate an iOS container on Linux or Windows, however, it but will require
manual installation of dependencies before the container can be built.
```

Not necessarily part of this PR, but I'm thinking: Since we already know that iOS container generation will fail on Linux and Windows, should we skip installation by default on these platforms? And print an info log to console?

belemaire:
Very constructive comments thanks 👍 
Will go with `--skip-install`, I'll update documentation accordingly as well.
Also that's a very good idea to default to this mode on Linux/Windows platforms. Will do the necessary.

friederbluemle:
Awesome! Yeah, skipping it by default on Linux/Windows definitely makes sense. That's also exactly what RN CLI does (pod install only on macOS):
https://github.com/react-native-community/cli/blob/master/packages/cli/src/commands/init/init.ts#L168

system-tests/src/tests/create-container-ios-fixture.js
```diff
@@ -21,7 +21,7 @@ const excludeFilter = [
 
 const command = `create-container --miniapps ${miniapps.join(
   ' '
-)} -p ios --out ${process.cwd()}`
+)} -p ios --extra --skipInstall --out ${process.cwd()}`
```
friederbluemle, line 24: Leftover `--extra` here?

PR #1669, opened 2020-06-19
Fix metro config blacklistRe generation on Windows

belemaire:
Kinda follow up to https://github.com/electrode-io/electrode-native/pull/1666

`blacklistRe` paths in `metro.config.js` are not generated properly on Windows.

For example, this is how a path will currently look like when generated `/C:\foo\node_modules\react-native\/.*/`, while valid path should instead be `/C:\\foo\\node_modules\\react-native\\.*/`

To be noted that there is some duplication here (in `start` and `generateComposite`), not limited to this line. Some refactoring should be done. Not part of this PR only addressing this issue with minimum changes, for patch release.

### DIFF
--- a/ern-composite-gen/src/generateComposite.ts
+++ b/ern-composite-gen/src/generateComposite.ts
@@ -282,7 +282,14 @@ You should resolve the following version mismatches prior to retrying.${os.EOL}`
         .concat(
           ...localMiniAppsPaths.map(p => path.join(p, 'node_modules/react'))
         )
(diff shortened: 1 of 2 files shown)


friederbluemle:
I actually did see this when I opened #1666, but wrongly assumed it was working as intended, since the single backslashes were not part of a normal string but a regular expression. My bad sorry about that.

belemaire:
No worry 👍  I assumed the same when looking at the generated file, took me a little while to figure out that this was an issue.

PR #1679, opened 2020-06-25
Remove jitpack from default repos

friederbluemle:


belemaire:
@friederbluemle, I am thinking about closing this PR.
Just happen to notice that [react native android app template itself is having jitpack declared in the list of default repositories](https://github.com/facebook/react-native/blob/e629e94b466ebbd5924b1d4493c026004dad707d/template/android/build.gradle#L35).
So instead of introducing a breaking change here and having to update existing plugin configurations in manifest that are pulling dependencies from jitpack, we should probably just keep it, we would also remain more closely aligned with react native itself.
I remember that we discussed getting rid of jitpack after noticing some internal issues with it from build machines, but I think it was only a temporary failure, not a real blocker. And even so, getting rid of it to work around the problem rather than to address the real problem, would not really have been the best solution.
What do you think ?

friederbluemle:
@belemaire - Sure, makes sense :+1:

PR #1682, opened 2020-06-28
Revert "Set Android container Gradle plugin to 3.5.2"

friederbluemle:
The update of the Android Gradle plugin in the container from `3.4.0` to `3.5.2` is causing some unexpected regressions. In certain project setups, it might result in the following container build error:

```
More than one file was found with OS independent path 'classes.jar'
```

Other people have faced similar issues, and there is no simple fix available.

https://stackoverflow.com/questions/57725323/more-than-one-file-was-found-with-os-independent-path-classes-jar-com-andro
https://github.com/gradle/gradle/issues/10882

It will most likely require some extended investigation to figure out exactly what's going wrong and what the proper fix will look like.

This reverts the update, and instead changes the default version from `3.4.0` to `3.4.2`, keeping the container plugin at the `3.4.x` series. `3.4.2` is confirmed to work as expected.

### DIFF
--- a/ern-container-gen-android/src/hull/build.gradle
+++ b/ern-container-gen-android/src/hull/build.gradle
@@ -4,7 +4,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.5.2'
+        classpath 'com.android.tools.build:gradle:{{{androidGradlePlugin}}}'
     }
 }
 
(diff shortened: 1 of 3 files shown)


samuelneff:
@friederbluemle Thank you for posting this. I ran into the same problem with a different component and downgrading Gradle resolved it. I didn't see this solution anywhere else online.

friederbluemle:
Thanks for the comment @samuelneff - Yes, it's quite surprising what can break with a minor (or sometimes even patch) update of the Android Gradle plugin.

PR #1687, opened 2020-07-07
Guarantee miniapps import order in composite

belemaire:
The order of miniapps imports in composite `index.js` was not deterministic, which could lead to some issues in the case of miniapps having to be imported before other ones (should never be the case, still).

This PR achieves the necessary to guarantee deterministic miniapps imports order.

To avoid introducing any additional configuration complexity, the order of imports will exactly match the order of the miniapps as defined in cauldron (when generating a composite from cauldron), or the order of the miniapps provided on command line (when generating a composite with a command taking arbitrary miniapps paths).

### DIFF
--- a/ern-composite-gen/src/createIndexJs.ts
+++ b/ern-composite-gen/src/createIndexJs.ts
@@ -1,26 +1,33 @@
 import path from 'path';
 import fs from 'fs-extra';
-import { readPackageJson } from 'ern-core';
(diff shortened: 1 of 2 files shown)


krunalsshah:
While doing a code-push, the mini-app with change would appear at the top of `"miniapps"`
While we create the bundle, do we get to any issues in the case of mini-apps having to be imported before other ones?

PR #1690, opened 2020-07-09
Fix hermes default version for RN63+

belemaire:
Fix very silly bug introduced in https://github.com/electrode-io/electrode-native/pull/1688 that could have been caught easily by unit tests. But there was no unit tests for this function, so adding some as well so that this does not happen again next time ;)


friederbluemle:
Haha I think we've had a similar one before? Good that there are some tests now 👍

PR #1693, opened 2020-07-16
Fix RN61+ iOS container issue with scoped packages

belemaire:
Scoped native modules packages are not copied to the proper destination in iOS container `node_modules` directory. 

For example `node_modules/@foo/bar` is currently copied into `node_modules/bar`

This issue doesn't cause container generation to fail _(which explains in part why it was only noticed now),_ but the CocoaPods of all such scoped modules are just not injected in the generated container.

This PR fixes this problem.


friederbluemle:
Nice catch!

PR #1696, opened 2020-07-20
Add overlay option to miniapp view init method

jw2175:
Add new init method to accept overlay option. Also some method name inconsistencies addressed. Helped method added for rootview color

### DIFF
--- a/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
+++ b/ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
@@ -139,6 +139,18 @@ NS_ASSUME_NONNULL_BEGIN
                      properties:(NSDictionary *_Nullable)properties;
 
 /**
+Returns a react native miniapp (from a JSBundle).
+
+@param name The name of the mini app, preferably the same name as the jsbundle
+without the extension.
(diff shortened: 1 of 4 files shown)


ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
```diff
@@ -138,6 +138,18 @@ NS_ASSUME_NONNULL_BEGIN
...
+
+@param name The name of the mini app, preferably the same name as the jsbundle
+without the extension.
+@param overlay Boolean determines if view should be rendered as an overlay.
+@param properties Any configuration to set up the mini app with.
+@return a UIView of the miniapp.
+*/
+- (UIView *)miniAppViewWithName:(NSString *)name
+                        overlay:(BOOL)overlay
+                     properties:(NSDictionary *_Nullable)properties
```
belemaire, line 152: Unless there is a reason to, can you move overlay as the last parameter here ?
All other overloads of `miniAppViewWithName` have properties as second parameter, so I think it would be better to keep this base ordering _(unless there is a specific reason not to)_

jw2175: moved overlay to last position

PR #1697, opened 2020-07-21
Fix typo in miniappview interface

jw2175:
Fixes missing semicolon in miniappview interface.
Also improper syntax is color method signature.


ern-container-gen-ios/src/hull/ElectrodeContainer/ElectrodeReactNative.h
```diff
@@ -149,7 +149,7 @@ without the extension.
 */
 - (UIView *)miniAppViewWithName:(NSString *)name
                      properties:(NSDictionary *_Nullable)properties
-                        overlay:(BOOL)overlay
+                        overlay:(BOOL)overlay;
 /**
```
friederbluemle, line 153: I think there should also be one blank line between declaration and the doc header of the next item?

PR #1700, opened 2020-07-22
Remove Pods directory from iOS container gitignore

belemaire:
When generating a complete (installed) container with ern (pod install being run), we don't want to exclude the Pods directory when publishing the container to a git repository.

Indeed, a complete iOS container published to git, should be 'buildable' fresh from a git clone, without having to run `yarn install` nor `pod install`.

Having this directory in the .gitignore of the iOS container, was causing such containers to fail building out of the box when cloned from the git repository.

### DIFF
--- a/ern-container-gen-ios/src/hull/.gitignore
+++ b/ern-container-gen-ios/src/hull/.gitignore
@@ -11,5 +11,4 @@ xcuserdata/
 
 # Third-party dependencies
 #
-Pods/
 Carthage/


friederbluemle:
Can we keep a todo item for a larger refactoring/restructuring at some point? We really should not be (ab)using a Git repository as a way to store third party dependencies..

belemaire:
Yeah :(
That being said we already have a 'way out' with the `--skipInstall` flag for post RN61 iOS containers that will create containers that didn't went through the `yarn install` / `pod install` steps, so bare of `node_modules` and `Pods` directories. The choice is up to the client. In our case for one of our main internal client we go with full containers as its an established (historic) convenience for all iOS developers clients of this repo, not to have to install any JS tooling on their machine (not even CocoaPods -they are on Carthage-) and be able to get a buildable container 'out of the box' using only standard iOS toolchain ... at the cost of a bigger repo size obviously, but on the 'up side', these dependencies are mostly fixed, changing very infrequently so its mostly a one time cost, and shouldn't grow the repo consequently with diffs over time.

PR #1704, opened 2020-07-23
Improve error message of extra option

belemaire:
`--extra/-e` option of some commands is highly prone to errors as its value can a JSON literal or a JSON file, that can be easily malformatted in some way. 

In that case, `ern` currently fails with somewhat obscure error `An error occurred: [parseJsonFromStringOrFile] Invalid JSON or file` which might not immediately lead the user to understand that something is wrong with the `extra` option value being provided.

This PR just improves the error message in case of an issue parsing the `extra` option value.


belemaire:
Good point. I'll go with your suggested simpler phrasing ;)

PR #1707, opened 2020-07-23
Fix iOS API generator for arrays of primitives

belemaire:
Fix a bug in iOS API generator that is producing invalid swift file when using an array of primitives as an object property type.

For example, defining an object `foo` with a `bar` property being a string array, as depicted by the following schema:

```json
{
  "foo": {
    "type": "object",
    "properties": {
       "bar": {
         "type": "array",
         "items": {
           "type": "string"
         }
       }
     }
  }
}
```

Is currently resulting in the following Swift code to be generated :

```objc
NSObject.generateObject(data: dictionary["bar"], classType: [Any].self, itemType: .self)
```

Instead of 

```objc
NSObject.generateObject(data: dictionary["bar"], classType: [Any].self, itemType: String.self)
```

The issue is not limited to string array type, but any array of primitive type is resulting in the same invalid generation.

This PR fixes that issue.

Related relevant code that is populating the mustache template can be found here in [DefaultCodegen.ts](https://github.com/electrode-io/electrode-native/blob/master/ern-api-gen/src/DefaultCodegen.ts#L1553-L1570)

### DIFF
--- a/ern-api-gen/resources/swift/libraries/ern/model.mustache
+++ b/ern-api-gen/resources/swift/libraries/ern/model.mustache
@@ -58,7 +58,7 @@
 {{/isPrimitiveType}}
 {{/isNotContainer}}
 {{^isNotContainer}}
-        if let valid{{nameInCamelCase}} = try? NSObject.generateObject(data: dictionary["{{name}}"], classType: Array<Any>.self, itemType: {{complexType}}.self),
+        if let valid{{nameInCamelCase}} = try? NSObject.generateObject(data: dictionary["{{name}}"], classType: Array<Any>.self, itemType: {{#items}}{{baseType}}{{/items}}.self),
              let {{baseName}}List = valid{{nameInCamelCase}} as? {{datatype}}  {
                  self.{{name}} = {{baseName}}List
              } else {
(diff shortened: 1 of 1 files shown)


belemaire:
I see where you're going, trying to gain some time here lol 🤣 
No chance for a test, I'm really not touching anything else in this package today, I'm not even sure how I was able to figure out a fix for this one 😉 That being said, I didn't opened the PR without going through some personal manual safety nets first. Did regenerate a bunch of internal APIs to make sure this code update doesn't cause any unwanted changes / damage. All good, so no time extension here ;)

belemaire:
Ok now Azure testing is going havoc, got 
```
electrode-io.electrode-native (UnitTests macOS) Failing after 3m 
```
and
```
electrode-io.electrode-native (UnitTests macOS) Successful in 3m 
```
Go figure.
Note to self : should remove Azure unit tests sooner than later ;)

belemaire:
@friederbluemle 
Okay, I said I wouldn't touch `ern-api-gen` anymore today, but it doesn't mean we have other ways of testing ;)
One is the E2E system tests that are generating APIs based on predefined schema fixtures, so just did that here : https://github.com/electrode-io/electrode-native/pull/1708
At least, if we introduce a regression for this array of primitive fix, we will know ;P

PR #1712, opened 2020-08-13
Add raw dependencies support to Android plugin configuration

belemaire:
Rather than introducing a new specific prefix for all type of dependencies statements as we started doing, this PR is adding a new `freeeform:` prefix for dependencies, to inject any arbitrary dependency string, allowing to cover all possible current and future needed dependency literal formats.

### DIFF
--- a/docs/platform-parts/manifest/native-modules.md
+++ b/docs/platform-parts/manifest/native-modules.md
@@ -152,26 +152,26 @@ The root directory containing the Android module. By default, the plugin configu
 
 - `dependencies`
 
-An array of one or more dependencies used to add to the container when injecting this plugin. Some plugins might have dependencies on extra libraries that need to be included in the container. The container generation adds all of these extra dependencies as `compile` statements to its `build.gradle` file.
+An array of one or more dependencies (in format "<groupId>:<artifactId>:<version>") that the native module is dependent on. The container generator will add all of these extra dependencies as `implementation` statements in the container `build.gradle` file.
 
-If the `transitive` flag is needed for a given dependency, it is possible to use the prefix `transitive:` in front of the dependencies.
+For special dependency statements needs, not using default `implementation <groupId>:<artifactId>:<version>` format, it is possible to set the full dependency statement to be injected in the `build.gradle` as illustrated in the following example:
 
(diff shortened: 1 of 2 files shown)


belemaire:
Dang 🤦 
While we're at it. 
Was just thinking about naming. Are you ok with `freeform:`. I was actually just thinking we could use something shorter such as `raw:` instead ? A preference here or something different ?

friederbluemle:
Good point! Yeah if I would need to choose between `raw:` and `freeform:` I'd probably pick `raw:`.

Another idea: What if we don't require a prefix at all and instead consider it raw/freeform _unless_ it falls into a category that is already supported? I think it can be done in a backwards compatible way. Something along the lines of:

1. Does it contain a known currently supported prefix (e.g. `transitive:`)? Process accordingly
2. Does it only contain two `:` and no spaces (for example)? Process it as a regular dependency
3. If not 1. and 2., consider it freeform/raw

Also need to update "as `compile` statements" to "as `implementation` statements` on line 155.

belemaire:
Yes that would be even better, thanks for the suggestion :+1:
Will have a take at it !

belemaire:
Ok done some refactoring to make it this way.
In addition to the `transitive:` prefix we also had support for `annotation:` and `files` prefixes.
Not sure where we were going with that, should have done "raw" support from the start not to bother adding special handling for all possible dependency statements case (there are a lot) would have made everyone lives simpler.
Anyway, deprecating all these prefixes ... I have removed mention of them from the documentation (well only `transitive:` was mentioned) and added deprecation warnings.
Will be able to remove them in a future release to clean things.
We now have only two cases now, either a string formatted as `<groupId>:<artifactId>:<version>` which will be injected as `implementation <groupId>:<artifactId>:<version>` and anything else, that will be injected raw.
Makes me wonder if we should even keep first format, but I'm OK with that so given that basic implementation statements are what is needed for 90% of dependencies, this just makes it shorter.

friederbluemle:
Awesome! :clap: Yeah, maybe good to keep for now for better backwards compatibility. We can eventually remove it during a bigger cleanup. :+1:

PR #1733, opened 2020-09-22
Update and standardize module names

friederbluemle:
A significant update that improves module naming consistency and establishes stronger default conventions more in line with standard practices (particularly for package.json based projects).

Primary components/commands affected: `create-api`, `create-api-impl`, and `create-miniapp`.

The best way to understand what's changed it probably to look at the unit tests in `ern-api-gen`/`ern-core`/`ern-local-cli`, as well as the updated system test fixtures.

`create-miniapp` will no longer ask for the npm package name if not specified - Instead, it will use the default based on the module name. It is still possible to override it with the existing `-p` flag like before.

Deprecated: `--skipNpmCheck` of `create-miniapp`. If a npm registry check is necessary, manually run `npm info`.


belemaire:
Big one !
Thanks for the clear description, and clean commits, LGTM :+1: 🚀

belemaire (approved):

PR #1736, opened 2020-10-05
Improve an error message in iOS container generator

belemaire:
Some users have reported the following obscure error when generating an iOS container 

`✖ An error occurred: TypeError: Cannot read property 'productReference' of null`

This error is coming from the library we use to parse iOS project files _(.pbxproj)_ and is usually caused by a plugin misconfiguration in manifest. Unfortunately this error is not very informative and does not clearly point to the cause of the problem.

This PR improves the error message in such a scenario, to throw this error instead _(sample)_

```
✖ An error occurred: Error when trying to inject react-native-reanimated in iOS container.
✖ This error is typically due to a plugin misconfiguration in manifest.
✖ Please make sure that the following static lib(s) reference(s) (name and target) in your configuration ...
✖ [
✖   {
✖     "name": "libRNReanimate.a",
✖     "target": "RNeanimated"
✖   }
✖ ]
✖ ... are matching the values defined in the native module pbxproj located in RNReanimated.xcodeproj/project.pbxproj.
```

### DIFF
--- a/ern-core/src/iosUtil.ts
+++ b/ern-core/src/iosUtil.ts
@@ -349,12 +349,34 @@ export async function fillProjectHull(
                 projectAbsolutePath,
                 staticLibs: project.staticLibs,
               };
-              iosProject.addProject(
-                project.path,
-                project.group,
-                target,
-                options,
-              );
+              try {
+                iosProject.addProject(
+                  project.path,
+                  project.group,
+                  target,
+                  options,
+                );
+              } catch (e) {
+                if (
+                  e.message.includes(
+                    "Cannot read property 'productReference'",
+                  ) &&
+                  options.staticLibs?.length >= 1
+                ) {
+                  throw new Error(
+                    `Error when trying to inject ${
+                      plugin.name
+                    } in iOS container.
+This error is typically due to a plugin misconfiguration in the manifest.
+Please make sure that the following static lib(s) reference(s) (name and target) in your configuration ...
+${JSON.stringify(options.staticLibs, null, 2)}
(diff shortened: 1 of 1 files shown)


friederbluemle:
Interesting.. So the error _always_ contains "`productReference`"? Also - Which library to parse pbxproj files are you referring to?

belemaire:
Yes it will always produce this error message for this specific error case
Library is [xcode-ern](https://github.com/belemaire/xcode-ern) which is a slightly modified version of [xcode](https://github.com/apache/cordova-node-xcode)

friederbluemle (approved):
I guess this is also ready to merge? Unless we can update/fix xcode-ern, and avoid the `productReference` special case in electrode-native?

belemaire:
Yes I'll rebase and merge :+1:
There is no "fix" possible here anyway as we need to get the proper native target from the native module pbxproj and we currently rely on plugin configuration in manifest to achieve that. If there is a typo in the configuration ern will not find it. The fix is to fix it in configuration.
What we can only do is improve the error logging _(as done in this pr)_ as a first step. 
We could take it further by identifying all native targets in the native module pbxproj (usually only one) and using manhantan distance to get closest matching target and indicate typo / make suggestion _(for example: Couldn't find a native target matching libRNReanimate.a. Did you mean libRNReanimated.a ?)_. 
But for now, an error message such as the one introduced in this PR is already much better than the cryptic one we had before. Given that this error will only happen in fringe cases and not frequently, there is no immediate need to spend too much time/effort on this one.

PR #1737, opened 2020-10-15
Add support for configurable iOS deployment target

friederbluemle:
Problem: iOS container generation will fail for RN 0.61+ if a dependency declares a deployment target that is higher than the React Native default in its Podfile.

Reproduce:
1. Create miniapp using latest ERN
2. Add RN dependency that contains `:ios, "11.0"` (or higher) as the platform property (OR use any dependency and manually edit its Podspec file to `11.0` to induce the error
3. `ern run-ios` will fail

Needs custom manifest from open PR: https://github.com/electrode-io/electrode-native-manifest/pull/193

Big thanks to @belemaire for adding the missing parts to this PR. We can now use the `--extra` option for `create-container` and `run-ios` (as well as the `ern` block in `package.json`) to set a custom iOS deployment target! See updated docs for more details


ern-container-gen-ios/src/IosGenerator.ts
```diff
@@ -137,6 +137,7 @@ Make sure to run these commands before building the container.`,
     };
 
     const projectSpec = {
+      iosDeploymentTarget: process.env.ERN_IOS_DEPLOYMENT_TARGET,
```
friederbluemle, line 140: This is a quick&dirty way to get the variable in here for testing. Hopefully it will never make it into a released version!

It should probably (?) be passed in via `config?.iosConfig?.iosDeploymentTarget` or similar). @belemaire any suggestions? Where would we declare such an override? Maybe via a cauldron?

belemaire: Yes, ideally user should be able to provide this one via configuration, similarly to how we allow providing via configuration the `compileSdkVersion` / `minSdkVersion` ... to be used for Android Container.
This way `iosDeploymentTarget` could either be set in cauldron configuration, and if cauldron is not used (for example direct use of create-container command, this extra configuration can still be passed using the `--extra` command line option. I'll have a quick look today as how this can be achieved, and report back here.

Aslo to be complete, in addition to the Podfile templating in manifest with `iosDeploymentTarget` should also add `{{{iosDeploymentTarget}}}` [HERE](https://github.com/electrode-io/electrode-native/blob/master/ern-container-gen-ios/src/hull/Config/ElectrodeContainer-Debug.xcconfig#L29), [HERE](https://github.com/electrode-io/electrode-native/blob/master/ern-container-gen-ios/src/hull/Config/ElectrodeContainer-Release.xcconfig#L26) and [HERE](https://github.com/electrode-io/electrode-native/blob/master/ern-container-gen-ios/src/hull/Config/ElectrodeContainer-QADeployment.xcconfig#L40) as to also align the Container deployment target on the Podfile one *(Otherwise I don't think Container can be built if it uses a deployment target of 10.0 while the Pod project uses a deployment target of 11.0)*

friederbluemle: Awesome, thanks for the suggestions!

friederbluemle: Perfect! I'll cherry-pick/squash the commit and add you as co-authored-by

belemaire (approved):
Thanks !

PR #1739, opened 2020-10-20
Fix iOS deployment target for native api impls

belemaire:
Follow up of https://github.com/electrode-io/electrode-native/pull/1737 which missed propagating the iOS deployment target for native api implementation project generation.

Moved `getDefaultIosDeploymentTarget` function to `iosUtil` part of this PR, as being a generic iOS utility function it belongs in this file, making it more accessible to other modules. Also make it return undefined instead of null, to comply with some expected types _(and in a sense it's more meaningful as the pod deployment target is not defined for versions of RN < 61 that are not using any Podfile)._


friederbluemle (approved):
Thanks! :+1:

PR #1743, opened 2020-11-13
Add custom Metro config extraNodeModules support

friederbluemle:
This adds support for custom `extraNodeModules` entries in `metro.config.json` of the Composite project.

Docs have been updated and a test case was added. The diff is pretty short and straightforward, so I won't put any additional description here.

I'm not exactly sure if something identical can be done with the existing Babel support - If that's the case then we might not need this.


belemaire (approved):
LGTM 👍 
Thanks for taking care of updating the documentation and extra tests.

PR #1747, opened 2020-12-04
Fix cauldron requirement for create-container command

belemaire:
Fix a `create-container` command bug introduced in`0.45.2` through #1745

The issue is on [this line](https://github.com/electrode-io/electrode-native/pull/1745/files#diff-21660d33eebc3be34c32e7b9851feab3c24608a6e00935391811d967c2f95782R188) where `cauldron` instance can be `undefined` _(if not connected to a cauldron, which is allowed if miniapps are explicitly provided to the command)_, leading to the command failing with `Cannot read property 'getCompositeGeneratorConfig' of undefined` in this case.

The fix here is only to replace `cauldron.getCompositeGeneratorConfig` with `cauldron?.getCompositeGeneratorConfig`.

The additional code updates in this PR is just due diligence to enhance type safety to catch such errors on cauldron instances retrieved via `getActiveCauldron` way earlier. Basically, instead of returning a `Promise<Cauldron>` the method will now return `Promise<Cauldron>` or `Promise<Cauldron | undefined>` based on the value of the `throwIfNoActiveCauldron` flag. Had this been implemented before, TypeScript would have caught this bug at build time.

### DIFF
--- a/ern-cauldron-api/src/getActiveCauldron.ts
+++ b/ern-cauldron-api/src/getActiveCauldron.ts
@@ -12,7 +12,33 @@ let currentCauldronHelperInstance: CauldronHelper;
 let currentCauldronRepoInUse: string;
 const ernPlatformUseCmdMsg = 'ern platform use <version> command';
 
-export default async function getActiveCauldron({
+export async function getActiveCauldron({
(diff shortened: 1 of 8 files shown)


friederbluemle (approved):
:+1:

PR #1752, opened 2020-12-18
Pass extraNodeModules to compositeGen in runner

friederbluemle:
With this change, `metroExtraNodeModules` are now also passed and taken into account for `ern run-android` and `ern run-ios`.

### DIFF
--- a/ern-orchestrator/src/generateContainerForRunner.ts
+++ b/ern-orchestrator/src/generateContainerForRunner.ts
@@ -48,6 +48,7 @@ export async function generateContainerForRunner(
       runLocalCompositeGen({
         baseComposite,
         jsApiImpls,
+        metroExtraNodeModules: extra?.compositeGenerator?.metroExtraNodeModules,
         miniApps,
         outDir: createTmpDir(),
       }),
--- a/ern-orchestrator/src/runMiniApp.ts
+++ b/ern-orchestrator/src/runMiniApp.ts
@@ -155,6 +155,17 @@ export async function runMiniApp(
       [];
   }
 
+  if (extra?.compositeGenerator?.metroExtraNodeModules) {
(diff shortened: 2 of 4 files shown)


ern-orchestrator/src/runMiniApp.ts
```diff
@@ -155,6 +155,15 @@ export async function runMiniApp(
       [];
   }
 
+  if (extra?.compositeGenerator?.metroExtraNodeModules) {
```
krunalsshah, line 158: we can have a test in https://github.com/electrode-io/electrode-native/blob/master/ern-orchestrator/test/runMiniApp-test.ts

ern-orchestrator/src/generateContainerForRunner.ts
```diff
@@ -48,6 +48,7 @@ export async function generateContainerForRunner(
       runLocalCompositeGen({
         baseComposite,
         jsApiImpls,
+        metroExtraNodeModules: extra?.compositeGenerator?.metroExtraNodeModules,
```
krunalsshah, line 51: https://github.com/electrode-io/electrode-native/blob/master/ern-orchestrator/test/generateContainerForRunner-test.ts
Test in

krunalsshah (approved):

ern-orchestrator/src/runMiniApp.ts
friederbluemle: Good point, I added a test :+1:

ern-orchestrator/src/generateContainerForRunner.ts
friederbluemle: OK, I added a test. At some point it would be a good idea to reevaluate how much value some of these "intermediate level" unit tests (just checking if other internal methods were called/passing through arguments). I see a lot of repetitive code that does not actually check for any correctness.

PR #1758, opened 2021-02-09
Add missing fs-extra dependency to ern-core

belemaire:
Add missing `fs-extra` dependency that is used by `ern-core` module but not declared in its package.json, causing some errors when `ern-core` is used a dependency in a project.


friederbluemle (approved):
👍

PR #1760, opened 2021-02-10
Regen system tests fixtures for RN 0.63.4

belemaire:
Regenerate system test fixtures following [dev manifest update bumping RN to 0.63.4](https://github.com/electrode-io/electrode-native-manifest/pull/199)


friederbluemle (approved):
:+1:

PR #1763, opened 2021-02-17
Add support for configurable pod binary version

belemaire:
CocoaPods support side-by-side installation of multiple versions of the `pod` executable.
If multiple versions are installed, it will by default pick the latest version (afaik).
It is possible to select a specific version of the executable for running pod commands, as indicated in [this SO post](https://stackoverflow.com/questions/20487849/how-to-downgrade-or-install-an-older-version-of-cocoapods).

When generating an iOS container, ern will run `pod install` as-is without specifying any version, so it will use the latest.

In some fringe cases, or for conducting some experimentation, it might be needed to use a specific version of the pod executable for running `pod install`.

This PR adds a new configuration property `podVersion` to ern global config, so that a specific pod executable version can be used when running `pod install` during iOS container generation, rather than using the default resolved version.

For example, if two versions of the pod executable are installed locally, `1.10.1` and `1.9.3`, `pod install` will by default use `1.10.1`. To force ern iOS container generator to use `1.9.3` instead, the following command can be run:

`ern platform config set podVersion 1.9.3`

And to reset to default version:

`ern platform config del podVersion`

### DIFF
--- a/docs/cli/platform/config/del.md
+++ b/docs/cli/platform/config/del.md
@@ -26,6 +26,7 @@
 - `manifest`
 - `max-package-cache-size`
 - `package-cache-enabled`
+- `podVersion`
 - `retain-tmp-dir`
 - `showBanner`
 - `sourceMapStoreProxy`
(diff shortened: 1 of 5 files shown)


friederbluemle (approved):
👍

PR #1765, opened 2021-02-23
Include dependency type in generated plugin config

friederbluemle:
**Update**:

This PR also increases `manifestBaseErnVersion` for `create-plugin-config` to `0.46.0`.

---

ERN started supporting "raw" dependency types in plugin configs with version 0.44 (see #1712). However, `ern create-plugin-config` did not include the dependency type in the generated file. With this small change, it will now correctly support and recognize additional types (e.g. `compileOnly`, `api`, ..) and not just add the dependency to a plain list (which would then _all_ be added using `implementation` to the container).

Example before:

```
{
  "android": {
    "root": "android",
    "dependencies": [
      "com.example:impl:1.0.0",
      "com.example:api:1.0.0",
      "com.example:annotation-processor:1.0.0",
      "com.example.compile-only:1.0.0"
    ]
  },
  "ios": {}
}
```

Now:

```
{
  "android": {
    "root": "android",
    "dependencies": [
      "implementation 'com.example:impl:1.0.0'",
      "api 'com.example:api:1.0.0'",
      "annotationProcessor 'com.example:annotation-processor:1.0.0'",
      "compileOnly 'com.example.compile-only:1.0.0'"
    ]
  },
  "ios": {}
}
```

It also includes a small fix to prevent the command from asking:
```
? Cannot resolve version for fileTree(dir: 'libs', include. Enter the version to use
```
This could happen when a `fileTree` dependency was written in different order (and result in a broken manifest config that needs to be cleaned up manually).

### DIFF
--- a/ern-core/src/AndroidPluginConfigGenerator.ts
+++ b/ern-core/src/AndroidPluginConfigGenerator.ts
@@ -32,7 +32,7 @@ export class AndroidPluginConfigGenerator {
   private readonly excludedDirectoriesRe = new RegExp(/sample|demo|example/i);
   private readonly exclusions: string[] = [
     'com.facebook.react:react-native',
-    'fileTree(include',
+    'fileTree(',
   ];
 
   private constructor(p: string) {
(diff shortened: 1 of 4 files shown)


belemaire (approved):
:+1:

friederbluemle:
Set `manifestBaseErnVersion` to `0.46.0` to make sure generated plugin config files will be put in the correct directory.

belemaire (approved):

PR #1766, opened 2021-03-02
Remove regex name match to determine module type

friederbluemle:
This _removes_ the regex name special casing for **api** and **api-impl** dependencies. With this change, very old projects may need to add a `{ ern: { moduleType: '<type>' } }` section to their `package.json`, in case they previously relied on the package _name_ to designate the module type. Type can be `ern-api`, `ern-js-api-impl`, or `ern-native-api-impl`.

Since this is a potentially breaking change, it should be stated explicitly in the next changelog (for version 0.46).

The first commit contains only test and module type import consolidation/simplification, and no functional changes.


belemaire (approved):
LGTM :+1:

PR #1771, opened 2021-03-04
Add map support to code generation

belemaire:
Follow up to https://github.com/electrode-io/electrode-native/pull/1755 just rebasing branch and adding a test use case to system test API schema. To be shipped in today's 0.46.0 release.

Closes https://github.com/electrode-io/electrode-native/pull/1755


friederbluemle (approved):
Looks good! But do you want to push that to @LcTwisk 's branch (#1755) instead, to avoid duplicate PRs and keep discussion in one place? Edits by maintainers are allowed:

image

```
git remote add lt https://github.com/LcTwisk/electrode-native.git
git push -f lt HEAD:add-map-support-to-code-generation
```

belemaire:
@friederbluemle Didn't knew about this feat. 
That's great, thanks for the tip :+1: Will do that instead indeed !

PR #1781, opened 2021-03-18
Fix generated miniapp activity

deepueg:
This PR fixes a long time pending issue where miniapp activities were using a very old implementation causing several crashes when used.  

Generated MiniApp activity will now use the latest ern-navigation base class making it usable by any project out of the box.


belemaire (approved):
:+1:

PR #1783, opened 2021-03-22
Do not mustache process Android res directory

belemaire:
Mustache processing the `res` container directory can cause mustache templating to fail, _(because for example some binary file present in res, such as an image png, contains {{ by chance, causing mustache templater to throw an error for unclosed placeholder in a file)_.

This PR addresses this issue by adding Android Container `res` directory to the list of directories excluded from mustache template processing.

### DIFF
--- a/ern-container-gen-android/src/AndroidGenerator.ts
+++ b/ern-container-gen-android/src/AndroidGenerator.ts
@@ -340,6 +340,7 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
     const pathLibSrcMainJniLibs = path.normalize('lib/src/main/jniLibs');
     const pathLibSrcMainAssets = path.normalize('lib/src/main/assets');
     const pathLibSrcMainJavaCom = path.join(pathLibSrcMain, 'java/com');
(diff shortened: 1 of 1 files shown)


friederbluemle:
Just double checking: Isn't is possible that there are placeholders in res files as well? Maybe a check for binary file would be safer (if the issue is only with binary files)? Or: Exclude files by extension, e.g. never try to mustache png, gif, svg, etc.

krunalsshah (approved):

belemaire:
Yes I kinda agree with this and had that in mind when working on this fix. 
But exclusion list would be complex, there are so many types of potential binary resource files extensions _(image / video / sound ... file types)_. Also what if an xml containing strings, contains a string that has some "{{" or "{{{", which is not intended for mustache templating _(fringe case for sure)_. We don't mustache process res files currently and it's not documented as something being supported so no official regression in this PR.
I don't really see a use case for it at this moment, so I'd rather remain on the simpler/safer side of things for now. 
If we have a real need arising at some point for mustache processing res files, we will advise accordingly.

PR #1787, opened 2021-03-26
Introduce new run-container-pipeline command

belemaire:
This PR introduces a new Electrode Native command, `run-container-pipeline`.

It allows to run a locally created container through a given Container pipeline.

A Container pipeline is a collection of Container `transformers` and/or `publishers` _(defined as `pipeline blocks`)_ that are run in sequence, one after the other, when using the `cauldron regen-container` command.

This is a convenient structure as it allows to define a chain of transformer and/or publishers to run, along with their configuration, in the form of JSON configuration files stored locally, or in a Cauldron. It ultimately simplifies the sequencing of transformers/publishers without having to chain multiple `transform-container`/`publish-container` commands manually, each having to be fed their own configuration.

One limitation, thus far, of such a pipeline configuration, is that it is only consumed when using the `cauldron regen-container` command. This command, first of all strongly requires a Cauldron, but also encompass many other tasks _(generating the container, updating the cauldron descriptor, yarn lock ...)_ which might not be needed or desired in case one just wants to run a sequence of transformer/publishers conveniently for a pre-generated container, and nothing more.

Up until this PR, the only was to run a container through a sequence of transformers/publishers, was to manually sequence `publish-container`/`transform-container` commands. This PR tackles this inconvenience, with the introduction of the `run-container-pipeline` command, that allows to run a container through a pipeline, configured in a JSON file _(either stored locally or in a remote cauldron)_, simplifying some use cases.

### DIFF
--- a/docs/SUMMARY.md
+++ b/docs/SUMMARY.md
@@ -75,6 +75,7 @@
 - [create-plugin-config](cli/create-plugin-config.md)
 - [publish-container](cli/publish-container.md)
 - [transform-container](cli/transform-container.md)
+- [run-container-pipeline](cli/run-container-pipeline.md)
 - [create-miniapp](cli/create-miniapp.md)
 - [link](cli/link.md)
 - [regen-api](cli/regen-api.md)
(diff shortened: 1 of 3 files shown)


friederbluemle (approved):
Nice 👍

PR #1790, opened 2021-03-31
Generate miniapp navigation controllers for iOS containers

jw2175:
@belemaire goal is to generate subclasses of ENBaseNavigationController inside the container. I'm able to generate the files, following the example from https://github.com/electrode-io/electrode-native/blob/master/ern-container-gen-android/src/AndroidGenerator.ts

However, not sure how to include the references to the new files inside the pbxproj file of the container so that it will show up in the project navigator when using Xcode.

### DIFF
--- a/ern-container-gen-ios/src/IosGenerator.ts
+++ b/ern-container-gen-ios/src/IosGenerator.ts
@@ -6,6 +6,7 @@ import {
   kax,
   log,
   manifest,
+  mustacheUtils,
   NativePlatform,
   PackagePath,
   PluginConfig,
@@ -34,6 +35,7 @@ import { Composite } from 'ern-composite-gen';
 import semver from 'semver';
 
 const ROOT_DIR = process.cwd();
+const PATH_TO_TEMPLATES_DIR = path.join(__dirname, 'templates');
 const PATH_TO_HULL_DIR = path.join(__dirname, 'hull');
 
 export default class IosGenerator implements ContainerGenerator {
@@ -266,6 +268,12 @@ Make sure to run these commands before building the container.`,
       mustacheView,
       config.composite,
     );
+
+    await kax
+      .task('Adding MiniAppNavigationController Classes')
+      .run(
+        this.addMiniAppNavigationControllerClasses(config, iosProject),
+      );
 
     await kax
       .task('Adding Native Dependencies Hooks')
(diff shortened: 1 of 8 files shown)


belemaire:
@jw2175 
Here is what I'd suggest:

- **Create a new group in the container to contain these generated source files _(not strictly needed, but cleaner than putting them directly in root ElectrodeContainer group)_**

This can be easily done directly editing the container pbxproj, no need to go through XCode. 
Just generate a 24 character random hex string _(or just pick an existing one from the pbxproj and just change a character)_ and add an entry for the new group reference after [THIS LINE](https://github.com/electrode-io/electrode-native/blob/be7e416fa0a1de4408d184d9773e47c22dd0fa94/ern-container-gen-ios/src/hull/ElectrodeContainer.xcodeproj/project.pbxproj#L183). You can name the group as you wish _(my suggestions would be something like `NavigationControllers` or `NavControllers`)_. You'll also need to add a group declaration in the pbxproj, for this new group. You can just copy this one _(APIs group)_ and paster new one just below it _(just reuse the same hex string id and group name as used before)_. That should be it to add this new group in container.

- **Add references to files in pbxproj**

This is your initial question. 
To achieve this, you should first relocate your new block of code in between [THIS LINE](https://github.com/electrode-io/electrode-native/blob/v0.46/ern-container-gen-ios/src/IosGenerator.ts#L268) and [THIS ONE](https://github.com/electrode-io/electrode-native/blob/v0.46/ern-container-gen-ios/src/IosGenerator.ts#L275). _(Ideally, if you can, please create a new method in IosGenerator.ts to contain your code rather than putting it directly inside `fillContainerHull`. This method is already way too big and needs a refactoring, let's try not to put more burden to it)_.
Now for the main question. Note that [this call to fillProjectHull](https://github.com/electrode-io/electrode-native/blob/v0.46/ern-container-gen-ios/src/IosGenerator.ts#L261) returns an `iosProject` object. This is the object that allows to parse & manipulate the pbxproj. 
You can use it to add files to the pbxproj. [This is how the `addFile` directive of manifest plugin configuration does it](https://github.com/electrode-io/electrode-native/blob/v0.46/ern-core/src/iosUtil.ts#L323-L334). So all you should need is a call to `iosProject.addFile` with path to file _(should just be filename itself I think)_ and group name _(the name of the group you created in previous step)_, you can do that for each file. I'm not sure why we have is target dependency step, I don't think you need it, just use addFile and see.
If you need to "export" the source, you may also want to call `addSourceFile` for each file [see here for invocation](https://github.com/electrode-io/electrode-native/blob/v0.46/ern-core/src/iosUtil.ts#L272-L276).

belemaire (approved):
:+1: Nice

PR #1795, opened 2021-04-21
Update iOS runner to use Swift 5

jw2175:
- Removed objc code
- Now 100% swift, (swift 5 compiler)
- Extends base navigation class from ern navigation api impl
- Also contains a fix for ios runner gen fixtures regeneration script


jw2175:
@belemaire Please provide some guidance on how to regen test fixtures for the runner.

belemaire:
@jw2175 
There is a `regen-fixtures` npm script in `ern-runner-gen-ios` directory.
Unfortunately I gave it a shot and it looks like it's missing something that cause it not to work as expected in certain circumstances. 
To fix the script, just edit [this line](https://github.com/electrode-io/electrode-native/blob/master/ern-runner-gen-ios/package.json#L11)
The command should be `"yarn build && ../node_modules/.bin/ts-node test/regen-fixtures"` _(just add missing `yarn build`)_.
Please commit back this fix part of this PR it's ok.
Then just run the following to regen the fixtures :
```
cd ern-runner-gen-ios
yarn regen-fixtures
```
You may also want to run afterwards :
```
yarn test
```
From the `ern-runner-gen-ios` directory _(so that it only runs ios runner tests)_ to validate that tests are ✅

belemaire (approved):

deepueg:
Looks good. Do we need to consider anything for backward compatibility here ? what happens for someone with an existing runner(old) ? Do we need to document this as a breaking change for them  ?

jw2175:
@deepueg  It's backwards compatible in the sense that if they already have an iOS folder and run ern run-ios, that will still work and run the existing objc code. otherwise, ios folder will be generated with swift code.

PR #1796, opened 2021-04-21
Add excludedNativeModules to ios container config

belemaire:
Add new `excludedNativeModules` iOS container configuration property.

It can be used to selectively exclude one or more native modules pods so that they don't get injected inside the container.
It has the same effect as using `ignorePodSpec` directive in a manifest plugin configuration.
While the native module pod will be ignored, causing the native module native code to be completly ignored and not injected in the container, it won't impact the JavaScript side of the native module which will still be bundled in case the JS application code is importing and consuming the native module JS surface.

As any iOS container generator configuration, this property can either be set in cauldron configuration via `containerGenerator.iosConfig` or directly supplied to `create-container` or `run-ios` commands, using the `--extra` option.
i.e `create-container --extra='{"iosConfig": {"excludedNativeModules": ["react-native-module-name"]}'`

This property can be used in case one container flavor should contain a certain native module native code _(in case the native application depending on the container doesn't have this native code)_, while another shouldn't _(in case the native application depending on the container already contains this native code)_.. This is obviously for fringe uses cases as the native module native code shouldn't be part of the client application _(unless this is a react native application, which Electrode Native doesn't officially support -yet-)_

Due to the very fringe nature of this feature, needed only for a single internal use case, it won't be documented at this point.

### DIFF
--- a/ern-container-gen-ios/src/IosGenerator.ts
+++ b/ern-container-gen-ios/src/IosGenerator.ts
@@ -267,13 +267,12 @@ Make sure to run these commands before building the container.`,
       config.plugins,
       mustacheView,
       config.composite,
+      config.iosConfig,
     );
 
     await kax
       .task('Adding MiniAppNavigationController Classes')
-      .run(
-        this.addMiniAppNavigationControllerClasses(config, iosProject),
-      );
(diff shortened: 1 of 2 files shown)


jw2175 (approved):
nice, thanks

PR #1798, opened 2021-04-29
Bump iOS runner deployment target to 11.0 for RN >=0.64.0

jw2175:
RN 0.64 bumped default deployment target to 11.0. Runner currently hardcoded to 10.3, and will not compile.


belemaire:
@jw2175 Nice catch :+1:

Let's set the deployment target based on RN version to avoid having to impose 11.0 for older versions of RN if it's unneeded. Moreover it's relatively straightforward to implement so no reason not to ;)
Would basically just need to update the mustache view object created [HERE](https://github.com/electrode-io/electrode-native/blob/master/ern-runner-gen-ios/src/IosRunnerGenerator.ts#L85), adding a `deploymentTarget` property to it, being set to `11.0` if React Native version is greater or equal than 0.64.0, and `10.3` otherwise. 
The mustache view [is already applied to the runner pbxproj when generating the runner](https://github.com/electrode-io/electrode-native/blob/master/ern-runner-gen-ios/src/IosRunnerGenerator.ts#L28) so you'll just need to add `{{{deploymentTarget}}}` new placeholder in runner pbxproj at adequate location(s).

Ping me back if you have questions or once done, so that I can final review.

belemaire (changes requested):
Thanks LGTM.

Could you please just add `semver` dependency (`"semver": "^5.5.0"`) to the `package.json` of `ern-runner-gen-ios`. Not having it there will work because `ern-runner-gen-ios` package is usually installed with all other packages from ERN, so the semver dependency will be installed through another package. However in case `ern-runner-gen-ios` is used as a standalone package for some reason, the import will fail. Even if that's never done in practice, it's still good to know which packages `ern-runner-gen-ios` is dependent on by looking at its `package.json`.

jw2175:
> Thanks LGTM.
> 
> Could you please just add `semver` dependency (`"semver": "^5.5.0"`) to the `package.json` of `ern-runner-gen-ios`. Not having it there will work because `ern-runner-gen-ios` package is usually installed with all other packages from ERN, so the semver dependency will be installed through another package. However in case `ern-runner-gen-ios` is used as a standalone package for some reason, the import will fail. Even if that's never done in practice, it's still good to know which packages `ern-runner-gen-ios` is dependent on by looking at its `package.json`.

@belemaire sounds good, done!

belemaire (approved):
Nice ! Thanks :+1:

belemaire (changes requested):
Oh wait, I approved a bit too fast.
There is still a linting issue :( 
```
ERROR: ern-runner-gen-ios/src/IosRunnerGenerator.ts:96:7 - The key 'deploymentTarget' is not sorted alphabetically
```
just move the `deploymentTarget` prop, at the top of `mustacheView` object, should go awayl

belemaire (approved):
Ok, should be all good now :) 👍

friederbluemle (approved):
:+1:

PR #1805, opened 2021-05-06
Bump dev dependencies

belemaire:
- Bump all dev dependencies to their latest
- Update husky configuration due to breaking changes in major version bump

Because of some dev dependencies involved in system tests, ran them all to ✅ completion to ensure nothing got broken there.


krunalsshah (approved):

package.json
```diff
@@ -18,6 +18,7 @@
     "coverage:upload": "nyc report --reporter=text-lcov | coveralls",
     "ern-debug": "node --nolazy --inspect-brk=5858 ern-local-cli/src/index.dev.js",
     "instrument": "lerna run instrument",
+    "prepare": "husky install",
```
friederbluemle, line 21: Can we leave this script out of package.json, and instead add instructions to README.md to run `yarn husky install` _for people who want husky set up locally_? I always prefer to run exactly what I want individually (which is why I've set an environment variable `HUSKY_SKIP_INSTALL` locally, but not sure if it is still recognized by updated husky).

friederbluemle (approved):
Nice. Long overdue. I've also attempted to update husky a few months ago, but wasn't sure how to separate it from the code base sufficiently so that people working with the repo still have a choice. I guess for now it's OK to have the `.husky` directory, another possible option in the future is to provide a way to explicitly modify the local scripts in `.git/hooks` directly (i.e. remove dependency on husky altogether).

package.json
friederbluemle: More context: https://github.com/typicode/husky/issues/468

Instead of using the `prepare` hook, maybe something like this would work too?

```
  "scripts": {
    "install-git-hooks": "npm install husky --no-save"
  },
```

friederbluemle: Yes, I agree that most contributors will be fine with the hooks being enabled by default (not many have complained so far ;)). That could be either because they find it useful and actually _want_ these hooks to be set up, or because they are _unaware_ of the scripts and don't mind the extra time it takes/consider it normal for this repo.

The point I'm trying to make:

1. Git is a purely **local** system (until commits are pushed to remote)
2. Verification of contributions and adherence to coding/formatting standards always needs to be done **remotely** anyway (or at least they should)

Exactly as you mentioned, local hooks can easily by skipped (`--no-verify`), while remote checks _cannot_ be skipped. Mixing the two _by default_ is not a good idea in my opinion (without an easy way to opt out).

Ultimately it's your decision, but hopefully you can give this another thought. In general it would be nice to _decrease_ the dependency on external tools and libraries over time.

A script/prompt that asks the first time it runs sounds like a good idea to me too..

friederbluemle:
After reading a bit more and going through the source code of the new Husky version, I have to say that I'm really not a big fan of the way it works now. That's unrelated to the fact that I prefer individual commands to do one thing ;) The low level way it integrates with existing systems and the files/folders involved:

Basically, it now both hijacks the npm `prepare` hook (which was never really meant for Husky), _and_ the native Git config system.
Previously, it was possible to avoid installation altogether by setting `HUSKY_SKIP_INSTALL`. Now, if I'm not mistaken, there is `HUSKY` which at first glance appears to provide a similar way to disable it. However, even with `export HUSKY=0` it _still_ adds a line to `.git/config`, and the generated/ignored script in `.husky/_`.

_Every_ time (even if already installed), a simple `yarn` to install/update deps, as well as other Yarn commands, Husky additionally prints to console:

```
$ husky install
husky - Git hooks installed
```

For such a simple task (basically copying Git hook scripts) the entire system just seems very intrusive and over-engineered.

friederbluemle:
Alternate suggestion to manually copy the scripts with a single command in #1806 - Please have a look and give it some thoughts @belemaire 🙏

belemaire:
@friederbluemle 
Thanks much for this extra work.
That being said, I'm merging this PR because it does what it says, bumping all dev dependencies and doing necessary changes to accommodate for breaking changes -nothing added / nothing removed-. 
Will follow up on #1806

PR #1806, opened 2021-05-07
Move husky hooks to individual script files

friederbluemle:
An alternate suggestion to the husky update in #1805.

This PR instead removes the `husky` dependency completely, and (like #1805), moves the commands from `.huskyrc` to separate script files.

Activating the hooks is done either via `setup-dev.js`, or a one-time command after cloning the repo:

```
cp -a .githooks/* .git/hooks/
```

Also added a few lines of documentation in CONTRIBUTING.md explaining what each script does.


friederbluemle:
I've rebase the PR and resolved the conflicts. As suggested, the scripts will now be installed automatically when running `setup-dev.js` as outlined in the [development setup instructions](https://github.com/electrode-io/electrode-native/blob/master/CONTRIBUTING.md#electrode-native-development-setup). At the same time we can get rid of the `husky` dependency and the `prepare` override.

PR #1807, opened 2021-05-11
Add support for Kotlin native modules

belemaire:
This PR adds support for Android native modules written in Kotlin.

It will add the necessary dependencies to the container **if and only if** at least one Kotlin native module has to be injected in the container. Indeed, if no Kotlin based native module has to be injected in a container, there is no point in adding any Kotlin dependencies overhead to the container.

The first native module that have been tested and validated using Kotlin support added in this PR is [react-native-pager-view](https://github.com/callstack/react-native-pager-view). It is a plugin that has been internally requested by one of our client teams. Plugin configuration will be added for this plugin, once ERN 0.48.0 is released with this PR. 

It is up to MiniApp development teams to decide whether or not to consume Kotlin native modules, knowing the potential overhead _(voided if the consuming application is partially or fully written in Kotlin)_. From Electrode Native perspective we are only offering support of such native modules starting in 0.48.0, but the container will remain empty of any Kotlin dependencies if no such native modules are used.

This PR is using the latest version of Kotlin of 1.4.x line (1.4.32) as the default Kotlin version instead of the latest (1.5.0), because this latest version requires Gradle >= 6.1, which would have caused other version changes in this PR and potential side effects. Another PR should bump to Gradle 6.1 for ERN if deemed necessary. To be noted that as for other Android tools versions, the default version hardcoded in ERN can be ovverriden via configuration.

Also to be noted that starting from React Native 0.65, Kotlin standard library will be transitively pulled in any case due to the recent bump to OkHttp v4 via https://github.com/facebook/react-native/pull/31084.  Starting from this version, no matter if the container is containing a Kotlin native module or not, Kotlin stdlib will always be pulled.


ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -314,6 +339,11 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
     androidDependencies.regular.push(
       `com.android.support:appcompat-v7:${versions.supportLibraryVersion}`,
     );
+    if (isKotlinEnabled) {
+      androidDependencies.regular.push(
+        `org.jetbrains.kotlin:kotlin-stdlib:${versions.kotlinVersion}`,
```
krunalsshah, line 344: It's likely the plugins may end up using Kotlin extensions that are included with Android Jetpack
We can consider including `androidx.core:core-ktx`

belemaire: Thanks @krunalsshah for the suggestion
I took a look at some random popular native modules written in Kotlin, and I don't see any of them using this dependency :
- [react-native-webview build.gradle](https://github.com/react-native-webview/react-native-webview/blob/89fd2db39b6b86e83b0c3c0ff5a31b83b9e564f3/android/build.gradle#L138)
- [react-native-pager-view build.gradle](https://github.com/callstack/react-native-pager-view/blob/fa14630fda5e202df559fd10441a6b0deca59fd1/android/build.gradle#L129)
- [react-native-tor build.gradle](https://github.com/Sifir-io/react-native-tor/blob/9f9f6a23d2a6150e5359656b737629aa1661fa24/android/build.gradle#L129)
- [react-native-file-access](https://github.com/alpha0010/react-native-file-access/blob/784645e5cf9cfb6e0ba3dfe24b7ee45912b4c2fe/android/build.gradle#L128)

While all of them are using stdlib _(duh ... no choice ;P)_
So it doesn't look that common for native modules to use these extensions _(exclusion rather than the norm)_
Given this, it's better not to add this dependency to the container to avoid extra unnecessary overhead. 
If a native module is using these extensions, the dependency on `android.core:core-ktx` should be added to the `dependencies` object of the corresponding plugin configuration in manifest, as any other additional dependencies that the plugin requires.

krunalsshah:
@belemaire thanks for the PR, we can update the documentation where we can pass this as an extra to override the version

krunalsshah (approved):

ern-container-gen-android/src/hull/build.gradle
```diff
@@ -5,6 +5,9 @@ buildscript {
     }
     dependencies {
         classpath 'com.android.tools.build:gradle:{{{androidGradlePlugin}}}'
+        {{#isKotlinEnabled}}
```
friederbluemle, line 8: Minor: Move template conditionals (line 8 and 10) to column 1, as indentation is not part of the line-by-line evaluation. This is to avoid potential trailing spaces in the final file. (see line 16/19 etc).

friederbluemle (approved):
Nice 👍 Definitely good to only conditionally include it. I figured you would need this to add support for `react-native-tab-view` (which depends on kotlin-containing `react-native-pager-view`). ;)

PR #1808, opened 2021-05-13
Bump all dependencies versions

belemaire:
Bump all dependencies versions to their latest and perform necessary code updates to accommodate for breaking changes.
Because quite some dependencies are now having a requirement on Node 12, and given that Node 10 has reached end of life, this PR also updates the minimum node version for ERN to 12 _(from 10)_.
It updates CI scripts accordingly to make Node 14 the base _(from Node 12)_, dropping Node 10 and adding Node 16 _(so that 12,14,16 instead of 10,12,14 are now used)_.

**For release notes: This is a breaking change**


krunalsshah (approved):

ern-cauldron-api/src/util.ts
```diff
@@ -1,5 +1,5 @@
 import _ from 'lodash';
-import Joi from '@hapi/joi';
+import Joi, { ValidationResult } from '@hapi/joi';
```
krunalsshah, line 2: `ValidationResult` is unused

PR #1809, opened 2021-05-20
Dedupe yarn.log and apply formatter

friederbluemle:
Two quick commits here with no functional changes:

- Deduplicate yarn.lock (removing > 100 lines of duplicate declarations)
- Format code base (there were a bunch of pending changes, likely a result of the `prettier` update in #1808)


krunalsshah (approved):
👍

PR #1812, opened 2021-05-24
Fix bug in code-push release

belemaire:
Fix a regression of `code-push release` command, that got introduced in [v0.44.1](https://github.com/electrode-io/electrode-native/releases/tag/v0.44.1) release through https://github.com/electrode-io/electrode-native/pull/1720

The bug comes from the fact that `unionBy` keeps entries from the first array, causing the current MiniApps SHA _(or JS API impls)_, as set in Cauldron, to be picked, instead of the ones supplied to the command, leading to the CodePush release being a noop.

Instead of preserving the ordering by switching `unionBy` arrays _(causing this buggy side effect)_, revert the order of the arguments to the right one, and reorder MiniApps/JS API implementations via a post processing function.

### DIFF
--- a/ern-orchestrator/src/codepush.ts
+++ b/ern-orchestrator/src/codepush.ts
@@ -371,16 +371,28 @@ export async function performCodePushOtaUpdate(
       );
     }
 
-    const miniAppsToBeCodePushed = _.unionBy(
-      referenceMiniAppsToCodePush,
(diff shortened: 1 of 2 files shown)


friederbluemle (approved):
👍 Sounds complicated, any chance of a few tests to make it easier to understand? ;) If urgent and tested manually, merge right away 🍻

PR #1813, opened 2021-05-24
Add support for Android dynamic feature modules

belemaire:
See documentation update part of this PR for additional details.

Please note that this solution for dynamic feature module support has been tested is already working in production for one of our internal client teams. Custom AARs `0.63.100` and `0.64.100` are already available.

Even though we really didn't want to have to fork react native, there was unfortunately no other way to fix this issue.
That being said, the fork for now is only limited to this change for dynamic feature modules support, and only for android react native library. We are not planning any other custom modifications.
Hopefully if we get to have these changes pushed to react native _(will require some refactoring to make it cleaner)_, we may get rid of the fork and this mechanism _(still this is debatable, in case we need some urgent fix that may take a while to get through official RN release, this might still come in handy)_.


friederbluemle (approved):
Wow, pretty advanced feature/use case, but good to have :+1:
As you mentioned, it would be nice to have this minor addition somehow incorporated into upstream `react-native`, but totally understand it could take quite a long time.
And something else to think about (you might have already considered this case): Using `100` as patch version will probably work in most cases, but could get ambiguous/irregular quickly. It is not clear whether `0.63.100` is based on `0.63.0`, `0.63.1`, `0.63.2`, etc. For example, if there is an important patch in react-native, would it _always_ mean a new patch "100+x", etc? What if other changes from ern need to be released on top of a "custom" release (yes, unlikely, but not impossible). Not sure what the best solution is, potential options:
`0.63.0` -> `0.63.100`
`0.63.1` -> `0.63.101`
etc.
-or (ab)using a prerelease tag, which would allow additional releases from the ern side-
`0.63.0` -> `0.63.0-ern.0`, `0.63.0-ern.1`, etc.
`0.63.1` -> `0.63.1-ern.0`
`0.63.2` -> `0.63.2-ern.0`

belemaire:
@friederbluemle Yes there is no ideal solution here. I chose 100 as custom patch version as I'm pretty confident no patch releases of React Native will reach this point, given their release history.
Idea I had in mind is as you mention, releasing new versions from 100 and onward for any new official patch release of the version line or custom additional ern changes. We should document these custom releases on our GitHub fork _(via standard GH release tag description)_ to indicate of which version _(patch)_ they are cut off. 
On client side, it might be better _(or we could suggest it in documentation)_ to use the `+` placeholder _(even if not super recommended, in this context should be fine)_ not to have to account for every new 100+ releases. _(i.e if your application is using 0.64 RN, just use version 0.64.+, which will get the latest of 100 line including any propagated official patch releases of this version line -as I don't see any reason not to always be on latest patch update for RN-)_

PR #1815, opened 2021-05-28
Update ElectrodeReactContainer.java and regen fixtures

friederbluemle:
A very minor update to `ElectrodeReactContainer.java` (to avoid an unnecessary change to the fixtures, as well as an unused import). Also the fixtures needed to be regenerated due to some recent changes in the code base (tests were failing).


belemaire (approved):
:+1:

krunalsshah (approved):

PR #1816, opened 2021-06-02
Make electrode plugin delegates nullable

jw2175:
This allows calling of start method from swift to be able to pass nil to electrode plugin config. Found after runner was moved to Swift.


belemaire (approved):
Thanks 👍

PR #1818, opened 2021-06-16
Pass resolutions to runLocalCompositeGen

belemaire:
`runLocalCompositeGen`, the function used to generate a composite without retrieving composite generator configuration from cauldron _(used indirectly by commands such as `run-android`/`run-ios`/`create-container` ..)_ is never invoked with the `resolutions` argument, meaning that it is not possible to leverage Yarn selective dependency resolutions support when using these commands.

This PR addresses this problem by making it now possible to supply `resolutions` to these commands via the `--extra` configuration object option _(in the `compositeGenerator` object)_

### DIFF
--- a/docs/platform-parts/composite/index.md
+++ b/docs/platform-parts/composite/index.md
@@ -10,8 +10,8 @@ You can also manually generate a Composite project, with the [ern create-composi
 
 ## Babel support
 
-Electrode Native has support for using Babel plugins in MiniApps.  
-Babel support is still in an early phase and will be improved and simplified over time.  
+Electrode Native has support for using Babel plugins in MiniApps.
+Babel support is still in an early phase and will be improved and simplified over time.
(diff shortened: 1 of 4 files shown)


friederbluemle (approved):
:+1:

belemaire:
@friederbluemle thanks for the review :+1:
Will merge despite the failing tests on Node 16. 
Initially thought it was a glitch as it sometimes happen on GitHub actions nodes, but after a few re-runs always fails the same way. Tried it locally with Node 16 and was able to repro the error.
The problem has not been introduced in this PR, it's failing the same way on master, so will merge this one and tackle issue with Node 16 as a separate PR.

friederbluemle:
Sure sounds good! I also noticed a couple of weird errors with Node 16 in different repos locally. We can fix it separately :+1:

PR #1823, opened 2021-07-15
Add MiniAppsConfig file to iOS container gen

jw2175:
This PR allows access to miniapp names for consuming native app.


belemaire (approved):
👍

PR #1824, opened 2021-07-21
Update plugin config gen with kotlin support

### LINKED ISSUE #1757: Cannot create plugin for modules written in Kotlin
Whenever I attempt to create a plugin for a third-party package that contains Kotlin (e.g., https://github.com/stashenergy/react-native-msal) it always fails in https://github.com/electrode-io/electrode-native/blob/3bc17330d3d0d9ad94dde523baf096cada954a1c/ern-core/src/AndroidPluginConfigGenerator.ts#L154 . 

Looks like it needs some tweaks to look for `*Package.kt` files as well, plus some additional parsing of the Kotlin files as well.

friederbluemle:
Several updates and improvements to `AndroidPluginConfigGenerator.ts`. Biggest user visible change is that it now includes basic support for detecting and processing Kotlin source files in plugins when running `ern create-plugin-config`. Added new test cases to verify new use cases.

Closes #1757

### DIFF
--- a/ern-core/src/AndroidPluginConfigGenerator.ts
+++ b/ern-core/src/AndroidPluginConfigGenerator.ts
@@ -12,11 +12,11 @@ import android.support.annotation.NonNull;
 import android.support.annotation.Nullable;
 
 import com.facebook.react.ReactPackage;
-import {{fullyQualifiedPackageClassName}};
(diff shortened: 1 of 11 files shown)


belemaire (approved):
:+1:

PR #1841, opened 2021-11-26
Automate publishing and versioning

friederbluemle:
This PR introduces a new workflow **version.yml**, which, together with **publish.yml**, fully automates new releases of Electrode Native.

After this is merged, the following is **no longer required**:

- Manual creation/synchronization of release branches
- Updates to `lerna.json`
- Locally committing and tagging
- Pushing of tags

Instead, we now have two separate workflows that work together as follows:

1. **Step 1: version.yml** (GitHub side, triggered **manually**)
  Once the repo is in a state ready for a new release, the version workflow is triggered on the respective branch:
    - The _default_ branch for a new _minor_ release
    - An existing `v0.x` branch for a new patch release

    The workflow will create the new release branch (if necessary), update versions, commit changes, create tags, and push it back to the repo. It will then auto populate release notes and create a _draft_ **GitHub Release** for the new version.
2. **Step 2: publish.yml** (npm side, triggered when a **GitHub Release is published**)
  After all versioning is done there is a final chance to verify everything works as expected. Then go to GitHub Releases, select the new draft, check and update release notes and click **Publish**.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,10 +1,9 @@
 name: publish
 on:
-  push:
-    tags: ['v[0-9]+.[0-9]+.[0-9]+']
+  release:
+    types: [published]
 jobs:
   publish:
-    if: github.event.repository.fork == false
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


belemaire (approved):
Nice ! 👍

PR #1848, opened 2021-12-20
Fix yarn info generic message when package not exists

### LINKED ISSUE #1764: ern regen-api error message misleading
Right now, when running `ern regen-api` in an api repo that has not been published yet, it will fail with:

```
error Received invalid response from npm.
```

At the very least, we should provide a better error message.

ozcanzaferayan:
When using ern regen-api in not published api folder, it throws an error saying "Received invalid response from npm.". This PR fixes this yarn info command issue and replaces generic message with 'Package not found'.

Closes #1764


friederbluemle (approved):
Thanks @ozcanzaferayan :+1:

PR #1849, opened 2021-12-20
Update typescript to 4.9.4

friederbluemle:
This updates the `typescript` dependency to `4.9.4`.

It also includes one preliminary commit to update and align `@types/inquirer` (previously, it was not aligned with the current version of inquirer used, and would cause errors with the new TypeScript version).


Karthiccc23 (approved):
lgtm

PR #1852, opened 2021-12-22
Fix isYarnInstalled issue on Windows OS

### LINKED ISSUE #1616: isYarnInstalled returns false on Windows
While testing #1612 on Windows I noticed that `ern platform install` was using npm even though Yarn is installed on the system.

ozcanzaferayan:
Fixes #1616

### DIFF
--- a/ern-core/src/Platform.ts
+++ b/ern-core/src/Platform.ts
@@ -191,7 +191,11 @@ export default class Platform {
 
   public static isYarnInstalled() {
     try {
-      execSync('yarn --version 1>/dev/null 2>/dev/null');
+      if (process.platform === 'win32') {
+        execSync('yarn --version > NUL 2>&1');
+      } else {
+        execSync('yarn --version 1>/dev/null 2>/dev/null');
+      }
       return true;
     } catch (e) {
       return false;


friederbluemle:
Awesome, thanks!

Just a minor thing: I don't have a way to test right now, but does Windows support stderr (`2>`)?
Can you maybe update the line to read `execSync('yarn --version > NUL');`?
See https://superuser.com/questions/134214/is-there-an-equivalent-to-dev-null-in-windows

Also please rebase the branch.

ozcanzaferayan:
> Awesome, thanks!
> 
> Just a minor thing: I don't have a way to test right now, but does Windows support stderr (`2>`)? Can you maybe update the line to read `execSync('yarn --version > NUL');`? See https://superuser.com/questions/134214/is-there-an-equivalent-to-dev-null-in-windows
> 
> Also please rebase the branch.

Hi, @friederbluemle 

I tried superuser's solution, but if yarn doesnt installed on machine, it prints error. And yes Windows support 2> syntax.

I rebased branch now. Thanks.

ozcanzaferayan:
I used this solution btw:
https://stackoverflow.com/a/4507627/2537972

friederbluemle:
Aah okay, thanks for the additional details!

In order to keep the previous/expected behavior which redirects _both_ stderr and stdout (like on Linux/macOS), does something like this work: `execSync('yarn --version > NUL 2>&1');`?
Related: Have you had a chance to also try the `where` command that some have suggested to find out if a command is available on Windows?

ozcanzaferayan:
> Aah okay, thanks for the additional details!
> 
> In order to keep the previous/expected behavior which redirects _both_ stderr and stdout (like on Linux/macOS), does something like this work: `execSync('yarn --version > NUL 2>&1');`? Related: Have you had a chance to also try the `where` command that some have suggested to find out if a command is available on Windows?

Yes i tried your mentioned command and it worked. My committed command also works because stdio default set as "pipe". But if this behavior changing in CI/CD, it will not gonna work as default. So i agree with 2>1 syntax 

Im tried where operator as well. We can use where operator or your 2>1 syntax. Your choice :)

friederbluemle (approved):
:+1:

PR #1855, opened 2022-01-20
Update file format when generating api

zaferbozkurt:
Updated the format of the js files exported with "ern regen-api".

VSCode intellisense not working with generated js files.

### DIFF
--- a/ern-api-gen/test/generateProject-test.js
+++ b/ern-api-gen/test/generateProject-test.js
@@ -70,7 +70,7 @@ describe('generatePackageJson', function () {
       name: 'test',
       version: '1.0.0',
       license: 'ISC',
-      main: 'javascript/src/index.js',
+      main: 'javascript/src/index.d.ts',
       scripts: {
         flow: 'flow',
       },


friederbluemle:
Hi @zaferbozkurt - Thanks for the PR.
I have not tested it locally yet, but I don't think just changing the file extensions make these files valid TypeScript files. There are (currently) no type annotations in the generated api files, so it really is just plain JavaScript.
Can you explain a bit more what you mean by "VSCode intellisense not working" - Is it possible to make VS Code recognize these files?

zaferbozkurt:
> Hi @zaferbozkurt - Thanks for the PR.
> I have not tested it locally yet, but I don't think just changing the file extensions make these files valid TypeScript files. There are (currently) no type annotations in the generated api files, so it really is just plain JavaScript.
> Can you explain a bit more what you mean by "VSCode intellisense not working" - Is it possible to make VS Code recognize these files?

Hi @friederbluemle, sorry for late reply.

When we import api in VSCode we get the following error and while using the api, we can’t view the methods with intellisense.

Screen Shot 2022-02-02 at 15 07 40


I agree these files not fully valid typescript. Flow type added to JS files. Therefore, simply converting the file extension to typescript will solve the error and intellesense.

PR #1860, opened 2022-03-28
Fix iOS container generation for RN 0.64+

friederbluemle:
iOS container builds for RN 0.64+ started failing on 2022-03-23 due to a change in the external dependency `react-native-codegen` (which the container generator added without specifying a version).

This aligns the version of react-native-codegen with the version of React Native during iOS container generation.


ozcanzaferayan:
This is one of the best PRs so far. Thank you  🤗

PR #1863, opened 2022-05-26
Update rule of create custom publisher/transformer

zaferbozkurt:
Updated prefix rule. 

Because we must use scope company prefix. Like this `@company/ern-container-transformer-zfr`


friederbluemle:
Hi @zaferbozkurt - The change looks good 👍 Can you rebase please?

friederbluemle:
I just rebased the commit and slightly updated it (made it clear that the string `ern-container-{publisher|transformer}` is no longer a _prefix_).

Will merge once CI is green. Thanks for the contribution @zaferbozkurt :+1:

PR #1870, opened 2023-01-05
update mavenPublisher in miniappRunner to query version from manifest

Karthiccc23:
- Add support to query maven publisher version from manifest for backward compatibility.
- Needs to be merged with -> https://github.com/electrode-io/electrode-native-manifest/pull/237


friederbluemle (approved):
👍

PR #1872, opened 2023-01-07
Migrate to androidx package

Karthiccc23:
Update erncontainer-gen and ern-api-gen to use androidx packages
Migration to Androidx - https://developer.android.com/jetpack/androidx/migrate


friederbluemle (approved):
👍

PR #1873, opened 2023-01-18
Automatically convert plugins to AndroidX if RN>60

friederbluemle:
When targeting React Native 0.60.0 or later (which uses AndroidX), attempt to automatically convert legacy plugin sources from `android.support` to `androidx`.

In a separate commit, this also contains minor updates to the container hull, and an update to the Android plugin config generator: Insert one blank line between `android.` and `androidx.` top level imports (this blank line is part of default Android Studio formatting, inserting it ensures no unnecessary reformatting will occur).


Karthiccc23 (approved):
lgtm

PR #1874, opened 2023-01-19
Ensure composite dependencies are version limited

friederbluemle:
Use versions restricted to semver major component for additional packages added to the composite to avoid breaking compatibility.


Karthiccc23 (approved):
lgtm

PR #1876, opened 2023-02-27
Use standard semver version for api gen bridge

friederbluemle:
Instead of `{major}.{minor}.x`, use standard `^{major}.{minor}.{patch}` for `react-native-electrode-bridge` in generated api projects.

Note: `{major}.{minor}.x` is effectively `~{major}.{minor}.{patch}` - There is no reason to limit only patch updates (if semver is followed correctly).


Karthiccc23 (approved):
lgtm

PR #1880, opened 2023-03-07
CI: Remove Node.js 12, add Node.js 18

friederbluemle:
Node.js 12 is end-of-life (and also deprecated on GitHub Actions). Remove it from CI matrix build, and add Node.js 18 instead.

image

~This also removes the engines restriction from package.json.~ (will be done in separate PR).


Karthiccc23 (approved):
lgtm

PR #1884, opened 2023-03-10
Update and fix files generated by create-miniapp

friederbluemle:
Fix logic error that prevented .npmignore from being created. Remove unnecessary files generated by later versions of RN.


Karthiccc23 (approved):
lgtm

PR #1885, opened 2023-03-10
Use androidx imports in api-impl-gen

friederbluemle:


Karthiccc23 (approved):

friederbluemle:
On closer inspection, it turns out there is quite a bit more support library leftover in ern-api-impl-gen which is currently causing some issues.
For the next release it is probably safer to revert this change - I'll follow up with a more thorough PR at a later date.

PR #1888, opened 2023-03-27
Update and fix plugin conversion to support Androidx

Karthiccc23:
When targeting React Native 0.60.0 or later (which uses AndroidX), attempt to automatically convert legacy plugin sources from android.support to androidx.

### DIFF
--- a/ern-container-gen-android/src/AndroidGenerator.ts
+++ b/ern-container-gen-android/src/AndroidGenerator.ts
@@ -218,27 +218,21 @@ export default class AndroidGenerator implements ContainerGenerator {
             config.outDir,
             'lib/src/main',
           );
+
           if (semver.gte(reactNativePlugin.version, '0.60.0')) {
-            const filesWithSupportLib: string[] = [];
-            shell.pushd(relPathToPluginSource);
-            shell.ls('**/*.java').forEach((file) => {
-              if (shell.grep('android.support', file).trim().length !== 0) {
-                filesWithSupportLib.push(file);
-              }
-            });
-            if (filesWithSupportLib.length !== 0) {
+            const convertedFiles = this.convertToAndroidX(
+              relPathToPluginSource,
+            );
+            if (convertedFiles > 0) {
               log.info(
                 `${plugin.name} contains source files with references to the Android Support Library (android.support.*)`,
               );
-              filesWithSupportLib.forEach((file) => {
-                shell.sed('-i', 'android.support', 'androidx', file);
-              });
               log.info(
-                `${filesWithSupportLib.length} files successfully converted to use AndroidX (androidx.*)`,
+                `${convertedFiles} files successfully converted to use AndroidX (androidx.*)`,
               );
             }
-            shell.popd();
           }
+
           shell.cp('-R', relPathToPluginSource, absPathToCopyPluginSourceTo);
         }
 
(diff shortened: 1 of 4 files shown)


ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -707,7 +691,49 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
...
+  /**
+   * Modify Plugins to support Androidx
+   * eg: import android.support.annotation.nonNull => import androidx.annotation.nonNull
+   */
+  public updatePluginsToAndroidxImports(
+    rnVersion: string,
+    plugin: PackagePath,
+    outDir: string,
+  ) {
+    if (semver.gte(rnVersion, '0.60.0')) {
```
friederbluemle, line 713: A few suggested changes for this function:

Move the if check _outside_ the function, i.e. no _not_ pass `rnVersion` into the function (that logic should not be part of it).
The only reason `plugin` is passed is to log it - Logging is also something that should _not_ be part of this function - Especially now that it is used in two separate places (the message would be inaccurate and misleading when used on the plugin _hook_ files). Instead, what I suggest is to **return the number of processed files** (`filesWithSupportLib.length`) from this function, and then perform any potential logging on the caller side.
The only remaining parameter to the function should be `outDir` (also suggest so rename it to simply `dir`).

ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -707,7 +691,49 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
...
+          `${plugin.name} contains source files with references to the Android Support Library (android.support.*)`,
+        );
+        filesWithSupportLib.forEach((file) => {
+          shell.sed('-i', 'android.support', 'androidx', file);
+        });
+        log.info(
+          `${filesWithSupportLib.length} files successfully converted to use AndroidX (androidx.*)`,
+        );
       }
+      shell.popd();
```
friederbluemle, line 736: This statement is currently _outside_ the if block, and we do _not_ want this to be executed if rn version is < 0.60 - Please move into if block (or, after addressing suggested refactoring, the if block will go away completely).

ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -707,7 +691,49 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
...
+          rnVersion,
+          plugin,
+          pathToCopyPluginConfigHookTo,
+        );
+      }
+    }
+  }
+
+  /**
+   * Modify Plugins to support Androidx
```
friederbluemle, line 705: Minor: Please use correct caps in docs: `Androidx` -> `AndroidX` - With the refactored generic version, the subject line here could be something like `Convert files in a directory from support library to AndroidX`.

ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -707,7 +691,49 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
...
+   * eg: import android.support.annotation.nonNull => import androidx.annotation.nonNull
```
friederbluemle, line 706: `nonNull` -> `NonNull`

ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -707,7 +691,49 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
...
+   */
+  public updatePluginsToAndroidxImports(
```
friederbluemle, line 708: Thoughts about a more accurate name for this function? How about `convertToAndroidX`?

friederbluemle (commented):
Great, thanks @Karthiccc23 👍 Left a few suggestions. Also, if possible, could you add a unit test for the new generalized function?

pawankst (approved):
the changes look good. thanks karthic, frieder

PR #1889, opened 2023-03-28
Update and fix plugin conversion to Androidx

Karthiccc23:
When targeting React Native 0.60.0 or later (which uses AndroidX), attempt to automatically convert legacy plugin sources from android.support to androidx.

### DIFF
--- a/ern-container-gen-android/src/AndroidGenerator.ts
+++ b/ern-container-gen-android/src/AndroidGenerator.ts
@@ -218,27 +218,21 @@ export default class AndroidGenerator implements ContainerGenerator {
             config.outDir,
             'lib/src/main',
           );
+
           if (semver.gte(reactNativePlugin.version, '0.60.0')) {
-            const filesWithSupportLib: string[] = [];
-            shell.pushd(relPathToPluginSource);
-            shell.ls('**/*.java').forEach((file) => {
-              if (shell.grep('android.support', file).trim().length !== 0) {
-                filesWithSupportLib.push(file);
-              }
-            });
-            if (filesWithSupportLib.length !== 0) {
+            const convertedFiles = this.convertToAndroidX(
+              relPathToPluginSource,
+            );
+            if (convertedFiles > 0) {
               log.info(
                 `${plugin.name} contains source files with references to the Android Support Library (android.support.*)`,
               );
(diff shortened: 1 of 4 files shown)


ern-container-gen-android/src/AndroidGenerator.ts
```diff
@@ -707,10 +701,51 @@ You should replace "${annotationProcessorPrefix}:${dependency}" with "annotation
...
+    shell
+      .ls('-R', '.')
+      .filter((file) => file.match(/\.(java|kt)$/))
+      .forEach((file) => {
+        if (shell.grep('android.support', file).trim().length !== 0) {
+          filesWithSupportLib.push(file);
+        }
+      });
+
+    if (filesWithSupportLib.length !== 0) {
```
friederbluemle, line 738: I don't think this `if` check here is necessary. `forEach` on an empty array will do the expected thing (nothing), or am I wrong?

ern-container-gen-android/test/convertToAndroidX-test.ts
```diff
@@ -0,0 +1,19 @@
+import { expect } from 'chai';
+import { AndroidGenerator } from 'ern-container-gen-android';
+import { shell, createTmpDir } from 'ern-core';
+import path from 'path';
+
+describe('Test Convert files in the directory to AndroidX Imports Logic', () => {
+  const tmpDir = createTmpDir();
+  const fixturesPath = path.join(__dirname, 'fixtures');
+
+  shell.cp('-rf', fixturesPath, tmpDir);
```
friederbluemle, line 10: OK for now - In the future we should look into a different approach that will allow the unit tests to run without actual file system I/O.

ern-container-gen-android/test/convertToAndroidX-test.ts
```diff
@@ -0,0 +1,19 @@
...
+
+  it('Expect files in directory to be converted to AndroidX', () => {
```
friederbluemle, line 12: Can you slightly reword these test cases, so that they read like a sentence? For example:
`it('converts files in a directory to AndroidX', ..`
Similar below: `it('returns ...`.

friederbluemle (approved):
Thanks @Karthiccc23 👍

Issue #1891, opened 2023-04-13
ern run-ios failing with mkdirp error


fhkarczeski:
Hello everyone,

Starting last Monday April 10 we noticed that existing apps can no longer run locally on iOS. Running the `ern run-ios` command gives this error:
```
✔ Booting iOS Simulator 0s
✖ Building iOS Runner project 36s
✖ An error occurred: iOS Runner build failed [xcbuild exit code 65].
✖ To troubleshoot this build failure, we recommend building the Runner iOS project from Xcode.
✖ You can open the Runner project in Xcode manually or by running 'open ios/ErnRunner.xcodeproj'.
✖ Building the Runner from Xcode will provide more meaningful error reporting that can be of help
✖ to pinpoint the cause of the build failure.
```

I tried to build the .xcworkspace on Xcode and I got this error:
```
/.ern/containergen/out/ios/node_modules/react-native/scripts/generate-specs-cli.js:47
  mkdirp.sync(outputDirectory);
         ^

TypeError: mkdirp.sync is not a function
    at generateSpec (/.ern/containergen/out/ios/node_modules/react-native/scripts/generate-specs-cli.js:47:10)
    at main (/.ern/containergen/out/ios/node_modules/react-native/scripts/generate-specs-cli.js:91:3)
    at Object.<anonymous> (/.ern/containergen/out/ios/node_modules/react-native/scripts/generate-specs-cli.js:94:1)
    at Module._compile (node:internal/modules/cjs/loader:1126:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1180:10)
    at Module.load (node:internal/modules/cjs/loader:1004:32)
    at Function.Module._load (node:internal/modules/cjs/loader:839:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:17:47
Command PhaseScriptExecution failed with a nonzero exit code
```

Looking at the `generate-specs-cli.js` file, I see that it is requiring mkdirp, this is the line throwing the error:
`const mkdirp = require('mkdirp');`

Looking into the mkdirp library it is loading, I see that it is exporting this:
```
export const mkdirp = Object.assign(async (path, opts) => {
    path = pathArg(path);
    const resolved = optsArg(opts);
    return useNative(resolved)
        ? mkdirpNative(path, resolved)
        : mkdirpManual(path, resolved);
}, {
    mkdirpSync,
    mkdirpNative,
[...]

PR #1894, opened 2023-04-14
Fixes bug with the latest version of the mkdirp

fhkarczeski:
Resolves the issue introduced by mkdirp v3.0.0 [described here](https://github.com/electrode-io/electrode-native/issues/1891#issuecomment-1509005678).

### DIFF
--- a/ern-container-gen-ios/src/IosGenerator.ts
+++ b/ern-container-gen-ios/src/IosGenerator.ts
@@ -388,8 +388,33 @@ Make sure to run these commands before building the container.`,
       await yarn.init();
       await yarn.add(PackagePath.fromString(`react-native-codegen@${version}`));
       // mkdirp and invariant are also needed
-      await yarn.add(PackagePath.fromString('mkdirp'));
-      await yarn.add(PackagePath.fromString('invariant'));
+      const requiredPackages = ['mkdirp', 'invariant'];
+
+      let yarnFile = '';
+      // Check if a yarn.lock file exists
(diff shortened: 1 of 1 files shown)


fhkarczeski:
@friederbluemle, @belemaire, @Karthiccc23, can you guys please take a look at this PR when you have a chance. It seems to be currently impacting all iOS container generation.

VinSpee:
Hi! this is preventing my team from getting started with electrode-native. Any thoughts of fixing/releasing?

fhkarczeski:
@VinSpee,

Our team is running the dev version with the fixes I proposed. You can check this document out on how to run it in dev:
https://github.com/electrode-io/electrode-native/blob/master/CONTRIBUTING.md

You can manually add the changes from this PR to your dev version until someone from the Electrode team has time to check this PR out.

friederbluemle:
Hi @fhkarczeski - Thanks a lot for the detailed analysis and the PR!
I did a quick test locally, and was not able to reproduce - But I also noticed that version 3.0.1 of mkdirp was released just a couple of days ago. Is it possible that it fixed the issue you were seeing? I saw something in the changelog that default/named exports were removed, then added back again.

As a side note (not related to ERN): `mkdirp` is a widely used, low level package that is 13 years old - Introducing such a breaking change (major version or not), after all this time, for no good reason (other than "cleanup"?) seems pretty reckless.. But in any case, it is also not ideal that we omit the version number when adding the package, and it's definitely an area with some major potential for improvement/simplification.

PR #1897, opened 2023-05-10
Update azure macOS image to 11

Karthiccc23:
MacOS 10.15 is not available in azure pipelines. Update to BigSur https://learn.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops&tabs=yaml


friederbluemle (approved):
Thanks @Karthiccc23 👍

friederbluemle (approved):
Thanks @Karthiccc23 👍

PR #1900, opened 2023-09-25
Fix pod install with repo update to update specs

Karthiccc23:
-> Updating Manifest pod source with new spec doesn't update the pod with new version or find the pod source. Pod install with repo-update will update the local pod specs.
 pod install --repo-update
-> CI builds tends to fail installing these pods when adding new pod spec sources.

### DIFF
--- a/ern-container-gen-ios/src/IosGenerator.ts
+++ b/ern-container-gen-ios/src/IosGenerator.ts
@@ -284,6 +284,7 @@ Make sure to run these commands before building the container.`,
 
     if (semver.gte(reactNativePlugin.version!, '0.61.0')) {
       // For full iOS container generation, run 'pod install'
+      // repo-update - to update local specs
       // and also add all essential node_modules (needed for the build)
       // to the container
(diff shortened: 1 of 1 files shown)


friederbluemle:
Hmm, so while `--repo-update` _can_ fix certain issues, I am not sure if it's the right approach to _always_/unconditionally pass that flag. This should be an explicit operation on the user side. We simply don't know what repos users have configured locally.

leekuo:
I believe the `--repo-update` flag just runs `pod repo update` before `pod install`.  Alternatively, the user can also just run `pod repo update` before generating the iOS container.

Source: https://github.com/CocoaPods/CocoaPods/blob/479e71131962e3e560e4cc0601d39860c2a590ce/lib/cocoapods/command/install.rb#L32

Karthiccc23:
@friederbluemle @leekuo  closing it, as this is not required from ERN platform. I have also tested pod repo update from generated container. Works good.

PR #1902, opened 2023-10-05
Add optional podRepoUpdate flag in iosConfig

Karthiccc23:
Add optional podRepoUpdate flag in iosConfig

- Can be used to update the pod specs when running pod install in container generation. Flag is by default disabled and can be enabled in cauldron below.
This flag was made to use in regen-container command - it generates the container with updating the pod specs during pod install and updates the cauldron with new version and yarnlocks.

```
{
  "containerGenerator": {
    "iosConfig": {
      "podRepoUpdate": true
    }
  }
}
```

or can be using in extra argument in create-container command 
--extra '{"iosConfig": {"deploymentTarget": "11.0", "podRepoUpdate": true}}'

### DIFF
--- a/docs/platform-parts/container/container-integration.md
+++ b/docs/platform-parts/container/container-integration.md
@@ -386,6 +386,18 @@ You can override the iOS deployment target version to use by setting `iosConfig`
 }
 ```
 
+You can run pod repo update before pod install by setting `podRepoUpdate` in the cauldron as show below.
+
+```json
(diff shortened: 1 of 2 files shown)


jperiyar (approved):

friederbluemle (approved):
Thanks @Karthiccc23 👍

leekuo (approved):
LGTM

PR #1903, opened 2023-12-30
ci: Use GITHUB_OUTPUT envvar instead of set-output command

arunsathiya:
`save-state` and `set-output` commands used in GitHub Actions are deprecated and [GitHub recommends using environment files](https://github.blog/changelog/2023-07-24-github-actions-update-on-save-state-and-set-output-commands/).

This PR updates the usage of `set-output` to `$GITHUB_OUTPUT`

Instructions for envvar usage from GitHub docs:

https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-output-parameter

### DIFF
--- a/.github/workflows/version.yml
+++ b/.github/workflows/version.yml
@@ -18,8 +18,8 @@ jobs:
       - name: Set variables
         id: vars
         run: |
-          echo ::set-output name=currentVersion::"$(jq -r .version lerna.json)"
-          echo ::set-output name=lastMinor::"$((npm --silent info ern-core version || echo '0.0.0') | sed 's/[0-9]*\.\([0-9]*\).*/\1/')"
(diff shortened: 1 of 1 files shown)


friederbluemle (approved):
Thank you @arunsathiya :+1:

Just one small update: I removed the commit [Quote envvar to match documentation](https://github.com/electrode-io/electrode-native/pull/1903/commits/d7d2fb0a0d5b81d167536dd95edb10304548d355) again - It is correct _some_ documentation sources include the quotes, but they don't appear to be required. [The official blog post](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/) has a sample with the non-quoted variable:

---

image

friederbluemle:
We'll address the failing Azure CI (unrelated to this PR) separately. Thanks @arunsathiya!

PR #1913, opened 2025-06-06
Update yargs to use last latest stable version

r0h0gg6:
For iOS

**yargs** has published a new version 18.0.0 which has introduced a breaking change for electrode-native
https://github.com/yargs/yargs/releases/tag/v18.0.0

Currently IosGenerator.ts is configured to always pull the latest version of **yargs** which causes build issues during container regeneration.

Fix is to stop pulling latest version of yargs and instead use the last latest stable version of **yargs** which is **17.7.2**


friederbluemle:
@r0h0gg6 - Can you please either: Add `r0h0gg6@homeoffice.wal-mart.com` to your GitHub account, or (preferably), change the author/committer email used with the commit in this PR. Right now, the CLA assistant is not able to recognize it:

---

image

